PR #143, opened 2015-03-25
Update YahooOpenIdAttributesDefinition.java

faisal77:
Yahoo return "fullname" attribute instead of "display_name"


leleuj:
YES

leleuj:
Thx for reporting the issue.

PR #145, opened 2015-03-29
Pac4j-144: Add a ClientsFactory

leleuj:
To remove the one already present in the `j2e-pac4j` library.


miremond:
+1 but the Object param is not Uncle Bob complient ;)

Issue #147, opened 2015-04-24
Need to be able to set redirect_uri to 'postmessage'


davideanderson:
I am working on a google integration using a one time server side code for authorization token exchange.  Up to this point, I have been able to use the google/oauth client of pac4j to do so.  However, I have run into a problem.  I need to be able to set the request_uri to be 'postmessage', since this flow doesn't actually do a redirect.  Unfortunately, the pac4j Clients class automatically appends the client_name to every redirect_uri which doesn't already include it.  This means the redirect_uri ends up looking like 'postmessage&client_name=GoogleClient"  Google then rejects the request indicating 'redirect_uri_mismatch'.  I looked at subclassing Clients so I could disable that portion of the initialization, but Clients is marked as final, so I can't do that either.  Here is where they talk about using 'postmessage' as the redirect uri (down in Migrations): https://developers.google.com/+/sign-in-faq

leleuj:
You're right: the pac4j `Clients` class adds the `client_name` parameter in the url if it doesn't find it. This happens during the (implicit) initialization, in the `internalInit()` method: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/client/Clients.java#L71.

I'm not sure you can easily trick the system on this. The right change for that would be to add a check on "postmessage" in the `internalInit` method of the `Clients` class: would you mind sending a PR for that?

davideanderson:
Yes, I can do the PR.  An alternative to doing a hard check for 'postmessage' would be to add an option to either the client or the Clients class to omit the implicit addition of the client_name parameter.  Would you prefer that?

leleuj:
I'm hesitating on this. As you suggest: not adding the `client_name` parameter seems to be a more generic mechanism, reusable for others purposes.
Let's do it your way and add a `boolean` to each `Client` which says if the `client_name` parameter must be added (`true` by default) and use it in the `internalInit()` method of the `Clients` class.

davideanderson:
Here is a PR: https://github.com/pac4j/pac4j/pull/149

leleuj:
Thanks for the PR, I just merged it. Change your dependency to the 1.7.1-SNAPSHOT version to use / confirm the fix.

davideanderson:
It works.  Thanks for the fast responses.

PR #149, opened 2015-04-27
Add option to prevent implicit addition of client name to client callback urls

davideanderson:
How does this look?

https://github.com/pac4j/pac4j/issues/147


leleuj:
Looks good! Thanks

PR #163, opened 2015-05-14
New authorizer concept

leleuj:


pac4j-core/src/main/java/org/pac4j/core/authorization/OrMultiAuthorizer.java
```diff
@@ -0,0 +1,49 @@
...
+ * @param <U> the user profile
+ * @author Jerome Leleu
+ * @since 1.8.0
+ */
+public class OrMultiAuthorizer<U extends CommonProfile> implements Authorizer<U> {
+
+    protected final Authorizer<U>[] authorizers;
+
+    public OrMultiAuthorizer(final Authorizer<U>[] authorizers) {
+        this.authorizers = authorizers;
```
miremond: maybe you can copy the array to avoid modification from the outside

miremond:
+1

pac4j-core/src/main/java/org/pac4j/core/authorization/OrMultiAuthorizer.java
leleuj: You're right: that may be a problem. Though, if you take a look at the `HasAccessRoleAuthorizer`, the `Authorizer` classes are built internally. Maybe I should make this constructor `protected`. Will keep that somewhere in my mind ;-)

jeroen4now:
+100

PR #167, opened 2015-05-30
Support of  ok.ru (popular social network in CIS countries) as OAuth2 provider

imayka:
Tested on *Spring Security pac4j Demo* project (will create pull request for that project later). No unit / integration tests implemented yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,7 @@ Learn more by browsing the [Javadoc](https://pac4j.github.io/pac4j/apidocs/pac4j
 <tr><td>CAS server using OAuth Wrapper</td><td>OAuth 2.0</td><td>pac4j-oauth</td><td>CasOAuthWrapperClient</td><td>CasOAuthWrapperProfile</td></tr>
 <tr><td>DropBox</td><td>OAuth 1.0</td><td>pac4j-oauth</td><td>DropBoxClient</td><td>DropBoxProfile</td></tr>
 <tr><td>Facebook</td><td>OAuth 2.0</td><td>pac4j-oauth</td><td>FacebookClient</td><td>FacebookProfile</td></tr>
+<tr><td>Odnoklassniki</td><td>OAuth 2.0</td><td>pac4j-oauth</td><td>OkClient</td><td>OkProfile</td></tr>
 <tr><td>GitHub</td><td>OAuth 2.0</td><td>pac4j-oauth</td><td>GitHubClient</td><td>GitHubProfile</td></tr>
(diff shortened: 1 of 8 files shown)


imayka:
Sample config below:

    <bean id="okClient" class="org.pac4j.oauth.client.OkClient">
        <property name="key" value="1139019264" />
        <property name="publicKey" value="CBAPAFOEEBABABABA" />
        <property name="secret" value="479452FD7CA726DF558B4303" />
    </bean>

leleuj:
Thanks for this new implementation. The source code looks great.

Some feedbacks:
- license headers are missing
- about testing, the idea is really to add unit tests to check basic stuffs on the client and profile and integration tests to ensure the authentication always peforms properly. If you take a look at the https://github.com/pac4j/pac4j/blob/master/pac4j-oauth/src/test/java/org/pac4j/oauth/client/FacebookClientIT.java class, the `testClone` and `testMissingFields` methods should be in a `FacebookClientTests` class (unit tests launched through `mvn test`) while the `FacebookClientIT` class is really for integration tests (performed via `mvn verify`).

imayka:
Jerome, please see my recent changes (integration test added), but it still doesn't work because i haven't idea how to setup properly callback url. Could you please explain on which port/url it runs by default? 

Also, as i can see few another IT test fails in some reason.

Regards

leleuj:
Excellent! Forget about the other IT tests which fail, I'll fix them later. Generally, I'll check all IT tests before a release.

In your IT test, you will perform a real authentication via ok.ru. After a successful authentication (it means that you use a real account), you will be redirected to the callback url: it must a real url existing in the internet, so I sometimes use www.google.com or www.pac4j.org to ensure these are real websites (depending on what I can configure in the provider). Assuming this callback url is a real one, the calback will work and you'll be able to extract the returned OAuth credentials (and finish the authentication process).

imayka:
Thanks for explanation. Not test partially works (user profile extracted), but fails 3 times on verifyProfile->           assertTrue(profile.getPictureUrl().contains(pictureUrl)) (maybe serialization issue), trying to handle it...

imayka:
Now test works, but sometimes fails because ok.ru returns http://i508.mycdn.me/res/stub_50x50.gif and http://i500.mycdn.me/res/stub_50x50.gif (differs in 1 digit - 0 and 8), so assertCommonProfile()  test fails. Very strange behavior :(. Maybe it related only to "stub" picture, because my profile doesn't have it.

imayka:
Done

leleuj:
Thanks. Will merge...

PR #164, opened 2015-05-17
Opensaml v311

mmoayyed:
Handles https://github.com/pac4j/pac4j/issues/162

PR #171, opened 2015-06-07
OpenSAML 3.x for master

mmoayyed:
Moving #164 and other subsequent commits on 1.7 over to master.


mmoayyed:
So, it looks like callback urls are gone in master. What is the alternative?

mmoayyed:
This should be ready to merg. I fixed the API issues, ran tests and merged with master.

leleuj:
Callback urls only exist now for indirect clients, the ones that need redirections to and from the identity provider, it comes from the `IndirectClient` class inheritance.

pac4j-saml/pom.xml
```diff
@@ -25,17 +25,138 @@
...
+            <groupId>org.apache.velocity</groupId>
+            <artifactId>velocity</artifactId>
+            <version>${velocity.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-core</artifactId>
+            <version>${spring.version}</version>
+            <scope>compile</scope>
+            <optional>true</optional>
```
leleuj: Is it really optional? When do I need to explicitly declare this dependency? For the metadata output?

pac4j-saml/src/main/java/org/pac4j/saml/context/SAML2ContextProvider.java
```diff
@@ -0,0 +1,169 @@
+/*
+  Copyright 2012 -2014 Michael Remond
```
leleuj: The license header should be `2012 - 2015 pac4j organization` and the author should be kept by the `@author` annotation in the class Javadoc.

pac4j-saml/src/main/java/org/pac4j/saml/crypto/ExplicitSignatureTrustEngineProvider.java
```diff
@@ -0,0 +1,63 @@
...
+import org.opensaml.xmlsec.config.DefaultSecurityConfigurationBootstrap;
+import org.opensaml.xmlsec.keyinfo.KeyInfoCredentialResolver;
+import org.opensaml.xmlsec.signature.support.SignatureTrustEngine;
+import org.opensaml.xmlsec.signature.support.impl.ExplicitKeySignatureTrustEngine;
+import org.pac4j.saml.exceptions.SAMLException;
+
+/**
+ * Provider returning well configured {@link SignatureTrustEngine} instances.
+ * 
+ * @author Michael Remond
```
leleuj: How can this one be a new class from the author `Michael Remond`?

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
```diff
@@ -0,0 +1,148 @@
...
+import org.pac4j.core.util.CommonHelper;
+import org.pac4j.saml.exceptions.SAMLException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Class responsible for loading a private key from a JKS keystore and returning the corresponding {@link Credential}
+ * opensaml object.
+ * 
+ * @author Michael Remond
```
leleuj: Same comment.

leleuj:
@miremond : I don't remember if you gave us your feedback on this PR (close to the one targeted version 1.7.x)

pac4j-saml/pom.xml
mmoayyed: Optional in the sense it only gets pulled when the module is actually used, otherwise no. The OpenSAML library requires Spring-core so it could construct Resource objects for pulling metadata from http and otherwise resources. It's mostly for pulling MD, not outputting it.

pac4j-saml/src/main/java/org/pac4j/saml/crypto/ExplicitSignatureTrustEngineProvider.java
mmoayyed: Happy to change, but this is mostly a refactoring of an older component that Michael had in place and I simply gave it a new name and added mods. That said, it makes sense to change it.

mmoayyed:
updated all copyrights for the saml module.

pac4j-saml/pom.xml
mmoayyed: Maybe we should make it a compile-time dependency. We cant really get by without it.

leleuj: So if it's not really optional, we should remove the `<optional>true</optional` tag. I feared to add such dependency.

Which `opensaml` library specifically requires it?

mmoayyed: The `net.shibboleth.ext.spring.resource.ResourceHelper;` class which is part of the Shibboleth-ext package. Used for resolving metadata from a variety of resources. The ResourceHelper, although part of Shibboleth, is sort of a helper library that pulls in Spring when constructing resources for opensaml's metadata. We could remove ResourceHelper and write our own Resources, but that probably would a more complicated task. 

The cool thing is, my dependency graph shows spring-core is the only thing that is required. Nothing else is pulled in. So fingers crossed, no other conflicts should pop up. Spring has become such a prevalent library that it's hard to do anything without it!

I'll remove the optional tag.

mmoayyed:
Pending further feedback, this is ready to merge.

leleuj:
I will merge this one and re-test the 1.7.1-SNAPSHOT to see if it's ready for release...

PR #175, opened 2015-06-11
Cas Rest Client

mmoayyed:
```java
final URL endpointURL = new URL("https://mmoayyed.unicon.net:8443/cas/v1/tickets");
final String casUrlPrefix = endpointURL.getProtocol() + "://" + endpointURL.getHost() + ":"
        + endpointURL.getPort() + "/" + endpointURL.getPath().split("/")[1];

final CasRestAuthenticator authenticator = new CasRestAuthenticator(endpointURL,
        new Cas20ServiceTicketValidator(casUrlPrefix));
final CasRestClient client = new CasRestClient(authenticator);
final UsernamePasswordCredentials creds = new UsernamePasswordCredentials("casuser", "Mellon", client.getName());

final MockWebContext context = MockWebContext.create();
context.addRequestParameter(client.getUsernameParameter(), creds.getUsername());
context.addRequestParameter(client.getPasswordParameter(), creds.getPassword());

final HttpProfile profile = client.requestTicketGrantingTicket(context);
final CasCredentials casCreds = client.requestServiceTicket(endpointURL, profile);
final CasProfile casProfile = client.validateServiceTicket(endpointURL, casCreds);
```

### DIFF
--- a/pac4j-cas/pom.xml
+++ b/pac4j-cas/pom.xml
@@ -35,9 +35,7 @@
 		<dependency>
 			<groupId>org.pac4j</groupId>
 			<artifactId>pac4j-http</artifactId>
-			<scope>compile</scope>
 			<optional>true</optional>
-			<version>${project.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.jasig.cas.client</groupId>
--- a/pac4j-cas/pom.xml
+++ b/pac4j-cas/pom.xml
@@ -33,6 +33,13 @@
 			<artifactId>pac4j-core</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.pac4j</groupId>
+			<artifactId>pac4j-http</artifactId>
+			<scope>compile</scope>
+			<optional>true</optional>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
 			<groupId>org.jasig.cas.client</groupId>
 			<artifactId>cas-client-core</artifactId>
 		</dependency>
--- a/pac4j-cas/src/main/java/org/pac4j/cas/client/rest/AbstractCasRestClient.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/client/rest/AbstractCasRestClient.java
@@ -57,17 +57,17 @@ public abstract class AbstractCasRestClient extends DirectHttpClient<UsernamePas
         setProfileCreator(new AuthenticatorProfileCreator<UsernamePasswordCredentials, HttpProfile>());
     }
 
-    public HttpProfile requestTicketGrantingTicket(final WebContext context) {
+    public HttpTGTProfile requestTicketGrantingTicket(final WebContext context) {
         final UsernamePasswordCredentials creds = this.extractor.extract(context);
         getAuthenticator().validate(creds);
-        return getProfileCreator().create(creds);
+        return (HttpTGTProfile) getProfileCreator().create(creds);
     }
 
-    public CasCredentials requestServiceTicket(final String serviceURL, final HttpProfile profile) {
+    public CasCredentials requestServiceTicket(final String serviceURL, final HttpTGTProfile profile) {
         HttpURLConnection connection = null;
         try {
             final URL endpointURL = new URL(getAuthenticator().getCasRestUrl());
-            final URL ticketURL = new URL(endpointURL, endpointURL.getPath() + "/" + profile);
+            final URL ticketURL = new URL(endpointURL, endpointURL.getPath() + "/" + profile.getTicketGrantingTicketId());
 
             connection = HttpUtils.openConnection(ticketURL);
             final String payload = HttpUtils.encodeQueryParam("service", serviceURL);
--- a/pac4j-cas/src/main/java/org/pac4j/cas/client/rest/AbstractCasRestClient.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/client/rest/AbstractCasRestClient.java
@@ -17,7 +17,6 @@
 
 package org.pac4j.cas.client.rest;
 
-
 import org.apache.commons.httpclient.HttpStatus;
 import org.jasig.cas.client.authentication.AttributePrincipal;
 import org.jasig.cas.client.validation.Assertion;
@@ -25,13 +24,10 @@ import org.jasig.cas.client.validation.TicketValidationException;
 import org.pac4j.cas.credentials.CasCredentials;
 import org.pac4j.cas.profile.CasProfile;
 import org.pac4j.core.client.BaseClient;
-import org.pac4j.core.client.Mechanism;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.exception.TechnicalException;
 import org.pac4j.http.client.direct.DirectHttpClient;
 import org.pac4j.http.credentials.UsernamePasswordCredentials;
-import org.pac4j.http.credentials.extractor.Extractor;
-import org.pac4j.http.credentials.extractor.FormExtractor;
 import org.pac4j.http.profile.HttpProfile;
 import org.pac4j.http.profile.creator.AuthenticatorProfileCreator;
 
@@ -44,29 +40,21 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 
 /**
- * This is {@link CasRestClient}.
+ * This is {@link CasRestFormClient} able to communicate to
+ * a CAS server via its REST api, to retrieve TGTs and STs
+ * and to construct CAS principals.
  *
  * @author Misagh Moayyed
- * @author Daniel Pacak
  * @since 1.8.0
  */
-public class CasRestClient extends DirectHttpClient<UsernamePasswordCredentials> {
-    public CasRestClient() {
+public abstract class AbstractCasRestClient extends DirectHttpClient<UsernamePasswordCredentials> {
+    public AbstractCasRestClient() {
+        super();
     }
 
-    public CasRestClient(final CasRestAuthenticator authenticator) {
+    public AbstractCasRestClient(final CasRestAuthenticator authenticator) {
         setAuthenticator(authenticator);
         setProfileCreator(new AuthenticatorProfileCreator<UsernamePasswordCredentials, HttpProfile>());
-        this.extractor = new FormExtractor(authenticator.getUsernameParameter(),
-                                            authenticator.getPasswordParameter(),
-                                            CasRestClient.class.getSimpleName());
-    }
-
-    @Override
-    protected BaseClient<UsernamePasswordCredentials, HttpProfile> newClient() {
-        final CasRestClient client = new CasRestClient();
-        client.setAuthenticator(getAuthenticator());
-        return client;
     }
 
     public HttpProfile requestTicketGrantingTicket(final WebContext context) {
@@ -75,14 +63,14 @@ public class CasRestClient extends DirectHttpClient<UsernamePasswordCredentials>
         return getProfileCreator().create(creds);
     }
 
-    public CasCredentials requestServiceTicket(final URL serviceURL, final HttpProfile profile) {
+    public CasCredentials requestServiceTicket(final String serviceURL, final HttpProfile profile) {
         HttpURLConnection connection = null;
         try {
-            final URL endpointURL = getAuthenticator().getEndpointURL();
+            final URL endpointURL = new URL(getAuthenticator().getCasRestUrl());
             final URL ticketURL = new URL(endpointURL, endpointURL.getPath() + "/" + profile);
 
             connection = HttpUtils.openConnection(ticketURL);
-            final String payload = HttpUtils.encodeQueryParam("service", serviceURL.toString());
+            final String payload = HttpUtils.encodeQueryParam("service", serviceURL);
 
             final BufferedWriter out = new BufferedWriter(new OutputStreamWriter(connection.getOutputStream()));
             out.write(payload);
@@ -102,10 +90,10 @@ public class CasRestClient extends DirectHttpClient<UsernamePasswordCredentials>
         }
     }
 
-    public CasProfile validateServiceTicket(final URL serviceURL, final CasCredentials ticket) {
+    public CasProfile validateServiceTicket(final String serviceURL, final CasCredentials ticket) {
         try {
             final Assertion assertion = getAuthenticator().getTicketValidator()
-                    .validate(ticket.getServiceTicket(), serviceURL.toExternalForm());
+                    .validate(ticket.getServiceTicket(), serviceURL);
             final AttributePrincipal principal = assertion.getPrincipal();
             final CasProfile casProfile = new CasProfile();
             casProfile.setId(principal.getName());
@@ -117,12 +105,16 @@ public class CasRestClient extends DirectHttpClient<UsernamePasswordCredentials>
     }
 
     @Override
-    public Mechanism getMechanism() {
-        return Mechanism.FORM_MECHANISM;
-    }
-
-    @Override
     public CasRestAuthenticator getAuthenticator() {
         return (CasRestAuthenticator) super.getAuthenticator();
     }
+
+    @Override
+    protected BaseClient<UsernamePasswordCredentials, HttpProfile> newClient() {
+        final AbstractCasRestClient client = newClientType();
+        client.setAuthenticator(getAuthenticator());
+        return client;
+    }
+
+    protected abstract AbstractCasRestClient newClientType();
 }
--- a/pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
@@ -64,7 +64,7 @@ public class CasRestAuthenticator implements Authenticator<UsernamePasswordCrede
             throw new TechnicalException("Credentials are required");
         }
         final String ticketGrantingTicketId = requestTicketGrantingTicket(credentials.getUsername(), credentials.getPassword());
-        final HttpTGTProfile profile = new HttpTGTProfile(ticketGrantingTicketId);
+        final HttpTGTProfile profile = new HttpTGTProfile(ticketGrantingTicketId, credentials.getUsername());
         credentials.setUserProfile(profile);
     }
 
--- a/pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
@@ -60,6 +60,9 @@ public class CasRestAuthenticator implements Authenticator<UsernamePasswordCrede
 
     @Override
     public void validate(final UsernamePasswordCredentials credentials) {
+        if (credentials == null || credentials.getPassword() == null || credentials.getUsername() == null) {
+            throw new TechnicalException("Credentials are required");
+        }
         final String ticketGrantingTicketId = requestTicketGrantingTicket(credentials.getUsername(), credentials.getPassword());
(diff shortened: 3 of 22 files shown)


pac4j-cas/pom.xml
```diff
@@ -33,6 +33,13 @@
 			<artifactId>pac4j-core</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.pac4j</groupId>
+			<artifactId>pac4j-http</artifactId>
+			<scope>compile</scope>
+			<optional>true</optional>
+			<version>${project.version}</version>
```
leleuj: The version should not be required here, this primary dependency should be defined in the parent `pom.xml` file.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
```diff
@@ -0,0 +1,86 @@
...
+ * This is {@link CasRestAuthenticator}.
+ *
+ * @author Misagh Moayyed
+ * @since 1.8.0
+ */
+public class CasRestAuthenticator implements Authenticator<UsernamePasswordCredentials> {
+    private final URL endpointURL;
+    private final TicketValidator ticketValidator;
+
+    public CasRestAuthenticator(final URL endpointURL, final TicketValidator ticketValidator) {
```
leleuj: This is not user friendly to need to instantiate a `ticketValidator`. I'd prefer to do something close to what is done in the `CasClient`: providing the `casServerPrefixUrl` might be sufficient to compute the `casRestUrl` or this latter can be set explicitly. I would expect two constructors:

```
public CasRestAuthenticator(final String casServerPrefixUrl)
public CasRestAuthenticator(final String casServerPrefixUrl, final String casRestUrl)
```

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
```diff
@@ -0,0 +1,86 @@
...
+            final String locationHeader = connection.getHeaderField("location");
+            final int responseCode = connection.getResponseCode();
+            if (locationHeader != null && responseCode == HttpStatus.SC_CREATED) {
+                return locationHeader.substring(locationHeader.lastIndexOf("/") + 1);
+            }
+
+            throw new IllegalStateException("Ticket granting ticket request failed: " +
+                    HttpUtils.buildHttpErrorMessage(connection));
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
```
leleuj: If a technical exception happens, a `TechnicalException` must be thrown and it won't be catched by the `DirectHttpClient`. If a functional exception happens, a `CredentialsException` or a subclass of it must be thrown and it will be catched by the `DirectHttpClient`

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
```diff
@@ -0,0 +1,86 @@
...
+            out.write(payload);
+            out.close();
+
+            final String locationHeader = connection.getHeaderField("location");
+            final int responseCode = connection.getResponseCode();
+            if (locationHeader != null && responseCode == HttpStatus.SC_CREATED) {
+                return locationHeader.substring(locationHeader.lastIndexOf("/") + 1);
+            }
+
+            throw new IllegalStateException("Ticket granting ticket request failed: " +
```
leleuj: Technical or functional exception? See my comment below...

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public CasRestAuthenticator(final URL endpointURL, final TicketValidator ticketValidator) {
+        this.endpointURL = endpointURL;
+        this.ticketValidator = ticketValidator;
+    }
+
+    @Override
+    public void validate(final UsernamePasswordCredentials credentials) {
+        final String ticketGrantingTicketId = requestTicketGrantingTicket(credentials.getUsername(), credentials.getPassword());
+        final HttpTGTProfile profile = new HttpTGTProfile(ticketGrantingTicketId);
```
leleuj: What's the identifier for this profile? We may use the `username` even if that doesn't change anything actually.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestClient.java
```diff
@@ -0,0 +1,143 @@
...
+    public CasRestClient() {
+    }
+
+    public CasRestClient(final CasRestAuthenticator authenticator) {
+        this(authenticator, new FormExtractor("username", "password",
+                CasRestClient.class.getSimpleName()));
+    }
+
+    public CasRestClient(final CasRestAuthenticator authenticator,
+                         final Extractor<UsernamePasswordCredentials> extractor) {
```
leleuj: I guess this constructor is when you want to use a basic auth extractor.

So I would do things another way:
- rename this class as `CasRestFormClient`
- refactor it to extract an abstract `CasRestClient` and create a new `CasRestBasicAuthClient` based on it

With the right mechanims...

What do you think?

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestClient.java
```diff
@@ -0,0 +1,143 @@
...
+            final BufferedWriter out = new BufferedWriter(new OutputStreamWriter(connection.getOutputStream()));
+            out.write(payload);
+            out.close();
+
+            final int responseCode = connection.getResponseCode();
+            if (responseCode == HttpStatus.SC_OK) {
+                final BufferedReader in = new BufferedReader(new InputStreamReader(connection.getInputStream()));
+                return new CasCredentials(in.readLine(), getClass().getSimpleName());
+            }
+            throw new IllegalStateException("Service ticket request for `" + profile + "` failed: " +
```
leleuj: `CredentialsException`?

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestClient.java
```diff
@@ -0,0 +1,143 @@
...
+                    HttpUtils.buildHttpErrorMessage(connection));
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
```
leleuj: `TechnicalException`?

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestClient.java
```diff
@@ -0,0 +1,143 @@
...
+        try {
+            final Assertion assertion = getAuthenticator().getTicketValidator()
+                    .validate(ticket.getServiceTicket(), serviceURL.toExternalForm());
+            final AttributePrincipal principal = assertion.getPrincipal();
+            final CasProfile casProfile = new CasProfile();
+            casProfile.setId(principal.getName());
+            casProfile.addAttributes(principal.getAttributes());
+            return casProfile;
+        } catch (TicketValidationException e) {
+            throw new RuntimeException(e);
```
leleuj: `TechnicalException`?

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/HttpTGTProfile.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.pac4j.http.profile.HttpProfile;
+
+/**
+ * This is {@link HttpTGTProfile}.
+ *
+ * @author Misagh Moayyed
+ * @since 1.8.0
+ */
+public final class HttpTGTProfile extends HttpProfile {
```
leleuj: All profiles must support to be built using this method: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java#L56 So you certainly need to add a few lines of code for this new profile.

leleuj:
Several comments about this new CAS REST client. I don't see any test as well.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
mmoayyed: I disagree. The client should be given the option to figure out how to construct ticket validation urls, and which validator should be chosen. This frees pac4j up to not make that decision for them, as Cas30Validator comes out and Cas40 and so on, developers can simply just plug in the right component, rather than waiting for pac4j to update itself, and even then, there is no choice in selecting the validator.

Also consider that future validators may not simply just need a cas url. They may need additional pieces of data. By passing pieces of data around, you tie the code here to the cas client api, whereas if we follow a DI model, the client can be injected with the thing that makes sense to the developer, and the developer knows how to construct it because he/she knows everything about the validator.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
mmoayyed: by identifier, you mean each profile has a "setId()" type of method? That just be the TGT, or the name of the class. What is the id itself used for?

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestClient.java
mmoayyed: As I understand, the REST api is always formed-based per the rest protocol. The basic auth extractor looks into headers to find credentials IIRC, and that's not now the way to pass creds per the rest protocol. Does that make sense?

mmoayyed:
Thanks for the review. I'll make a few adjustments in the coming days. 

Tests are all IT tests. Do we have something I can work with? Mock tests are no use here.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
mmoayyed: Based on your explanation, I suppose all exceptions are Technical. Checked exceptions are a nuisance in Java and should die a quick and painful death :)

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestClient.java
mmoayyed: I had a chance to review this again, and now that I think about it, I think it's best to remove the extractor option. The rest client is always based on a form-extractor and it's not needed to ask the developer to choose between Form/Basic. The developer never cares.

mmoayyed:
I updated the pull to address feedback posted here and added IT tests for the client.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
leleuj: I understand your opinion and value it. It would make sense for the LDAP implementation if I rely on ldaptive and an `Authenticator`. Because it's a very complex object I don't really want to replace it (so much work).

The philosophy of `pac4j` is to hide and abstract as much as possible from internal libraries. You should never know that it uses the `cas-client-core` library for Java or `Scribe` for OAuth support.
In the future, if `pac4j` becomes internally reactive (I had one more discussion with Tim Fox, the creator of Vertx about that), we may change the underlying library.

Your argument (be able to switch from Cas30 to Cas40) is relevant but in practice, it doesn't work: if ever we upgrade to Cas40, I'm pretty sure we'll need to change the internal logic as well. So it doesn't really make sense to offer this ability.
The last update to version 3.3.3 of the CAS client generates breaking changes: upgrading is not that seamless as well.

So I'd really prefer to set the CAS server prefix url and the CAS REST url and hide the ticket alidator.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
leleuj: The identifier must be unique for the identity provider, it's the identifier (example: 1234)
To get a unique identifier accross all identity providers, it's the typed identifier (example: FacebookProfile#1234).

It's used to rebuild a identity and rebuild a profile if necessary. It would be best to have one, the username seems to have here the right properties (unique).

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestClient.java
```diff
@@ -0,0 +1,128 @@
...
+import java.net.URL;
+
+/**
+ * This is {@link CasRestClient}.
+ *
+ * @author Misagh Moayyed
+ * @author Daniel Pacak
+ * @since 1.8.0
+ */
+public class CasRestClient extends DirectHttpClient<UsernamePasswordCredentials> {
```
leleuj: The CAS REST protocol is form based: credentials must be passed into a POST request to the CAS server. Though, the mechanism of the client is the one used by the client, not internally to communicate with the CAS server.

And the client can receive credentials from a form or a basic auth before passing them to the CAS server in a form way. So I think we really need to distinguish between a `CasRestFormClient` and a `CasRestBasicAuthClient` inheriting both from a `AbstractCasRestClient` gathering most of the logic. Both sub clients having different mechanisms...

leleuj:
What I mean by IT tests is that they should occur with a real CAS server, certainly with the one of `pac4j-test-cas`. They are generally pretty complex to setup and maintain.
But they are really necessary with changing identity providers like Facebook or Twitter which adds, updates or removes fields from time to time.

I guess you can only rely on unit tests and a simulated CAS REST API here. It will make your life easier...

leleuj:
Thanks for your work. I guess some code still needs to be adjusted, see my new comments. I'll provide more explanations if necessary...

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
mmoayyed: I think I have taken care of this. I would opt for the TGT itself as it's both unique and self explanatory. I fear providing a username as an id "may" mistakenly convey this idea to the caller that this in fact is the principal, instead of relying on the profile.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
mmoayyed: I understand you point about developer-friendliness and abstraction. I'll make the change. 

Long term, this strategy may not deliver as much because it would simply be too much work for pac4j to upgrade itself to take advantage of the underlying libs and to provide options for every feature, (because we rely on impls rather than apis) Something to think about later...

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestClient.java
mmoayyed: Got it. I do think this is just going to lead to API proliferation because at this point, the mechanism field is more of a nuisance perhaps but as you said in the other comment, it might be better to hide the extractor directly inside a dedicated class.

mmoayyed:
IT tests are done by the way.

mmoayyed:
Pull should be updated now.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/AbstractCasRestClient.java
```diff
@@ -0,0 +1,120 @@
...
+            throw new TechnicalException("Service ticket request for `" + profile + "` failed: " +
+                    HttpUtils.buildHttpErrorMessage(connection));
+        } catch (final IOException e) {
+            throw new TechnicalException(e);
+        } finally {
+            HttpUtils.closeConnection(connection);
+        }
+    }
+
+    public CasProfile validateServiceTicket(final String serviceURL, final CasCredentials ticket) {
```
leleuj: Do we really need this one? Will we really validate internally the service ticket we create? I don't think so. I guess we will call web services using service tickets, won't we?

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/CasRestAuthenticator.java
```diff
@@ -0,0 +1,117 @@
...
+ */
+public class CasRestAuthenticator implements Authenticator<UsernamePasswordCredentials> {
+    private final String casServerPrefixUrl;
+    private final String casRestUrl;
+
+    public CasRestAuthenticator(final String casServerPrefixUrl) {
+        this(casServerPrefixUrl, buildCasRestUrlFromCasServerPrefixUrl(casServerPrefixUrl));
+    }
+
+    private static String buildCasRestUrlFromCasServerPrefixUrl(final String casServerPrefixUrl) {
```
leleuj: Thanks

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/HttpTGTProfile.java
```diff
@@ -0,0 +1,64 @@
...
+    private static final long serialVersionUID = -1688563185891330018L;
+
+    private final String ticketGrantingTicketId;
+
+    public HttpTGTProfile() {
```
leleuj: Just about the identifier: yes, we can use the TGT, it's a fairly valid option, but theoretically speaking, I'd prefer something more long lasting like a username. It would make `toString()` call more relevant.
One more argument: it would be more consistent with what we do for OAuth and Facebook for example where we have a real technical identifier (close to the username) and an access token as a property (close to the TGT).

leleuj: That said, it's not a big deal. I don't want to push you too hard on this, you have already done a lot of work on this.

pac4j-cas/src/test/java/org/pac4j/cas/client/rest/TestCasRestClient.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.pac4j.core.profile.CommonProfile;
+
+/**
+ * The {@link TestCasRestClient} is responsible for
+ * executing tests against {@link CasRestFormClient}.
+ *
+ * @author Misagh Moayyed
+ */
+public class TestCasRestClient extends org.pac4j.core.profile.TestProfileHelper  {
```
leleuj: I messed up things around tests naming. Just to let you know that unit tests can be named TestSomething, SomethingTest or SomethingTests. My fault.

leleuj:
Just a few last comments before I merge. Thanks again for you work and tenacity.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/AbstractCasRestClient.java
mmoayyed: That depends on the app, if indeed the rest api is being used by a web service. If it's a standalone java application that wants to remotely act with the CAS server and have its own login screen, then probably a validation call is required.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/HttpTGTProfile.java
mmoayyed: No that's perfectly fine. I can make the change. It's best to keep things consistent.

pac4j-cas/src/main/java/org/pac4j/cas/client/rest/AbstractCasRestClient.java
leleuj: You're right. That's a valid use case.

mmoayyed:
I updated the pull to include uid as the identifier for the profile. all IT tests should pass.

Issue #173, opened 2015-06-08
Travis should not attempt to deploy PRs


mmoayyed:
Reports:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:2.7:deploy (default-deploy) on project pac4j: Failed to deploy artifacts: Could not transfer artifact org.pac4j:pac4j:pom:1.8.0-20150608.225607-7 from/to sonatype-nexus-snapshots (https://oss.sonatype.org/content/repositories/snapshots/): Failed to transfer file: https://oss.sonatype.org/content/repositories/snapshots/org/pac4j/pac4j/1.8.0-SNAPSHOT/pac4j-1.8.0-20150608.225607-7.pom. Return code is: 401, ReasonPhrase: Unauthorized. -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.

leleuj:
You're completely right. We should certainly do something like for CAS.

mmoayyed:
Already taken care of :) only commits to master are pushed.

mmoayyed:
We should try to devote some time to this PR. It's now difficult to determine if a PR actually passes, or the failure is just  noise because Travis is trying deploy that over to sonatype.

leleuj:
You're right. Here it is: https://github.com/pac4j/pac4j/pull/190, greatly inspired from CAS ;-)

mmoayyed:
Awesome! Thanks for doing this.

PR #186, opened 2015-07-04
Backported Scope Setting from 1.7.1

ParthKolekar:
`cas-support-pac4j` package has compatibility with the 1.4.x version of pac4j library. Backported the scope for the GitHubClient to add functionality.

### DIFF
--- a/pac4j-oauth/src/main/java/org/pac4j/oauth/client/GitHubClient.java
+++ b/pac4j-oauth/src/main/java/org/pac4j/oauth/client/GitHubClient.java
@@ -38,35 +38,40 @@ import com.fasterxml.jackson.databind.JsonNode;
  * @since 1.0.0
  */
 public class GitHubClient extends BaseOAuth20Client<GitHubProfile> {
-    
+  
+    public final static String DEFAULT_SCOPE = "user";
+  
+    protected String scope = DEFAULT_SCOPE;
+
     public GitHubClient() {
     }
-    
+  
     public GitHubClient(final String key, final String secret) {
         setKey(key);
         setSecret(secret);
     }
-    
+  
     @Override
     protected GitHubClient newClient() {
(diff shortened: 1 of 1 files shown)


ParthKolekar:
Build Passed. 
Deployment failed. 

:grin: 

You might want to stop auto deployments for pull requests. PR's won't have deployment tokens. 

:grin:

leleuj:
I get the idea: allow users to customize the Github scope: no problem with that.

I always prefer to start with a PR targeting the master and then backport the change if it is really necessary.

Currently, the master is on 1.8.0-SNAPSHOT and the only maintained branch is the 1.7.x which will be used by the CAS server v4.1.0 (to be released soon).
So I won't accept a PR for the 1.4.x branch. The best way to proceed here is to create a `CustomGithubClient` in your code for CAS v4.0 and propose the changes for the master and the 1.7.x branch, so that you get your change when you'll upgrade your CAS server.

And you're absolutely right: PR should never been deployed -> https://github.com/pac4j/pac4j/issues/173

ParthKolekar:
Around when is the v4.1.0 expected to be released?
I can afford to wait for the update if it is < 2 months.

mmoayyed:
Definitely less than 2 months and we'll do maintenance releases on 4.0.x as well so you can get the pac4j. If you want to post a PR to CAS, that'd be great. Target master and we'll backport it to 4.0.x

ParthKolekar:
Less than two months is good. 
As for a PR to CAS, well, I have not come across anything that is broken or needs extending yet. :smile: 

Just to clarify, when 4.1.0 comes out, I update my cas-server-* dependencies version to 4.1.0 as well? and pac4j-oauth to latest release of 1.7.x, right?

mmoayyed:
yes, though 4.1 will likely be on pac4j 1.8.

PR #187, opened 2015-07-05
Add JWT support

leleuj:
generation + authentication


leleuj:
@miremond : we talked about it, I did it ;-)

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.util.Date;
+
+/**
+ * Generates a JWT token from a user profile.
+ *
+ * @author Jerome Leleu
+ * @since 1.8.0
+ */
+public class JwtGenerator<U extends UserProfile> {
```
miremond: What is the purpose of this class? is it only for tests? If no, maybe we should parameterized the signature or encryption methods. Otherwise we can move it to the test directory

miremond:
Congrats @leleuj ! I give you a kudo!
+1

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
leleuj: Good question! This class is meant to generate a JWT from an existing user profile. The typical use case is an application where the user can login with Facebook and then generates a token to access web services. This was inspired by this discussion: https://groups.google.com/forum/?fromgroups#!topic/pac4j-users/kJMw8vvZBJY

PR #188, opened 2015-07-08
allow subclassing of SAML2Client

papegaaij:
All relevant methods in SAML2Client are either final or private, making it impossible to subclass it. In my case, I would like to use a different implementation of the CredentialProvider interface, which does not load the keystore from resource: or file:, but from the database.

This pull request contains a single commit against 1.7.x, but it should also apply to master.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
@@ -112,7 +112,7 @@ public class SAML2Client extends BaseClient<SAML2Credentials, SAML2Profile> {
     }
 
     @Override
-    protected final void internalInit() {
+    protected void internalInit() {
         CommonHelper.assertNotBlank("callbackUrl", this.callbackUrl);
         if (!this.callbackUrl.startsWith("http")) {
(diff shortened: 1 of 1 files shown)


leleuj:
Indeed, we may open a little more this class. @mmoayyed : what do you think?

mmoayyed:
Sure, that's fine with me.

mmoayyed:
Although, you are interested in extending CredentialProvider , yet most if not all methods are marked as non-final. Why? Are you just trying to cover all bases?

papegaaij:
Yes, it seemed logical to me to allow other parts to be changed as well. Most defaults seem sensible, but they are based on a rather static configuration. The used interfaces would have allowed me to change whatever I need changed, but the locked init methods prevent me from doing so. For now, I've copied the entire client class to my own project, but I'd rather modify only those parts that need to be changed for my use case.

leleuj:
@papegaaij: are we good to merge or do you want to propose more changes?

papegaaij:
@leleuj: with these changes, this class should be ok

leleuj:
OK. Thx. I'm going to merge it. I'm expecting the same PR for the master.

PR #195, opened 2015-07-13
Authnrequest is optional, do not require it

papegaaij:
This check for 'InResponseTo' breaks IDP initiated signon: there is no Authnrequest when using IDP initiated signon. Simply removing the check allows the signon to proceed.


leleuj:
No problem with this change: did you successfully test it?

papegaaij:
Yes, IDP initiated signon fails on this check. I'm now using a copy of this class, with this check removed, and it works fine.

leleuj:
Perfect. Thx.

Issue #165, opened 2015-05-17
Missing checks in Saml2ResponseValidator : inResponseTo, Destination, AuthnContext


elklaaso:
I have compared Saml2ReponseValidator with spring security saml's WebSSOProfileConsumerImpl and found some missing checks, of most i found TODO's in the pac4j Saml2Response Validator, the TODO's in there are:

validateSamlProtocolResponse: add Destination and inResponseTo Validation
isValidBearerSubjectConfirmationData: Validate inResponseTo
validateAuthenticationStatements: implement authnContext validation

Do you have any idea when and if you will implement these?

To what degree does this effect how secure pac4j saml is? For example, are there other checks that make these checks not so important?

thank you!

elklaaso:
And is there any way to replace the Saml2ReponseValidator with my own without modifying the code, i cant change the responseValidator in SAML2Client, because there is no getter used which i can override.

miremond:
Hi,

You are really relevant in your analysis. There are some checks not
implemented yet because when I wrote pac4j-saml, it was to bring a first
working version of SAMLv2 to pac4j.
I omitted these checks because they are not the most critical to me. They
just enforce additional security checks to be more confident about the SAML
assertion.

The class Saml2ResponseValidator can not currently be overriden as it is an
internal class but feel free to submit a pull request to implement the
checks or introduce a customization mechanism.

Regards,

Michaël

2015-05-17 17:13 GMT+02:00 elklaaso <notifications@github.com>:

leleuj:
Maintaining all `pac4j` projects is a tremendous: any contribution will be welcome.

elklaaso:
Hi,

I have made a simple implementation of response.inresponseto / response.getdestination and subject inresponseto validation. I got access to the private properties of Saml2Client using java reflect. I know this is bad, but i dont have time to download and rebuild this project etc.

The files are here:
https://www.dropbox.com/s/03q59bkdbhgpk0q/EnhancedSaml2Client.java?dl=0
https://www.dropbox.com/s/9iiseejatw8j1a4/EnhancedSaml2ResponseValidator.java?dl=0

Its a really easy implementation, i'm not sure if its how its supposed to happen but maybe you guys can use it as a starting point.

leleuj:
OK. Thanks for sharing. Maybe it can give @miremond some ideas...

mmoayyed:
This is now done via the move to opensaml. If you find other missing checks, please let me know.

[...]

PR #196, opened 2015-07-13
Authnrequest is optional, do not require it

papegaaij:
Same as #195 but on 1.7.x

This check for 'InResponseTo' breaks IDP initiated signon: there is no Authnrequest when using IDP initiated signon. Simply removing the check allows the signon to proceed.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
@@ -204,10 +204,6 @@ public class SAML2DefaultResponseValidator implements SAML2ResponseValidator {
             throw new SAMLException("Response issue instant is too old or in the future");
         }
 
-        if (response.getInResponseTo() == null) {
-            throw new SAMLException("InResponseTo field cannot be empty");
-        }
-
         AuthnRequest request = null;
         final SAMLMessageStorage messageStorage = context.getSAMLMessageStorage();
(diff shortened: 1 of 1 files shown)


mmoayyed:
This probably would work better if the client was configured for idp-initiated logon. The check is required for SP initiated. So this commit sort of breaks that rule.

papegaaij:
What do you mean with 'was configured for idp-initiated logon'? The client works fine with IDP initiated logon, just skip the initial authnrequest, and start with getting the credentials from the response. Some parts of the client (such as the keystore configuration) are only used for the authnrequest, but they do not hinder IDP initiated signon. In our case, we want to support both.

mmoayyed:
configured, meaning the client was explicitly told to operate in idp initiated mode. what you have here works, but now the sp initiated mode is sort of broken because a condition is taken out that is really required to be there.

mmoayyed:
@leleu I think we should revert this commit back, and think about how we can support all cases without losing additional checks.

mmoayyed:
https://github.com/pac4j/pac4j/issues/165

papegaaij:
As far as I know, the only difference in the SAML response between SP and IDP initiated signon is the presence (or absence) of the InResponseTo field. A simple boolean in the configuration to reject responses without InResponseTo should suffice. Note that in the spring version of this class, this requirement is disabled by default (meaning IDP initiated signon is supported by default).

mmoayyed:
OK. Thanks for the follow up. This seems to be an important check and I'd like to check with our OpenSAML experts to see what the appropriate approach is and how best we should handle this. Stay tuned.

mmoayyed:
I think we are ok without the check.

PR #197, opened 2015-07-14
remove jar plugin and change package type to bundle.

hboumedane:
with this correction you still have a jar file but this jar is an OSGI bundle as well


leleuj:
The build is broken because we still need to generate the `test-jar` artifact for other modules. Can you update your PR to keep it and see if the build passes this time?

hboumedane:
will correct that this afternoon

leleuj:
@hboumedane : shouldn't we do the same for all the other modules (`pac4j-cas`, `pac4j-oauth`, ...)?

hboumedane:
that would be the correct way, i noticed also that the kryo librairy is old and not osgi , so can i or you change the implementation of this librairy. because even if we produce pac4j-core as osgi it will not work because it has a non osgi dependency

leleuj:
Yes, we should upgrade the Kryo library as well. I think we should do that in another PR. Let's finish this one with the "bundle" update, open a new issue and push a new PR for the Kryo update (be careful to properly migrate the existing serializers: https://github.com/pac4j/pac4j/tree/master/pac4j-core/src/main/java/org/pac4j/core/kryo). Thx

pac4j-core/pom.xml
```diff
@@ -14,113 +14,114 @@
...
-		<dependency>
-			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-api</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>javax.servlet</groupId>
-			<artifactId>javax.servlet-api</artifactId>
-		</dependency>
+    <artifactId>pac4j-core</artifactId>
+    <packaging>jar</packaging>
```
leleuj: The packaging is not "bundle" any more, is it?

pac4j-core/pom.xml
```diff
@@ -14,113 +14,114 @@
...
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <!-- for testing -->
+    </dependencies>
 
-	<build>
```
leleuj: Please don't change the formatting, it's hard to see what are the real changes.

pom.xml
```diff
@@ -358,6 +367,21 @@
...
+							<phase>process-classes</phase>
+							<goals>
+								<goal>manifest</goal>
+							</goals>
+						</execution>
+					</executions>
+					<configuration>
+						<!-- configure plugin to support jar packaging -->
+						<supportedProjectTypes>
+							<supportedProjectType>jar</supportedProjectType>
```
leleuj: OK. I guess it's why we can keep "jar" for the packaging.

pac4j-core/pom.xml
leleuj: I guess there is no changes in this `pom.xml` now, isn't it?

leleuj:
I feel that the only changed file is the parent `pom.xml` now. Update the PR accordingly. Thx.

hboumedane:
the PR is already update , i don't understand what do you mean ?

leleuj:
If you take a look at the changes, you see that there are pretty clear differences for the `pom.xml` file.

Nonetheless, there are formatting (but no real) changes in the `pac4j-core/pom.xml file: if this file has not changed, it should not appear in the PR + only the differences should be visible (like for `pom.xml` file).

Does it make sense?

mmoayyed:
Could we also review the indentation policy here? This PR uses 6-spaces?

PR #200, opened 2015-07-16
generate osgi bundle - branch 1.7.x

hboumedane:


leleuj:
There is a problem with you PR (look at the changes and number of commits). You should not create it from your master, but from your own 1.7.x branch with the appropriate changes...

Issue #202, opened 2015-07-17
Provide a template ClientFactory


mmoayyed:
Presently, the clientsFactory requires a concrete implementation; it cannot be easily instantiated via Spring and DI. It would be better if the project provided a template class that implemented the interface and accepted a number of clients via the constructor. Once we have the constructors, we can call spring to invoke the build method.

Suppose we had a basic implementation of the interface that accepted a list of client objects as a ctor parameter. once the object is prepped, the build method can be called to return the final list of clients. this is useful to be able to instantiate and plug in all forms of clients via spring without having to modify the code.

Marking this for 1.8

PR #203, opened 2015-07-17
Create Spring-instantiable concrete ClientsFactory class

dbhankins:
Fixes https://github.com/pac4j/pac4j/issues/202

Create a concrete class ClientsFactoryDIImpl that implements ClientsFactory and provides a constructor taking a List<Client>. This list is returned as a Clients object when build() is called on the object, and the environment argument to build() is ignored.

Please also review @mmoayyed.


pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+package org.pac4j.core.client;
+
+import java.util.List;
+
+import org.pac4j.core.credentials.Credentials;
+import org.pac4j.core.profile.UserProfile;
+
+/**
+ * A Spring-initializable factory to build clients.
```
mmoayyed: The class is not spring specific. Please update the comment

pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,59 @@
...
+ * 
+ * @author Daniel Hankins
+ * @since 1.8.0
+ *
+ */
+public class ClientsFactoryDIImpl implements ClientsFactory {
```
mmoayyed: Doesn't seem like a very appropriate name. How about DefaultClientsFactory? DI and Impl are not very descriptive names.

pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    private Clients clientList;
+
+    /**
+     * Constructor that Spring can use to inject the client list to be returned
+     * by build()
+     * 
+     * @param clientList
+     */
+    public ClientsFactoryDIImpl(List<Client> clientList) {
```
mmoayyed: Missing final

pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Return the list of clients injected by Spring. This implementation
+     * ignores the env object, as the list has already been built by Spring.
+     * 
+     * @see org.pac4j.core.client.ClientsFactory#build(java.lang.Object)
+     */
+    @Override
+    public Clients build(Object env) {
```
mmoayyed: Missing final

pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,59 @@
...
+     * by build()
+     * 
+     * @param clientList
+     */
+    public ClientsFactoryDIImpl(List<Client> clientList) {
+        this.clientList = new Clients(clientList);
+    }
+
+    /**
+     * Return the list of clients injected by Spring. This implementation
```
mmoayyed: Also nothing specific about Spring in this class. While it can used by Spring, it's not specific to that use case.

pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,59 @@
...
+     * ignores the env object, as the list has already been built by Spring.
+     * 
+     * @see org.pac4j.core.client.ClientsFactory#build(java.lang.Object)
```
mmoayyed: @See not necessary.

pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,59 @@
...
+     * 
+     * @param clientList
+     */
+    public ClientsFactoryDIImpl(List<Client> clientList) {
+        this.clientList = new Clients(clientList);
+    }
+
+    /**
+     * Return the list of clients injected by Spring. This implementation
+     * ignores the env object, as the list has already been built by Spring.
```
mmoayyed: What env object are we referencing here?

pac4j-core/src/test/java/org/pac4j/core/client/TestClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,52 @@
...
+     * Test method for
+     * {@link org.pac4j.core.client.ClientsFactoryDIImpl}
+     * .
+     */
+    @Test
+    public final void testClientsFactoryDIImpl() {
+        Client clientA = new MockBaseClient("client a");
+        Client clientB = new MockBaseClient("client b");
+        Client clientC = new MockBaseClient("client c");
+        ArrayList<Client> clientList = new ArrayList<Client>();
```
mmoayyed: Left-hand side should always be the interface type.

mmoayyed: Also this should build with jdk7 conventions. So this would turn into ArrayList<>()

pac4j-core/src/test/java/org/pac4j/core/client/TestClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ */
+public class TestClientsFactoryDIImpl {
+    /**
+     * Test method for
+     * {@link org.pac4j.core.client.ClientsFactoryDIImpl}
+     * .
+     */
+    @Test
+    public final void testClientsFactoryDIImpl() {
```
mmoayyed: You might want to describe the test method

pac4j-core/src/test/java/org/pac4j/core/client/TestClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * This class tests the {@link ClientsFactoryDIImpl} class.
+ * 
+ * @author Daniel Hankins
+ * @since 1.8.0
+ *
+ */
+public class TestClientsFactoryDIImpl {
+    /**
```
mmoayyed: Javadoc not needed for test methods unless something peculiar is going on

pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
leleuj: I guess we could have a constructor with varags: `public ClientsFactoryDIImpl(Client... client)`as well.

pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
```diff
@@ -0,0 +1,59 @@
...
+    private Clients clientList;
+
+    /**
+     * Constructor that Spring can use to inject the client list to be returned
+     * by build()
+     * 
+     * @param clientList
+     */
+    public ClientsFactoryDIImpl(List<Client> clientList) {
+        this.clientList = new Clients(clientList);
```
leleuj: More important: I don't see any constructor with a `callbackUrl` input parameter, I think we should have two (one with a List, one with a varags) to follow all https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/client/Clients.java#L57 constructors.

leleuj:
Thanks for your contribution. My main comment is that you should have four constructors mapping to the four constructors of the `Clients` class.
The test class is a good thing.

I let you update the PR with all comments before I'll merge.

pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
dbhankins: I see three items final could apply to: the class, the instance variable, and the constructor parameter. I must be missing something, because I don't see why you'd want to declare any of those three final, with the possible exception of the constructor argument.

Why would you want to make the class non-extensible or immutable? I can envision scenarios in which you'd want to make a class that can be initialized by dependency injection, then have some of the clients removed because they don't fit the context you're using the factory in.
The same scenario applies to the member variable.

I mean, you might want to construct a second list of clients from the first meeting some criteria, then assign it to the instance variable. Not possible if it's final.

I can see making the constructor argument final, as that would prevent you from inadvertently reassigning the argument variable to something else. Is that what you meant?

pac4j-core/src/main/java/org/pac4j/core/client/ClientsFactoryDIImpl.java
dbhankins: What's missing final? The constructor argument? build()? The build() argument? Is this class meant to be non-extensible?

pac4j-core/src/test/java/org/pac4j/core/client/TestClientsFactoryDIImpl.java
dbhankins: "Client" _is_ the interface type.

Code compiles with JDK7 constraints without error in Eclipse. This compiles clean with JDK5 constraints in Eclipse.

leleuj:
+1

@mmoayyed : are we good for the merge?

mmoayyed:
Yup. Go ahead please. :+1:

PR #205, opened 2015-07-21
SQL support

leleuj:


pac4j-sql/src/main/java/org/pac4j/sql/credentials/authenticator/DbAuthenticator.java
```diff
@@ -0,0 +1,173 @@
...
+            } else if (results.size() > 1) {
+                throw new MultipleAccountsFoundException("Too many accounts found for: " + username);
+            } else {
+                final Map<String, Object> result = results.get(0);
+                final String expectedPassword = passwordEncoder.encode(credentials.getPassword());
+                final String returnedPassword = (String) result.get("password");
+                if (CommonHelper.areNotEquals(returnedPassword, expectedPassword)) {
+                    throw new BadCredentialsException("Bad credentials for: " + username);
+                } else {
+                    final DbProfile profile = createProfile(username, attributes.split(","), result);
```
mmoayyed: This will not work if fields are being aliased.

pac4j-sql/src/main/java/org/pac4j/sql/credentials/authenticator/DbAuthenticator.java
```diff
@@ -0,0 +1,173 @@
...
+            if (h != null) {
+                h.close();
+            }
+        }
+    }
+
+    protected DbProfile createProfile(final String username, final String[] attributes, final Map<String, Object> result) {
+        final DbProfile profile = new DbProfile();
+        profile.setId(username);
+        for (String attribute: attributes) {
```
mmoayyed: I think it will work better if get the key as the attribute name, from the map.

pac4j-sql/src/main/java/org/pac4j/sql/credentials/authenticator/DbAuthenticator.java
leleuj: Could you elaborate a little more on this?

pac4j-sql/src/main/java/org/pac4j/sql/credentials/authenticator/DbAuthenticator.java
leleuj: You mean loop on the `result` map and populate the `profile`. It will be the same result, won't it? What do you have in mind with "work better"?

pac4j-sql/src/main/java/org/pac4j/sql/credentials/authenticator/DbAuthenticator.java
mmoayyed: Your attributes in the sql query are basically just other extra fields. What you have here, I think, is  way to take those fields literally as they were defined, but that will not work if they are aliased. 

For example, what you have for attributes could be "firstname, lastname" and that would work. 

If I alias them in the query, which would "firstname AS F, lastname as L", that will not work when you do the parsing because you would get back F and L as names. So it's better to use the key for the attribute name off of the resulting map.

pac4j-sql/src/main/java/org/pac4j/sql/credentials/authenticator/DbAuthenticator.java
mmoayyed: See my comment above please about aliases.

pac4j-sql/src/main/java/org/pac4j/sql/credentials/authenticator/DbAuthenticator.java
leleuj: OK. I get it now. It was clear to me that fields need to be searched without any aliasing: the `attributes` `String` should be a list of attribute names separated by commas. So I can add a comment to explain that clearly.

But I guess your idea behind this comment is more about mapping the column names in the database into real field names, isn't it? If you want a specific mapping, the idea in `pac4j` has always been to create the appropriate profile with the appropriate mapping inheriting from `DbProfile` (in that case, the `createProfile` method should be overriden as well).

mmoayyed: Understood, but that seems like a pretty far fetched strategy to handle a simple mapping to me. I think the code should automatically handle this sort of thing, because it can do that easily without passing that on to the developer. That said, this may be an edge case.

leleuj: No, you're right: it seems pretty complicated to create a new user profile. That said, it's consistent with the entire library.
I'm not sure about the right solution, let's keep things that way and merge. We'll see that the requests will be on that topic before the release.

leleuj:
@mmoayyed + @miremond : Main REST supports have been added: LDAP + DB. I'll work later on Mongo, maybe Stormpath. I will start updating the implementations with pac4j v1.8 starting by j2e-pac4j and play-pac4j.

mmoayyed:
Sounds good. We do have a stormpath authenticator in the cas-addons module that might be useful for you. 

https://github.com/Unicon/cas-addons/blob/master/src/main/java/net/unicon/cas/addons/authentication/handler/StormpathAuthenticationHandler.java

leleuj:
Yes, I remember. Thanks for the link.

Issue #194, opened 2015-07-12
OidcProfile is not Serializable


alkemist:
This field is not serializable: https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/OidcProfile.java#L35

mmoayyed:
And it should be serializable because...? Are you seeing a particular error?

alkemist:
Because it implements java.io.Serializable by extending UserProfile.

This prevents it from being used in a distributed session.

leleuj:
Indeed, all user profiles must be serializable because user profiles always go into a web session or a cache and that requires Java serialization.
In CAS, as the attributes of the user profiles are used to built the principal (part of the main `Authentication` object), it must be serializable via Kryo when saved into Memcached.

mmoayyed:
Problem here is that the class BearerAccessToken is not part of pac4j. I'll post a solution shortly.

leleuj:
Thx

PR #211, opened 2015-07-27
Fix OIDC profile serialization issue

mmoayyed:
Fixes https://github.com/pac4j/pac4j/issues/194

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
@@ -15,6 +15,10 @@
  */
 package org.pac4j.core.profile;
 
+import java.io.Externalizable;
+import java.io.IOException;
+import java.io.ObjectInput;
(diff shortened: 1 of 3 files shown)


leleuj:
I was expecting a different solution only focused on the `OidcProfile` class where you would have implemented by hand the `writeObject` and `readObject` methods (to handle the `BearerAccessToken` serialization): isn't that possible?

mmoayyed:
I thought about that too; Some considerations:

1. I do prefer Externalizable, simply because it is an actual interface with clearly exposed methods and feels less magical than embedded write/read methods. It is also faster than native serialization which uses reflection. 

2. Even if I had gone that route, using write/read methods embedded, I still needed to find a way to externalize my super, and then realized if the profile was extended and a developer decided to stuff a field in the extended profile that was non serializable, they would have to duplicate code to also serialize super. Hence the method you see here.

leleuj:
Thanks for the explanations. It makes sense.

PR #214, opened 2015-07-28
improve bundles and upgrade kryo version

hboumedane:


leleuj:
Did you close this PR on purpose?

PR #217, opened 2015-07-29
Path fix for IdP metadata loading

### LINKED ISSUE #212: xml metadata file is always of type null
Hi,
I'm using pac4j 1.8 and i'm facing a hard issue.
I use the xml file provided with play-demo project to config my clients.

I'm getting this issue :  
```java
->java.lang.NullPointerException : null
 org.opensaml.saml.metadata.resolver.impl.DOMMetadataResolver.initMetadataResolver(DOMMetadataResolver.java:68)
 org.opensaml.saml.metadata.resolver.impl.AbstractMetadataResolver.doInitialize(AbstractMetadataResolver.java:191)
 net.shibboleth.utilities.java.support.component.AbstractInitializableComponent.initialize(AbstractInitializableComponent.java:61)
 org.pac4j.saml.metadata.SAML2IdentityProviderMetadataResolver.resolve(SAML2IdentityProviderMetadataResolver.java:95)
 org.pac4j.saml.client.SAML2Client.initIdentityProviderMetadataResolver(SAML2Client.java:183)
 org.pac4j.saml.client.SAML2Client.internalInit(SAML2Client.java:126)
 org.pac4j.core.util.InitializableObject.init(InitializableObject.java:37)
 org.pac4j.core.client.IndirectClient.getRedirectAction(IndirectClient.java:106)
 org.pac4j.core.client.IndirectClient.redirect(IndirectClient.java:85)
```

I noticed when i debug my app that this line of code in the file SAML2IdentityProviderMetadataResolver.java
```java
final Document inCommonMDDoc = Configuration.getParserPool().parse(in);
```

the inCommonMDDoc is always of type null, it has a refrence but the type is null so when my app try to unmarshall the the xml data i'm getting the exception above .

is this normal ?

mmoayyed:
Fixes #212

* Make sure filesystem resources are used as a fallback option when loading metadata.
* Explain failures with a better error message.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2IdentityProviderMetadataResolver.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2IdentityProviderMetadataResolver.java
@@ -32,6 +32,8 @@ import org.pac4j.core.exception.TechnicalException;
 import org.pac4j.core.util.CommonHelper;
 import org.pac4j.saml.exceptions.SAMLException;
 import org.pac4j.saml.util.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.core.io.ClassPathResource;
 import org.springframework.core.io.FileSystemResource;
 import org.springframework.core.io.UrlResource;
@@ -39,6 +41,7 @@ import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 
 import javax.annotation.Nullable;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Iterator;
@@ -49,6 +52,7 @@ import java.util.Iterator;
  * @since 1.7
  */
 public class SAML2IdentityProviderMetadataResolver implements SAML2MetadataResolver {
+    protected final Logger logger = LoggerFactory.getLogger(getClass());
 
     private final String idpMetadataPath;
     private String idpEntityId;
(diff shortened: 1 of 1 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2IdentityProviderMetadataResolver.java
```diff
@@ -71,27 +75,31 @@ public final MetadataResolver resolve() {
                     path = "/" + path;
                 }
                 resource = ResourceHelper.of(new ClassPathResource(path));
-            } else if (this.idpMetadataPath.startsWith("file:")) {
-                resource = ResourceHelper.of(new FileSystemResource(this.idpMetadataPath));
-            } else if (this.idpMetadataPath.startsWith("http")) {
+            }  else if (this.idpMetadataPath.startsWith("https")) {
```
leleuj: Why do you check on `https` (instead of `http`)?

mmoayyed: Metadata needs to come from an https source, so that we ca have a bit of extra security via SSL. A URL metadata typically comes from an SP endpoint itself, and the SP must be protected via https. If your app isn't secure, it's really not worth SSO'ing into :)

leleuj: You're right: HTTPS is the foundation of security. But what about the development environment and local tests? I guess it would be better for this kind of scenario to allow `http`:  don't you think?

mmoayyed: Good point. I think we might abstract out the loading of resources, to a different component. So that by default, there is a MetadataResourceLoader that does what we have today, and we can then pass in other impls of that component that allow for local dev/testing? perhaps a mocked one?

leleuj: I would go for the easiest solution: allow HTTP and maybe add a warning if the url is in HTTP, but your proposal makes sense as well even if it's more work.

mmoayyed: Sounds good. Let's do my solution in a separate pull then. I'll add back support for http with a warning.

mmoayyed: Done

leleuj:
Thx

PR #220, opened 2015-08-12
Added request/response to web context

### LINKED ISSUE #219: SAML support is broken due to the use of the J2EContext
Indeed, it only works in frameworks using the `J2EContext`, for others (like Sparkjava for which I just did the test), I get a `ClassCastException`.

The issue is here: https://github.com/pac4j/pac4j/blob/master/pac4j-saml/src/main/java/org/pac4j/saml/context/SAML2ContextProvider.java#L104

mmoayyed:
Closes #219

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/context/J2EContext.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/context/J2EContext.java
@@ -182,4 +182,14 @@ public class J2EContext implements WebContext {
             return requestURL.append('?').append(queryString).toString();
         }
     }
+
+    @Override
+    public HttpServletRequest getNativeRequest() {
+        return this.request;
+    }
+
+    @Override
+    public HttpServletResponse getNativeResponse() {
+        return this.response;
+    }
 }
--- a/pac4j-core/src/main/java/org/pac4j/core/context/J2ERequestContext.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/context/J2ERequestContext.java
@@ -18,6 +18,7 @@ package org.pac4j.core.context;
 import java.util.Map;
 
 import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 
 /**
  * This implementation uses the J2E request.
@@ -29,8 +30,12 @@ public class J2ERequestContext extends BaseResponseContext {
 
     private final HttpServletRequest request;
 
-    public J2ERequestContext(final HttpServletRequest request) {
+    private final HttpServletResponse response;
+
+    public J2ERequestContext(final HttpServletRequest request,
+                             final HttpServletResponse response) {
         this.request = request;
+        this.response = response;
     }
 
     /**
@@ -136,6 +141,16 @@ public class J2ERequestContext extends BaseResponseContext {
         } else {
             return requestURL.append('?').append(queryString).toString();
         }
+    }
+
+    @Override
+    public HttpServletRequest getNativeRequest() {
+        return this.request;
+    }
+
+    @Override
+    public HttpServletResponse getNativeResponse() {
+        return this.response;
     }
 
 }
--- a/pac4j-core/src/main/java/org/pac4j/core/context/WebContext.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/context/WebContext.java
@@ -15,6 +15,8 @@
  */
 package org.pac4j.core.context;
 
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import java.util.Map;
 
 /**
@@ -31,14 +33,14 @@ public interface WebContext {
      * @param name name of the parameter
      * @return the request parameter
      */
-    public String getRequestParameter(String name);
+     String getRequestParameter(String name);
 
     /**
      * Return all request parameters.
      *
      * @return all request parameters
      */
-    public Map<String, String[]> getRequestParameters();
+     Map<String, String[]> getRequestParameters();
 
     /**
      * Return a request attribute.
@@ -46,7 +48,7 @@ public interface WebContext {
      * @param name the name of the attribute
      * @return the attribute
      */
-    public Object getRequestAttribute(String name);
+     Object getRequestAttribute(String name);
 
     /**
      * Save a request attribute.
@@ -54,7 +56,7 @@ public interface WebContext {
      * @param name the name of the attribute
      * @param value the attribute
      */
-    public void setRequestAttribute(String name, Object value);
+     void setRequestAttribute(String name, Object value);
 
     /**
      * Return a request header.
@@ -62,7 +64,7 @@ public interface WebContext {
      * @param name name of the header
      * @return the request header
      */
-    public String getRequestHeader(String name);
+     String getRequestHeader(String name);
 
     /**
      * Save an attribute in session.
@@ -70,7 +72,7 @@ public interface WebContext {
      * @param name name of the session attribute
      * @param value value of the session attribute
      */
-    public void setSessionAttribute(String name, Object value);
+     void setSessionAttribute(String name, Object value);
 
     /**
      * Get an attribute from session.
@@ -78,35 +80,35 @@ public interface WebContext {
      * @param name name of the session attribute
      * @return the session attribute
      */
-    public Object getSessionAttribute(String name);
+     Object getSessionAttribute(String name);
 
     /**
      * Return the request method.
      *
      * @return the request method
      */
-    public String getRequestMethod();
+     String getRequestMethod();
 
     /**
      * Return the remote address.
      *
      * @return the remote address.
      */
-    public String getRemoteAddr();
+     String getRemoteAddr();
 
     /**
      * Write some content in the response.
      *
      * @param content content to write in response
      */
-    public void writeResponseContent(String content);
+     void writeResponseContent(String content);
 
     /**
      * Set the response status.
      *
      * @param code status code to set for the response 
      */
-    public void setResponseStatus(int code);
+     void setResponseStatus(int code);
 
     /**
      * Add a header to the response.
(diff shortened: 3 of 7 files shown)


leleuj:
Unfortunately, it won't work as the `HttpServletRequest` and `HttpServletResponse` do not exist in every framework.
Notice the `J2EContext` already provides access to the native request and response: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/context/J2EContext.java#L109

mmoayyed:
OK, that makes sense. How should we tackle the casting problem then?

leleuj:
We should not cast, but use the appropriate methods of the `WebContext`. Everywhere you need the HTTP request, pass the web context instead and replace the `request.getXXX()` by `context.getXXY()`.

It seems the request is used by the `SAMLMessageStorage` and other related objects. For the `HttpSessionStorage`, use the appropriate methods: `context.getSessionAttribute` and `context.setSessionAttribute`.

mmoayyed:
Sure. I'll look into this more, thanks.

mmoayyed:
Though the question is: if a request object is not available in all platforms, how is this going to work eventually? We solve the casting issue, but the functionality is still broken because you nonetheless need a request to work with, right?

leleuj:
pac4j is meant for web applications: the request / response should always be available in all platforms. Practically, it is the case, through various objects.

mmoayyed:
OK, I guess I am confused by your recent comment where you said HttpServletRequest and HttpServletResponse do not exist in every framework. What am I missing? If they don't exist in every framework which prompted the original issue, then why are we moving methods from J2EContext to WebContext? hmmm...

leleuj:
Indeed, `HttpServletRequest` and `HttpServletResponse` do not exist in every framework, but similar objects exist to represent the HTTP request and response.

mmoayyed:
OK, all good. I'll see if I can cook up a new PR shortly.

mmoayyed:
>We should not cast, but use the appropriate methods of the WebContext. Everywhere you need the HTTP request, pass the web context instead and replace the request.getXXX() by context.getXXY().

This solution is not possible. We need access to the native request, because

```java
        final J2EContext j2EContext = (J2EContext) webContext;
        final SimpleRequestAdapter inTransport = new SimpleRequestAdapter(j2EContext);
        final MessageContext<SimpleRequestAdapter> inCtx = new MessageContext<SimpleRequestAdapter>();
        inCtx.setMessage(inTransport);

...
public class SimpleRequestAdapter extends HttpServletRequestWrapper {
    private J2EContext context;

    public SimpleRequestAdapter(final J2EContext request) {
        super(request.getRequest());
    }
```

The underlying opensaml code that prepares the message context expects a request object and a response object. SimpleRequestAdapter  is a native request by implementing the request-wrapper, and in order for it to do so, it needs access to a native request. So we have to cast.

mmoayyed:
The only way you could get around the casting problem, is, if you move the native request/response accessors from J2E to the WebContext, but that is going to break other things. So, we might just accept this as "this is how it works. you need both a request and response".

leleuj:
It means you need to go further on the use of the web context objects: how the `SimpleRequestAdapter` is used and how can you update it or its caller to use the web context instead and no native request / response?

The native request / response will never go into the web context: the SAML support would be unavailable for most implementations. At worst, I prefer to revert the new SAML support and replace it by the old one (which had no dependency on the native request / response), but I'm pretty sure there is a way to make it work.

mmoayyed:
It's not an issue with SimpleRequestAdapter  per se. The underlying opensaml code expects a native request to work with. It wants from us an HttpServletRequest and it's based on stack, at least the version that we run. There is no way around it. You cannot change opensaml not to do that, and you cannot "pass" a fake request object unless you start to implement the entire request interface in WebContext, which would be futile. 

Also, note that reverting to a dead version of the software for what could be considered a marginal use case would not optimal. You will eventually run into issues with support, features, and possibly security updates that are not issued because the project is EOL. Reverting would also mean that CAS will not be able to use any of SAML support in Pac4j. So cons outweigh the pros entirely.

If a framework does not have access to a native j2e request, have "them" fake it instead.

mmoayyed:
Also, a few other things: The SAML support presented here supports the Web SSO Browser Profile. As the name suggests, it's web based, there requests and responses involved in a web environment. So the adopter would have to assume and accept the fact that in order for SAML support to work, the code would be used in a web application based on the J2E stack. Otherwise, without changing opensaml, I am not sure if we can make this work in ALL environments. 

I'll take a look anyway and see if something can be fixed, but I am not optimistic at all.

mmoayyed:
Example what would happen:

```
	at org.pac4j.saml.sso.impl.SAML2WebSSOMessageSender.sendMessage(SAML2WebSSOMessageSender.java:114)
	at org.pac4j.saml.sso.impl.SAML2WebSSOMessageSender.sendMessage(SAML2WebSSOMessageSender.java:46)
	at org.pac4j.saml.sso.impl.SAML2WebSSOProfileHandler.send(SAML2WebSSOProfileHandler.java:50)
	at org.pac4j.saml.sso.impl.SAML2WebSSOProfileHandler.send(SAML2WebSSOProfileHandler.java:34)
	at org.pac4j.saml.client.SAML2Client.retrieveRedirectAction(SAML2Client.java:233)
	at org.pac4j.core.client.IndirectClient.getRedirectAction(IndirectClient.java:123)
	at org.pac4j.core.client.IndirectClient.redirect(IndirectClient.java:85)
	at org.pac4j.saml.client.PostSAML2ClientIT.getRedirectionPage(PostSAML2ClientIT.java:56)
	at org.pac4j.core.client.ClientIT.testAuthenticationAndUserProfileRetrieval(ClientIT.java:94)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at junit.framework.TestCase.runTest(TestCase.java:176)
	at junit.framework.TestCase.runBare(TestCase.java:141)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:129)
	at junit.framework.TestSuite.runTest(TestSuite.java:255)
	at junit.framework.TestSuite.run(TestSuite.java:250)
	at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:84)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:160)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:116)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:234)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:74)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:140)
Caused by: net.shibboleth.utilities.java.support.component.ComponentInitializationException: HTTP servlet response cannot be null
	at org.opensaml.messaging.encoder.servlet.AbstractHttpServletResponseMessageEncoder.doInitialize(AbstractHttpServletResponseMessageEncoder.java:64)
	at org.opensaml.saml.saml2.binding.encoding.impl.HTTPPostEncoder.doInitialize(HTTPPostEncoder.java:126)
	at net.shibboleth.utilities.java.support.component.AbstractInitializableComponent.initialize(AbstractInitializableComponent.java:61)
	at org.pac4j.saml.sso.impl.SAML2WebSSOMessageSender.sendMessage(SAML2WebSSOMessageSender.java:102)
	... 30 more
```

mmoayyed:
Another option is, pac4j writes its own Post Encoder/Decoder components for SAML.

leleuj:
I admit that reverting to an old opensaml implem is not a good solution, but it's not an edge case. Most pac4j implementations don't have a `HttpServletRequest` / `HttpServletResponse`.

Indeed, maybe some native opensaml component needs to copy/paste and the request/response replace by the web context. I'm confident we can find a solution on this, but even it's a lot harder than just implementing it for J2E, it has great benefits: working out of the box for 10 different frameworks.

mmoayyed:
OK, sure. I think I have pretty much narrowed it down to the encoders/decoders. If we write impls for those that are based on WebContext, we may be able to solve the problem entirely.

mmoayyed:
Testing. The TestShib Idp shows:

```
Relying party 'urn:mace:saml:pac4j.org' requested the response to be returned to endpoint with ACS URL 'http://localhost:8080/callback?client_name=SAML2Client'  and binding 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST' however no endpoint, with that URL and using a supported binding,  can be found in the relying party's metadata 
07:29:33.003 - ERROR [edu.internet2.middleware.shibboleth.idp.profile.AbstractSAMLProfileHandler:447] - No return endpoint available for relying party urn:mace:saml:pac4j.org
```

My metadata has:

```
<md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="http://localhost:8080/callback?client_name=SAML2Client" index="0"/>
```

Is it possible that the copy of the metadata testshib has for pac4j has changed?

PR #221, opened 2015-08-12
Upgrade test cas module to v3.6.0

mmoayyed:


leleuj:
Any reason for this upgrade to 3.6.0 (rather than directly to 4.0.0) except that it's straightforward ;-) ?

mmoayyed:
Yup, it was just easy to do! Upgrading to 4 is going to need a lot more effort.

PR #223, opened 2015-08-14
Update for the j2e-pac4j library v1.2

leleuj:


pac4j-core/src/main/java/org/pac4j/core/config/ConfigFactory.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @author Jerome Leleu
+ * @since 1.8.0
+ */
+public interface ConfigFactory {
+
+    /**
+     * Build a configuration.
+     *
+     * @return
```
mmoayyed: Maybe take this line out?

pac4j-core/src/main/java/org/pac4j/core/config/ConfigInstance.java
```diff
@@ -0,0 +1,48 @@
...
+ * the configuration must be shared across several elements.
+ *
+ * @author Jerome Leleu
+ * @since 1.8.0
+ */
+public class ConfigInstance {
+
+    protected static Config config = new Config();
+
+    public synchronized static void build(final String factoryName) {
```
mmoayyed: Is it ever possible that I would want multiple factories per instance of this class?

pac4j-core/src/main/java/org/pac4j/core/context/J2ERequestContext.java
```diff
@@ -79,9 +80,14 @@ public Object getSessionAttribute(final String name) {
     }
 
     /**
-     * Return the request method : GET, POST...
-     *
-     * @return the request method
+     * {@inheritDoc}
+     */
+    public void invalidateSession() {
+        this.request.getSession().invalidate();
```
mmoayyed: You might want to do a null check here.

pac4j-core/src/main/java/org/pac4j/core/config/ConfigFactory.java
leleuj: I will update the comment accordingly.

pac4j-core/src/main/java/org/pac4j/core/config/ConfigInstance.java
leleuj: No, generally, you only have one factory and one config instance per application (or none). Though, from a design POV, it would be better to have a class in charge of building configuration, distinct from the one that stores the configuration.

pac4j-core/src/main/java/org/pac4j/core/context/J2ERequestContext.java
leleuj: On the `request`? Because the `getSession()` method will automatically create a session and return a not null object.

leleuj:
Replace `ConfigInstance` by `ConfigSingleton` and `ConfigBuilder`. Slightly change the logic in `DefaultAuthorizerBuilder`.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
```diff
@@ -34,19 +34,16 @@ public ProfileManager(final WebContext context) {
...
-     * Retrieve the current user profile.
+     * Retrieve the current user profile (from request first and then from the session if not found and requested).
      *
      * @param readFromSession if the user profile must be read from session
      * @return the user profile
      */
     public U get(final boolean readFromSession) {
-        U profile = null;
-        if (readFromSession) {
+        U profile = (U) this.context.getRequestAttribute(Pac4jConstants.USER_PROFILE);
```
leleuj: Notice that now, the user profile is first retrieved from the request before being potentially retrieved from the session: in that case, a REST authentication (by JWT token) can override an authentication from an indirect client (like Facebook) if the session is not used by the direct client.

leleuj:
For: https://github.com/pac4j/j2e-pac4j/pull/18 & https://github.com/pac4j/j2e-pac4j-demo/pull/16

PR #224, opened 2015-08-16
Fix MetaDataResolver to read from file.

thesamet:
Prior to this change, FileSystemResource would get the `file:` prefix, which makes it unable to load files.


leleuj:
Yes, good point. It makes me think the problem exists also in the master. The "file:" string is not taken into account and should be clean if found. Here: https://github.com/pac4j/pac4j/blob/master/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2IdentityProviderMetadataResolver.java#L86

Issue: https://github.com/pac4j/pac4j/issues/225

Issue #227, opened 2015-08-20
Retrieve **Open ID Connect** "ID Token"


tombatron:
Greetings!

I'm currently leveraging `pac4j-oidc` to integrate an existing Java web application to an identity server that is based on [IdentityServer3](https://identityserver.github.io/Documentation/).

One of the requirements that I have is that a user is redirected to a specified URL upon logging out of their Open ID Connect session. In order to do this, one of the required parameters is the ID token of the user who is logging out. [Here](http://leastprivilege.com/2014/10/14/identityserver-v3-and-post-logout-redirect/) is the reasoning behind this requirement. 

That said, I need to access the ID token that is retrieved by the `OidcClient` while creating an instance of `OidcProfile`. While reviewing the source, I didn't see any obvious places where I could create a sub-class to introduce the desired functionality. 

Short of making another request for the token, is there something I can do to retrieve the ID token?

Thanks for your time,

Tom

leleuj:
Is the ID token retrieved by the `OidClient` saved in the `OidcProfile`? If so, you can get it using the `getAttribute` method. Otherwise, I think we could change the source code to save it into the profile.
What is exactly the ID token?

tombatron:
It doesn't look like the ID token is saved into the `OidcProfile`. 

From my reading of the source it looks like it is retrieved from the provider, validated, used to populate the claims for the profile, and the "discarded". 

Here is the snippet of code where I'm seeing that [OidcClient.java:310-322](https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/main/java/org/pac4j/oidc/client/OidcClient.java)

```
ReadOnlyJWTClaimsSet claimsSet = this.jwtDecoder.decodeJWT(tokenSuccessResponse.getIDToken());

if (useNonce()) {
    String nonce = claimsSet.getStringClaim("nonce");
    if (nonce == null || !nonce.equals(context.getSessionAttribute(NONCE_ATTRIBUTE))) {
        throw new TechnicalException(
                "A nonce was sent in the authentication request but it is missing or different in the ID Token. "
                        + "Session expired or possible threat of cross-site request forgery");
    }
}

// Return profile with Claims Set, User Info and Access Token
OidcProfile profile = new OidcProfile(accessToken);
profile.setId(claimsSet.getSubject());
profile.addAttributes(claimsSet.getAllClaims());
profile.addAttributes(userInfo.toJWTClaimsSet().getAllClaims());

return profile;
```

[...]

PR #228, opened 2015-08-20
"ID Token" Storage on `OidcProfile`

tombatron:
Proposed changes as discussed in issue #227.


pac4j-oidc/src/main/java/org/pac4j/oidc/profile/OidcProfile.java
```diff
@@ -40,6 +43,8 @@
     private static final long serialVersionUID = -52855988661742374L;
 
     private BearerAccessToken accessToken;
+    private String idTokenString;
+    private JWT idToken;
```
mmoayyed: Is this serializable?

tombatron: Oh man. 

It wasn't. I just updated the pull request with the `writeExternal` and `readExternal` methods updated to handle the `idTokenString` property. 

My intention with the `idToken` property was to have it lazy loaded.

leleuj:
Looks good to me.

PR #231, opened 2015-08-24
Allow for arbitrary attributes on a profile

leleuj:


pac4j-core/src/main/java/org/pac4j/core/profile/AttributesDefinition.java
```diff
@@ -106,7 +106,11 @@ public Object convert(final String name, final Object value) {
         if (converter != null && value != null) {
             return converter.convert(value);
         } else {
-            return null;
+            if (ProfileHelper.isEnforceProfileDefinition()) {
```
leleuj: Here is the real change.

PR #244, opened 2015-08-31
Deprecate BasicAuthClient

leleuj:


pac4j-http/src/main/java/org/pac4j/http/client/BasicAuthClient.java
```diff
@@ -0,0 +1,25 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+package org.pac4j.http.client;
+
+import org.pac4j.http.client.indirect.IndirectBasicAuthClient;
+
+/**
+ * Will be removed. Use {@link IndirectBasicAuthClient} instead.
```
mmoayyed: Should also add the @deprecated tag here to the javadocs

mmoayyed: Also, is this new class that is being added or is it moved? If it's moved, I am not sure you gain much from deprecating it because that will break BC.

leleuj: Indeed, I will add the `@deprecated` in the javadocs.

leleuj: This was an old class I first moved to another package and changed the behaviour radically. As I realize it is a bad idea, I recreate it empty, inheriting from the class which replaces it, to help users find the new non-deprecated class they should use.
We now have a `DirectBasicAuthClient` for stateless use cases and a `IndirectBasicAuthClient` for stateful scenarios.

Does it make sense?

mmoayyed: Ah I see. Ok, sure. :+1:

PR #246, opened 2015-08-31
Check the unicity of the clients' name in the Clients object

leleuj:


pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -73,28 +74,32 @@ public Clients(final Client... clients) {
...
-                    // has indicated to not include it.
-                    if (indirectClient.isIncludeClientNameInCallbackUrl() &&
-                            indirectClientCallbackUrl.indexOf(this.clientNameParameter + "=") < 0) {
-                        indirectClient.setCallbackUrl(CommonHelper.addParameter(indirectClientCallbackUrl, this.clientNameParameter,
-                                indirectClient.getName()));
-                    }
+        final HashSet<String> names = new HashSet<>();
+        for (final Client client : this.clients) {
+            final String name = client.getName();
+            if (names.contains(name)) {
```
mmoayyed: wouldnt it better if you handled this logic inside the `Clients` instead? Why should be I be allowed to configure duplicate client names in Clients, only to be errored out later?

leleuj: I'm not sure to understand what you mean by " handled this logic inside the Clients instead". By default, client names are the classes names so if you have twice the same client, you will have the same name twice which would lead to unpredictable behaviours.
The idea here is to prevent that.

mmoayyed: `Clients` gets configured somewhere right? These class names that get put into Clients are just Strings? 

My idea is to ensure that Clients prevents duplicates as opposed to manually doing it here.

leleuj: The `Clients` is built from a callback url and a list of clients, but I fear that users may have twice the same clients (and indeed the same names) without noticing. At the time of creation, I don't see how the user would check the names (which are not defined, but inferred from the clients).
Example: https://github.com/pac4j/j2e-pac4j-demo/blob/master/src/main/java/org/pac4j/demo/j2e/config/DemoConfigFactory.java#L27

mmoayyed: Ah, ok. Understood :) :+1:

PR #249, opened 2015-09-02
Remove isAjax

leleuj:


pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
```diff
@@ -260,4 +264,12 @@ public void setCallbackUrl(final String callbackUrl) {
     public String getCallbackUrl() {
         return this.callbackUrl;
     }
+
+    public AjaxRequestResolver getAjaxRequestResolver() {
+        return ajaxRequestResolver;
+    }
+
+    public void setAjaxRequestResolver(AjaxRequestResolver ajaxRequestResolver) {
+        this.ajaxRequestResolver = ajaxRequestResolver;
```
mmoayyed: Is a default resolver set somewhere?

leleuj: Yes, line 59: https://github.com/pac4j/pac4j/pull/249/files#diff-d3bf314104b3efbdbf8ec39f13e4a6ddR59

leleuj:
Assuming everybody is ok with that change, I will merge in a few hours...

PR #252, opened 2015-09-03
The AuthenticatorProfileCreator should be defined by default

leleuj:


leleuj:
Will merge unless someone has a concern.

Issue #250, opened 2015-09-03
Make Credentials and UserProfile? cleanable


leleuj:
For Spring Security at first.

See: https://github.com/pac4j/spring-security-pac4j/pull/13

JacobASeverson:
Just opened PR https://github.com/pac4j/pac4j/pull/254 . A few things to note:
* I wasn't sure where you would want `Clearable` in the package structure since it is used by both `UserProfile` and `Credentials`. Let me know if you think it should go somewhere else.
* Looking through the `UserProfile` implementations, I wasn't sure what should be cleared for most of them. The only Implementation I added was for the `OAuth20Profile` type.

JacobASeverson:
Just pushed a commit to the PR with those last changes we were discussing, let me know if there is anything else you can think of. Thanks!

leleuj:
Just merged. Thanks again.

PR #254, opened 2015-09-03
Added the Clearable abstraction and applied to Credentials and User Profile per #250

JacobASeverson:
A Clearable abstraction will be useful for a types such as Credentials and UserProfile that may be
storing sensitive information. This new abstraction will allow implementers to clear out any sensitive
or unneeded information once it is no longer needed after authentication.

* Added the Clearable type

* Added Clearable to Credentials and implemented/test it for all subtypes

* Added Clearable to UserProfile and implemented/tested for OAuth 2

### DIFF
--- a/pac4j-http/src/main/java/org/pac4j/http/credentials/TokenCredentials.java
+++ b/pac4j-http/src/main/java/org/pac4j/http/credentials/TokenCredentials.java
@@ -27,7 +27,7 @@ public class TokenCredentials extends HttpCredentials {
 
     private static final long serialVersionUID = -4270718634364817595L;
 
-    private final String token;
+    private String token;
 
     public TokenCredentials(String token, final String clientName) {
         this.token = token;
@@ -41,5 +41,12 @@ public class TokenCredentials extends HttpCredentials {
     @Override
     public String toString() {
         return CommonHelper.toString(this.getClass(), "token", this.token, "clientName", getClientName());
+    }
+
+    @Override
+    public void clear() {
+        this.token = null;
+        this.setClientName(null);
+        this.setUserProfile(null);
     }
 }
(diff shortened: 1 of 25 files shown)


pac4j-http/src/main/java/org/pac4j/http/credentials/TokenCredentials.java
```diff
@@ -42,4 +42,11 @@ public String getToken() {
     public String toString() {
         return CommonHelper.toString(this.getClass(), "token", this.token, "clientName", getClientName());
     }
+
+    @Override
+    public void clear() {
+        this.token = null;
+        this.setClientName(null);
+        this.setUserProfile(null);
```
leleuj: If ever we think we need to clear the user profile, you should call the `clear` method of the user profile.

leleuj: In fact, my suggestion might not be a good idea. The user profile saved into this credentials is passed this way between an `Authenticator` and a `ProfileCreator` to eventually return it.
So clearing the credentials would clear the user profile, which we may not want.

leleuj:
It's an excellent work. Thanks again.

I'm questioning myself: what's the real idea behind clearing credentials / user profiles? Removing critical information from the memory to avoid these information from being stolen?

leleuj:
Thinking a little more about this, I stay with the idea that the `clear` concept should target critical information enabling authentication replay from attackers: in that case, it targets username/password in credentials, but also access token or PGT in user profiles.

I also think that clearing credentials should be independent from clearing user profiles. So I would keep the pull request as is except that the https://github.com/pac4j/pac4j/blob/master/pac4j-cas/src/main/java/org/pac4j/cas/client/rest/HttpTGTProfile.java should also clear its `ticketGrantingTicketId` and the https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/OidcProfile.java should also clear its `accessToken`.

Thanks

JacobASeverson:
Apologies, I went on vacation for the weekend - I'll make the recommended changes either tonight or tomorrow. Thanks!

leleuj:
No problem.

PR #255, opened 2015-09-03
DefaultClientFinder

leleuj:


leleuj:
Will merge unless someone has a concern. It will be easy to understand this change when I'll update the implementations.

mmoayyed:
:+1:

Issue #257, opened 2015-09-04
SAML Client name not properly set in SAML Credentials


jkacer:
In PAC4J 1.7.1, the name of the SAML Client in the credentials is a hard-coded string ("SAML2Client"), not the real name of the SAML Client. This causes major problems when more SAML Clients are used together; for example, I have about 10 of them in my CAS.

When a SAML response arrives from an IdP server, CAS/PAC4J is not able to assign the right SAML client to the SAML credentials extracted from the response because the hard-coded client name "SAML2Client" does not match any real client name. The PAC4J ClientAction in CAS is not hen able to finish authentication and throws an error.

I have analyzed the differences between 1.6.0 and 1.7.1. In 1.6.0, the credentials are constructed directly in the client:
```
ClientAction.doExecute()
BaseClient.getCredentials()
Saml2Client.retreiveCredentials()
Saml2Client.buildSaml2Credentials()
Saml2Client.getName() - this returns the right name
```

In 1.7.1, it's quite different:
```
ClientAction.doExecute()
BaseClient.getCredentials()
SAML2Client.retreiveCredentials() - the last point we know the right name
SAML2WebSSOProfileHandler.receive()
SAML2WebSSOMessageReceiver.receiveMessage()
SAML2DefaultResponseValidator.validate()
SAML2DefaultResponseValidator.buildSAML2Credentials() - sets constant name "SAML2Client"
```

jkacer:
Affected version: 1.7.1, maybe 1.7.0 too.
I'll try to provide a fix as a pull request into 1.8.0.

jkacer:
Hi, pull request created:
https://github.com/pac4j/pac4j/pull/258

Some unit tests are failing but so were they before the fix.

PR #258, opened 2015-09-04
PAC4J issue #257

jkacer:
SAML Client name is now properly set in SAML Credentials. Needed when a
SAML response arrives and it must be assigned to the right SAML client.


mmoayyed:
:+1:

Issue #256, opened 2015-09-04
SAML Client Configuration: Maximum Authentication Lifetime must always be set (v. 1.7.1)


jkacer:
It seems that the **maximumAuthenticationLifetime** property must now be set on the SAML Client Configuration object. If it is not set, the default value of 3600 seconds in SAML2DefaultResponseValidator is not used; it is always overwritten in the constructor with the value coming from SAML2Client.initSAMLResponseValidator().

This causes problems when the IdP server remembers authentication for some time and does not ask the user to provide his credentials again. Then the assertion sent to the client has an older Authentication Instant, which does not pass validation because maximumAuthenticationLifetime is 0.

It would be good to distinguish whether the property has really been set or not and use the default value in the validator if it hasn't.

jkacer:
Affected version: 1.7.1, maybe 1.7.0 too.
I'll try to provide a fix as a pull request.

leleuj:
Please provide a patch for the master (version 1.8.0) as no new version 1.7.x will be released.

jkacer:
Aha, good to know, thanks. Will do so.

jkacer:
Hi, pull request created:
https://github.com/pac4j/pac4j/pull/259

Some unit tests are failing but so were they before the fix.

PR #259, opened 2015-09-04
PAC4J issue #256

jkacer:
The maximum authentication lifetime passed to the constructor is only
used if it is positive. Otherwise, the default value is used instead.
This avoids situations where zero is used for the lifetime instead of
the default value because a SAML Client Configuration object has no
explicit value set.


leleuj:
Thanks

PR #260, opened 2015-09-06
Deprecate requireAnyRole and requireAllRoles

leleuj:


pac4j-core/src/main/java/org/pac4j/core/authorization/RequireAllRolesAuthorizer.java
```diff
@@ -27,28 +29,46 @@
  */
 public class RequireAllRolesAuthorizer<U extends UserProfile> implements Authorizer<U> {
 
-    private final String[] expectedRoles;
+    private String[] roles;
```
mmoayyed: Why not use a List?

mmoayyed: Or better yet, a Set?

pac4j-core/src/main/java/org/pac4j/core/authorization/RequireAnyRoleAuthorizer.java
```diff
@@ -27,28 +29,46 @@
  */
 public class RequireAnyRoleAuthorizer<U extends UserProfile> implements Authorizer<U> {
 
-    private final String[] expectedRoles;
+    private String[] roles;
```
mmoayyed: See my comment above please

pac4j-core/src/main/java/org/pac4j/core/authorization/RequireAllRolesAuthorizer.java
leleuj: No specific reason. A `Set` would be better indeed.

leleuj:
Updated according to @mmoayyed 's feedbacks.

mmoayyed:
:+1:

PR #261, opened 2015-09-07
SAML support is broken due to the use of the J2EContext

mmoayyed:
Closes https://github.com/pac4j/pac4j/issues/219


mmoayyed:
So this pull removes the use of J2EContext in favor of WebContext. For SAML, additional encoders/decoders are provided that work based on WebContext rather than the native http request and response. The WebContext itself is modified to handle response-related functions a bit more. All tests should pass for SAML. (I ended up updating our metadata on testshib, by the filename `pac4j-sp-metadata.xml`)

leleuj:
Great! Do you want me to test it or just review the source code?

mmoayyed:
It would be great if you could do both. Certainly testing it with the demo app you had that demonstrated the problem would be nice.

leleuj:
:+1:

I'll make a real test tomorrow.

leleuj:
Successfully tested!

mmoayyed:
parfait!

PR #262, opened 2015-09-08
No credentials -> null

leleuj:


leleuj:
Merging (tiny update)...

Issue #247, opened 2015-09-01
Create a LocalCachingAuthenticator


leleuj:
For stateless use cases, the `Authenticator` may be called at each request. So it might be necessary to cache locally the result for the given credentials, to avoid any performance issue.

mmoayyed:
How would you capture the result, if you have for example:

```
public class LocalCachingAuthenticator<T extends Credentials> implements Authenticator<T>
```

mmoayyed:
assuming the authentication is delegated to something else. for example:

```
public class LocalCachingAuthenticator<T extends Credentials> implements Authenticator<T> {

    private Map<String, Object> cache;

    private final Authenticator<T> delegate;

    public LocalCachingAuthenticator(final Authenticator<T> delegate) {
        this.delegate = delegate;
    }

    @Override
    public void validate(final T credentials) {
        this.delegate.validate(credentials);

    }
}
```

leleuj:
There might be an issue in the definition of the `DirectHttpClient` and `IndirectHttpClient`: both should have credentials inheriting from `HttpCredentials` (which may store a user profile). `TokenCredentials` and `UsernamePasswordCredentials` both inherit from `HttpCredentials`.

So I guess the cache should be a `Map<HttpCredentials, UserProfile>`: if no exception occurs, the credentials is saved with the associated user profile.

leleuj:
Or maybe a `Map<HttpCredentials, Object>` would be enough where `Object` is `null` or another meaningless value. And the `equals` and  `hash` methods of the `HttpCredentials` would take into account the credentials values and the user profile identifier.

leleuj:
@mmoayyed : in fact, there is an issue.

Here is my test:

```
private class OnlyOneCallAuthenticator implements UsernamePasswordAuthenticator {
  private int n = 0;
  public void validate(final UsernamePasswordCredentials credentials) {
    if (n > 0) {
      throw new IllegalArgumentException("Cannot call validate twice");
    }
    credentials.setUserProfile(new UserProfile());
    n++;
  }
}

@Test
public void testDoubleCalls() {
  final OnlyOneCallAuthenticator authenticator = new OnlyOneCallAuthenticator();
  final LocalCachingAuthenticator localCachingAuthenticator = new LocalCachingAuthenticator(authenticator, 10, 10, TimeUnit.SECONDS);
  final HttpCredentials credentials1 = new UsernamePasswordCredentials("a", "a", this.getClass().getName());
[...]

PR #272, opened 2015-09-15
LocalCachingAuthenticator Impl

mmoayyed:
Closes https://github.com/pac4j/pac4j/issues/247

Let's review the design, and once that is consolidated, I'll add tests, etc.

### DIFF
--- a/pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/LocalCachingAuthenticator.java
+++ b/pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/LocalCachingAuthenticator.java
@@ -15,13 +15,16 @@
  */
 package org.pac4j.http.credentials.authenticator;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import org.pac4j.core.exception.TechnicalException;
 import org.pac4j.core.profile.UserProfile;
 import org.pac4j.http.credentials.HttpCredentials;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.HashMap;
-import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
 
 /**
  * An authenticator that caches the result of an authentication event locally.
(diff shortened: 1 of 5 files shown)


pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/LocalCachingAuthenticator.java
```diff
@@ -0,0 +1,71 @@
...
+ * An authenticator that caches the result of an authentication event locally.
+ * The authentication is delegated to a pluggable component.
+ * @author Misagh Moayyed
+ * @since 1.8
+ */
+public class LocalCachingAuthenticator<T extends HttpCredentials> implements Authenticator<T> {
+
+    protected final Logger logger = LoggerFactory.getLogger(getClass());
+
+    private final Map<T, UserProfile> cache;
```
leleuj: I was thinking about using a Guava cache, something like: `CacheBuilder.newBuilder().maximumSize(size).expireAfterWrite(timeout, TimeUnit.MINUTES).build();``

Because a simple `Map` will grow forever and can cause `OutOfMemoryError`...

mmoayyed: Sure. Is guava already available in pac4j?

mmoayyed: This reminds me actually; that a guava cache impl should also be designed for CAS when handling principal attributes

leleuj: Guava is not used by pac4j yet, but it can be added, it's a pretty common dependency.

mmoayyed: OK, I did a first try with CAS. Perfectly simple. Will post a similar pull here shortly.

mmoayyed: Updated pull

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/LocalCachingAuthenticator.java
```diff
@@ -0,0 +1,82 @@
...
+                                     final TimeUnit timeUnit) {
+        this.delegate = delegate;
+        this.cache = CacheBuilder.newBuilder().maximumSize(cacheSize)
+                            .expireAfterWrite(timeout, timeUnit).build();
+    }
+
+    @Override
+    public void validate(final T credentials) {
+        try {
+            final UserProfile profile = this.cache.get(credentials, new Callable<UserProfile>() {
```
leleuj: Thinking a little more about that, I'm not sure it works. How new credentials are compared to stored ones (with potential user profile)? Using the `equals` method?

The source code is a bit complicated. I was expecting something like:

```
if (cache.containsKey(credentials)) {
   profile = cache.get(credentials);
   credentials.setUserProfile(cache);
} else {
  delegate.validate(credentials);
}
```

Can you add a test to ensure the correct behaviour?

Thanks

mmoayyed: Sure, wanted to make sure we agree on the initial design before I add tests.

The API you have in your snippet cant exist with Guava. You get something from a cache, if it's there. If not, you make a call to a Callable to build the object and cache it.

leleuj: OK. Thanks. In that case, in your current algorithm, when do you put the data in the cache?

mmoayyed: I updated the pull to make the design a bit more explanatory. The idea is, there is `cacheLoader` that gets called by the cache itself when an item is not found. Items are cached by their hashcode. When cache fails to find an item,  via `final UserProfile profile = this.cache.get(credentials);` it will call on the cache loader to produce the object.

mmoayyed: There are test cases in the pull to verify the behavior.

pac4j-http/src/test/java/org/pac4j/http/credentials/authenticator/IpRegexpAuthenticatorTests.java
```diff
@@ -16,11 +16,9 @@
 package org.pac4j.http.credentials.authenticator;
 
 import org.junit.Test;
-import org.pac4j.core.context.MockWebContext;
```
mmoayyed: Removed a number of unused imports here

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/LocalCachingAuthenticator.java
```diff
@@ -0,0 +1,102 @@
...
+            this.delegate = delegate;
+        }
+
+        public UserProfile load(final T credentials) throws Exception {
+            logger.debug("Delegating authentication to {}...", delegate);
+            delegate.validate(credentials);
+            final UserProfile profile = credentials.getUserProfile();
+            logger.debug("Cached authentication result");
+            return profile;
+        }
```
mmoayyed: The loader simply calls the delegate to validate the credential and get the profile. It will automatically go into the cache.

leleuj:
Thanks

PR #273, opened 2015-09-15
Null credentials check

robgratz:
Check for null credentials before doing the validation check.  If null, require http action.


leleuj:
Thanks

PR #274, opened 2015-09-16
Replace AuthorizationBuilder by AuthorizationChecker

leleuj:


mmoayyed:
:+1:

leleuj:
@mmoayyed : thanks for taking time to review most of my contributions ;-)

I realized I forgot to post the main explanation about this change:

At first, I had in mind to define only one `Authorizer` for the `RequiresAuthenticationFilter`, but I realized than in most practical use cases, I may want to check multiple things like "being an admin, having the read permission", so a `String`:  `authorizerName`, being a list of authorizer names (separated by commas), would be better.

For example (in Play):

```
@RequiresAuthentication("FacebookClient", "roleadmin,readpermission")
```

With:

```
config.addAuthorizer("roleadmin", new RequireAnyRoleAuthorizer("ROLE_ADMIN"));
config.addAuthorizer("readpermission", new RequireAnyPermissionAuthorizer("read"));
```

Notice that the `RequireAnyPermissionAuthorizer` does not exist yet.

It can be used for more advanced use cases like CSRF protection. We could have a CSRF authorizer which checks if a provided token (header or parameter) is the same as the one saved into session. To take a really advanced example, we could have:

```
@RequiresAuthentication("FacebookClient", "csrf,denyframe,roleadmin,readpermission")
```

leleuj:
To achieve the best design, I'm trying to match advanced use cases and even if I won't commit anything yet, I want to ensure that pac4j is / seems to be compliant with them. So don't be surprised if I made some more changes.

mmoayyed:
Makes good sense. Thanks for the explanations.

leleuj:
Trying to address advanced use cases, I realized that the `webContext` does not provide any method related to cookies. I just added the following issue: https://github.com/pac4j/pac4j/issues/279

leleuj:
When trying to match advanced use cases, I realized that there could be multiple direct authentications while it doesn't make any sense to have more than one indirect authentication. We have similar algorithm in the CAS server where we can have several non interactive action and only one UI interaction.

So the `DefautlClientFinder` now returns a list of clients: https://github.com/pac4j/pac4j/pull/274/files#diff-b96b9ec49fc893afe0c425b05b03393aR54 and implementations try successively all direct clients: https://github.com/pac4j/j2e-pac4j/pull/20/files#diff-9586bb9ec1fa9f17f8772b43b5e9b27fR109 and https://github.com/pac4j/play-pac4j/pull/77/files#diff-87ee29ad93408aafea848ca0e61c627bR115

PR #275, opened 2015-09-17
Add equals / hashCode on credentials to make LocalCachingAuthenticator work

leleuj:


mmoayyed:
Looks great. Do we have apache commons to generate equals and hashCode? 

I also think we should enforce these two methods at the Credentials level. So extensions are forced to add them. It's too easy to forget and the compiler does not care.

leleuj:
The `pac4j-http` module already depends on Guava and commons-codec, so I guess we could add the dependency on commons.

How do you want to enforce these two methods (I'm thinking of subclasses)? In `HttpCredentials` you mean?

mmoayyed:
Yes, mark the both as abstract

leleuj:
Updated pull.

mmoayyed:
:+1:

Issue #279, opened 2015-09-18
Supports cookies in the WebContext


leleuj:
- create a new `Cookie` object

- create the `Cookie[] getRequestCookies()` and `void addResponseCookie(Cookie cookie)` methods to the `WebContext` interface

- create a `CookieClient`

millross:
Could you provide an example of the use case for working directly with cookies from pac4j please? I want to make sure that I don't conflict with existing vert.x cookie handling :) or solutions to problems we might use cookies to solve

leleuj:
I want the "best" design for pac4j v1.8 and implementations. So I have started trying to implement advanced use cases (I didn't commit anything).
My use case: I have a CAS server which generates a remember-me cookie which must be validated via a web service to grant the user a remember-me status.
To address that, I need to access the cookies of the request thus this issue.

mmoayyed:
:+1: The WebContext interface already has methods that deal with responses. Adding cookie support is the next step.

millross:
Ok, I think that works for me, though in the vert.x world it makes us dependent on use of a CookieHandler. Is use of your CAS server available to direct, or only indirect clients? The reason I ask is because setting up for an indirect flow in vert.x requires enabling of the CookieHandler anyway, but otherwise I'll need to document the requirement for the CookieHandler when using advanced use cases which need cookies (which in turn means that I'll need to know what those are :))

mmoayyed:
I suppose cookie authN would be relevant to `DirectHttpClient`?

leleuj:
The CAS server was just an example. It's the use case where the identity of the user is held by a cookie: it should be a direct client.

leleuj:
Update in Play: https://github.com/pac4j/play-pac4j/pull/80

PR #283, opened 2015-09-21
Initial cookie design for webcontext

mmoayyed:
Closes https://github.com/pac4j/pac4j/issues/279

Let's review the current design, and I will continue to update pulls with docs, and tests.


leleuj:
The build fails: `org.pac4j.http.credentials.CookieCredentials is not abstract and does not override abstract method hashCode() in org.pac4j.http.credentials.HttpCredentials`

pac4j-core/src/main/java/org/pac4j/core/context/WebContext.java
```diff
@@ -162,9 +163,24 @@
...
-     * 
+     *
      * @return the URL
      * @since 1.5.0
      */
     String getFullRequestURL();
+
+    /**
+     * Retrieves request cookies.
+     * @since 1.8.1
```
leleuj: If the PR is ready, let's do it for 1.8.0, no?

pac4j-http/src/main/java/org/pac4j/http/client/direct/CookieClient.java
```diff
@@ -0,0 +1,83 @@
...
+ * @since 1.8.1
+ */
+public class CookieClient extends DirectHttpClient<CookieCredentials>  {
+    private String cookieName;
+    private String cookieValue;
+
+    public CookieClient() {
+    }
+
+    public CookieClient(final CookieAuthenticator cookieAuthenticator) {
```
leleuj: I think there should be a constructor with a cookie name.

pac4j-http/src/main/java/org/pac4j/http/client/direct/CookieClient.java
```diff
@@ -0,0 +1,83 @@
...
+import org.pac4j.http.credentials.extractor.CookieExtractor;
+import org.pac4j.http.profile.creator.ProfileCreator;
+
+/**
+ * @author Misagh Moayyed
+ * @since 1.8.1
+ */
+public class CookieClient extends DirectHttpClient<CookieCredentials>  {
+    private String cookieName;
+    private String cookieValue;
```
leleuj: I don't see why we need a `cookieValue`: it should be the returned value for the `CookieExtractor`, shouldn't it?

pac4j-http/src/main/java/org/pac4j/http/credentials/CookieCredentials.java
```diff
@@ -0,0 +1,59 @@
...
+
+package org.pac4j.http.credentials;
+
+import org.pac4j.core.util.CommonHelper;
+
+/**
+ * @author Misagh Moayyed
+ * @since 1.8.1
+ */
+public class CookieCredentials extends HttpCredentials {
```
leleuj: I would have expected to use the `TokenCredentials` set with the cookie value (search based on the name)

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/CookieAuthenticator.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.common.base.Predicate;
+import com.google.common.base.Predicates;
+import org.pac4j.core.exception.CredentialsException;
+import org.pac4j.http.credentials.CookieCredentials;
+
+/**
+ * @author Misagh Moayyed
+ * @since 1.8.1
+ */
+public class CookieAuthenticator implements Authenticator<CookieCredentials> {
```
leleuj: I don't think we need that by default: the cookie authenticator will be some dedicated implementation from the `TokenAuthenticator`.

pac4j-http/src/main/java/org/pac4j/http/credentials/extractor/CookieExtractor.java
```diff
@@ -0,0 +1,51 @@
...
+    private final String clientName;
+
+    public CookieExtractor(final String cookieName, final String cookieValue, final String clientName) {
+        this.cookieName = cookieName;
+        this.clientName = clientName;
+        this.cookieValue = cookieValue;
+    }
+
+    @Override
+    public CookieCredentials extract(final WebContext context) {
```
leleuj: I would have expected this extractor to return the value of the found cookie (by its name).

pac4j-core/src/main/java/org/pac4j/core/context/WebContext.java
mmoayyed: Could be ready yes. If the schedule is by Thu, then we can make this work

leleuj:
One more thing: do you think you can update the Play web context: https://github.com/pac4j/play-pac4j/blob/master/play-pac4j-java/src/main/java/org/pac4j/play/PlayWebContext.java? Or do you want me to handle that?

mmoayyed:
I will review the comments here shortly, thanks for reviewing. As for play, go for it. I have not had much experience with play, and it would take me twice as long :)

leleuj:
@mmoayyed : do you think you'll be able to finish that today (for the RC1 tomorrow) or do we keep it for 1.8.1?

mmoayyed:
I think we should do this for 1.8.1. Sorry, caught in a very confusing troubleshooting session right now :( I wont have time until likely next monday.

leleuj:
No problem.

mmoayyed:
Might be able to resume working on this pull this afternoon. Let's check back tomorrow morning to see where I stand.

mmoayyed:
OK, I am back to work on this. Stay tuned.

mmoayyed:
Made a few updates. The problem I have now, is that we want authentication to act based on a cookie in the response, correct? This is not possible without drastic measures because the response does not expose a way to read cookies.

mmoayyed:
Think I have that wrong. The authentication should work based on request cookies!

mmoayyed:
Updated pull.

leleuj:
Yes, request cookies. Will review.

mmoayyed:
Not sure why the build fails! Everything compiles just fine for me. Maybe the build status is old?

leleuj:
Regarding the build, you're not up-to-date with the master. I cleaned classes you are still referencing...

```
[ERROR] /home/travis/build/pac4j/pac4j/pac4j-http/src/test/java/org/pac4j/http/client/direct/CookieClientTests.java:[46,15] cannot find symbol
[ERROR] symbol:   class SimpleTestUsernameProfileCreator
```

pac4j-http/src/test/java/org/pac4j/http/client/direct/CookieClientTests.java
```diff
@@ -0,0 +1,96 @@
...
+ * @author Misagh Moayyed
+ * @since 1.8.0
+ */
+public class CookieClientTests implements TestsConstants {
+
+    @Test
+    public void testClone() {
+        final CookieClient oldClient = new CookieClient();
+        oldClient.setName(TYPE);
+        final SimpleTestUsernameProfileCreator profileCreator = new SimpleTestUsernameProfileCreator();
```
leleuj: This profile creator does not exist anymore as the authenticator saves its profile in the credentials for the default `AuthenticatorProfileCreator`.

leleuj:
+1 Just the test (and the build) must be fixed

mmoayyed:
Merged.

leleuj:
Build fixed. Will merge and update play-pac4j. Thanks.

PR #284, opened 2015-09-21
Clean and deprecate useless concepts

leleuj:


mmoayyed:
:+1: though looks like the build broke?

leleuj:
I forgot to fix the tests. New commit for that. The build passes.

Issue #289, opened 2015-09-25
SAML2 client not working if IdP is in different time zone


DanielVocke:
Turns out the function that detemines whether a SAML2 authentication request is still valid does not take time zones into consideration.
```SAML2DefaultResponseValidator.isDateValid()``` returns false if the IdP is in a different time zone and leads to an invalid authentication request.

DanielVocke:
as workaround until this is fixed, set the setMaximumAuthenticationLifetime to 24h:

```((SAML2DefaultResponseValidator) saml2Client.getResponseValidator()).setMaximumAuthenticationLifetime(24 * 3600 * 1000);```

it would probably be nice to expose that setter method on the interface as well in future...

mmoayyed:
Thanks for the report. Looks like the time received is handled via joda time which should consider time zones, but the sp time is handled by system milliseconds which does not. I think both time instances should be based on jodatime to mitigate the problem.

leleuj:
Thanks for reporting and the workaround. Would you mind proposing some PR to mitigate the problem?

mmoayyed:
I'll put together a PR for this issue today.

leleuj:
Thanks.

PR #295, opened 2015-09-30
Use the default system timezone when validating issue instant

mmoayyed:
Closes https://github.com/pac4j/pac4j/issues/289

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/sso/SAML2ResponseValidator.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/sso/SAML2ResponseValidator.java
@@ -27,4 +27,8 @@ import org.pac4j.saml.context.SAML2MessageContext;
 public interface SAML2ResponseValidator {
 
     Credentials validate(SAML2MessageContext context);
+
+    void setMaximumAuthenticationLifetime(int maximumAuthenticationLifetime);
+
+    void setAcceptedSkew(int acceptedSkew);
 }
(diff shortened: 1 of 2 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/sso/SAML2ResponseValidator.java
```diff
@@ -27,4 +27,8 @@
 public interface SAML2ResponseValidator {
 
     Credentials validate(SAML2MessageContext context);
+
+    void setMaximumAuthenticationLifetime(int maximumAuthenticationLifetime);
```
mmoayyed: Per the issue request, made these available.

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
```diff
@@ -681,9 +681,9 @@ protected final void validateSignature(final Signature signature, final String i
     }
 
     private boolean isDateValid(final DateTime issueInstant, final int interval) {
-        final long now = System.currentTimeMillis();
-        return issueInstant.isBefore(now + this.acceptedSkew * 1000)
-                && issueInstant.isAfter(now - (this.acceptedSkew + interval) * 1000);
+        final DateTime before =  DateTime.now().plusSeconds(acceptedSkew);
```
mmoayyed: This basically looks at the system property to find the timezone, then looks at the TimeZone class of Java to get the default timezone of the system.

leleuj:
+1

@DanielVocke : would you mind testing this change?

Issue #299, opened 2015-10-02
pac4j-saml: Failure when verifying Recipient URL with default port in SAML response


jkacer:
Pac4J 1.7.1 throws a validation error if a SAML response arrives with a value of SubjectConfirmationData/Recipient that does NOT contain a port number (because the default port is assumed). PAC4J compares the URL value against against a URL that always contains a port number.

I think this behavior is different from version 1.6.0.

Code responsible for the verification is in class  SAML2DefaultResponseValidator. In 1.6.0, it was class Saml2ResponseValidator.

See full discussion and details here: https://groups.google.com/forum/#!topic/pac4j-users/pfQQBu23_ME

mmoayyed:
Can you update the metadata to remove the port?

jkacer:
Misagh, the issue happens when the metadata does NOT contain the port. If I add the port (https://server:443/something) it works OK.

I could add the port everywhere but unfortunately we have our metadata already distributed for almost a year, so I would have to re-negotiate and re-test the system with all partner companies. That's almost impossible.

leleuj:
So the real solution is to update the metadata which is not always practically possible. We could help and relax on the check of the port, some kind of improvement. Maybe it should be something we could enable or disable with a flag.

@jkacer : would you mind proposing a PR in that direction? We'll put in the 1.1.8 release (this Thursday) if possible.

mmoayyed:
Yes exact-match comparisons would be a problem. The code should switch to URL comparisons.

jkacer:
HI, here is the pull request:
https://github.com/pac4j/pac4j/pull/303

It ignores default ports for HTTP and HTTPS.


BTW, please have a look at SAML2DefaultResponseValidator:513
        // TODO Validate inResponseTo
I am pretty sure this TODO does not exist in 1.7.1/1.7.x anymore and there is a regular check of InResponseTo instead. This would mean that some code from 1.7.x did not get into 1.8.x.
We failed on this check several times last week with a partner company, that's why I remember it. They did not return InResponseTo; this passed fine with 1.6.0 but not with 1.7.1.

PR #303, opened 2015-10-05
Issue #299 - Failure when verifying Recipient URL with default port in SAML response

jkacer:
Changed the way how the SubjectConfirmationData/Recipient is verified -
default port numbers (80 for HTTP and 443 for HTTPS) are now ignored.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
@@ -68,10 +68,14 @@ import org.pac4j.saml.crypto.SAML2SignatureTrustEngineProvider;
 import org.pac4j.saml.exceptions.SAMLException;
 import org.pac4j.saml.sso.SAML2ResponseValidator;
 import org.pac4j.saml.storage.SAMLMessageStorage;
+import org.pac4j.saml.util.UriUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.xml.namespace.QName;
+
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
```diff
@@ -523,23 +527,29 @@ protected final boolean isValidBearerSubjectConfirmationData(final SubjectConfir
...
+            } else {
+                final Endpoint endpoint = context.getSAMLEndpointContext().getEndpoint();
+                if (endpoint == null) {
+                    logger.warn("No endpoint was found in the SAML endpoint context");
+                    return false;
+                }
+    
+                final URI recipientUri = new URI(data.getRecipient());
+                final URI appEndpointUri = new URI(endpoint.getLocation());
+                if (!UriUtils.urisEqualAfterPortNormalization(recipientUri, appEndpointUri)) {
```
mmoayyed: Isn't there some Apache library that does this sort of thing?

jkacer: I looked into org.apache.http.client.utils.URIUtils and it doesn't do that.

It could be done with simple URL.equals() but there are reasons why not to use URL for comparison, please see http://stackoverflow.com/questions/3771081/proper-way-to-check-for-url-equality

Maybe there is a library that does exactly what I wrote but I wasn't able to find it.

pac4j-saml/src/main/java/org/pac4j/saml/util/UriUtils.java
```diff
@@ -0,0 +1,105 @@
+package org.pac4j.saml.util;
```
leleuj: License header please.

pac4j-saml/src/main/java/org/pac4j/saml/util/UriUtils.java
```diff
@@ -0,0 +1,105 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * URI utilities.
+ * 
+ * @author jkacer
+ */
```
leleuj: I like to see `@since` annotation to track versions.

leleuj:
Almost perfect! Thanks

@mmoayyed : ok for you?

mmoayyed:
Yes. I dont know if I like the UriUtils class. Would have preferred if something else could handle that but I dont have any better answers.

leleuj:
Yep, I'm not a big fan of having a pac4j `UriUtils`, I wish it would be available in some kind of library, but I haven't been able to find it somewhere.

I'm comforted by the quality of code and the associated tests.

Will merge.

jkacer:
Hi, shall I add the license header and the missing since annotation, once already merged?

leleuj:
No, I was too quick, I will fix it directly. Thanks for reminding.

jkacer:
OK, thank you.

PR #318, opened 2015-10-15
Manage path exclusions

leleuj:


pac4j-core/src/main/java/org/pac4j/core/context/ExcludePathMatcher.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public String getExcludePath() {
+        return excludePath;
+    }
+
+    public void setExcludePath(String excludePath) {
+        this.excludePath = excludePath;
+        if (CommonHelper.isNotBlank(excludePath)) {
+            logger.warn("Excluding paths is an advanced feature: be careful when defining your regular expression to avoid any security issue!");
+            if (!excludePath.startsWith("^") || !excludePath.endsWith("$")) {
```
mmoayyed: Why this rule?

pac4j-core/src/main/java/org/pac4j/core/context/J2ERequestContext.java
```diff
@@ -160,4 +160,11 @@ public String getFullRequestURL() {
         }
         return pac4jCookies;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public String getPath() {
```
mmoayyed: Do an @Override here, and I think that makes the inheritdoc unnecessary.

pac4j-core/src/main/java/org/pac4j/core/context/ExcludePathMatcher.java
leleuj: That's a good question. At first, I wanted to allow people to define the regexp they want to match a path, but I realized it could be very dangerous.

Let's say I have two pages: `unsecure.html` and `secure.html`. I put a filter on both and want to exclude one with `unsecure\.html` as the `excludePath`. It works perfectly and it feels good, but it's a **security breach**.

Because, as an attacker, I can call `secure.html?attack=unsecure.html` and it matches the exclusion path and no security applies!

So I decide to force a full path starting with `^`  and ending with `$`.  If you use `^unsecure\.html$` as the `excludePath`, it works without any security issue.

pac4j-core/src/main/java/org/pac4j/core/context/J2ERequestContext.java
leleuj: Yes, I will add the missing `@Override`.

leleuj: Reading the Javadoc documentation, it even seems to be unnecessary:

--

**Automatic Copying of Method Comments**

The Javadoc tool has the ability to copy or "inherit" method comments in classes and interfaces under the following two circumstances.

Inherit from classes and interfaces - Inheriting of comments occurs in all three possible cases of inheritance from classes and interfaces:

- When a method in a class overrides a method in a superclass
- When a method in an interface overrides a method in a superinterface
- When a method in a class implements a method in an interface

--

So I did the test and the comment is actually inherited without using the `{@inheritDoc}` or the `@Override` annotation.

leleuj:
Updated pull with `Override`.

mmoayyed:
:+1:

PR #326, opened 2015-10-22
Improve Config building

leleuj:


leleuj:
Small PR. Assuming everything is ok. Will merge.

PR #327, opened 2015-10-22
Improve IP authentication support

leleuj:


pac4j-core/src/main/java/org/pac4j/core/client/AnonymousClient.java
```diff
@@ -0,0 +1,68 @@
...
+import org.pac4j.core.exception.TechnicalException;
+import org.pac4j.core.profile.AnonymousProfile;
+
+/**
+ * Anonymous client. Not to be used except for advanced use cases.
+ *
+ * @author Jerome Leleu
+ * @since 1.8.1
+ */
+public class AnonymousClient extends BaseClient<AnonymousCredentials, AnonymousProfile> {
```
leleuj: To secure the /status and /statistics endpoints of the CAS server, I use the `IpClient` with an `IpRegexpAuthenticator`: from a pac4j point of view, the user is not authenticated if he has not the right IP. Thus, a 401 HTTP error. Spring Security returns a 403 HTTP error.
To fully mimic the Spring Security behaviour, the solution at the pac4j level is to use an `AnonymousClient` (the user is authenticated even if anonymous) and an `IpRegexpAuthorizer` to generate a 403 HTTP error in case of improper IP.

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/IpRegexpAuthenticator.java
```diff
@@ -46,9 +53,19 @@ public void validate(final TokenCredentials credentials) {
         if (!this.pattern.matcher(ip).matches()) {
             throw new CredentialsException("Unauthorized IP address: " + ip);
         }
+
+        final IpProfile profile = new IpProfile(ip);
```
leleuj: The profile should be saved in the credentials as the default profile creator gets it there.

pac4j-core/src/main/java/org/pac4j/core/client/AnonymousClient.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    protected AnonymousProfile retrieveUserProfile(final AnonymousCredentials credentials, final WebContext context) {
+        return PROFILE;
+    }
+
+    @Override
+    public ClientType getClientType() {
+        return null;
```
mmoayyed: Would this cause problems later on? Shall we define "UNKNOWN" for example?

mmoayyed:
:+1:

pac4j-core/src/main/java/org/pac4j/core/client/AnonymousClient.java
leleuj: You're right: I will change that to prevent any future issue.

PR #335, opened 2015-11-02
Clean logging dependencies

leleuj:
Mark logback as a test dependency for pac4j-saml
Exclude commons-logging and add jcl-over-slf4j


mmoayyed:
:+1:

PR #336, opened 2015-11-02
Add security header authorizers

leleuj:


pac4j-http/src/main/java/org/pac4j/http/authorization/authorizer/IpRegexpAuthorizer.java
```diff
@@ -13,7 +13,7 @@
    See the License for the specific language governing permissions and
    limitations under the License.
  */
-package org.pac4j.http.authorization;
+package org.pac4j.http.authorization.authorizer;
```
leleuj: Push this new authorizer in the `authorizer` package. See: https://github.com/pac4j/pac4j/issues/337

pac4j-oauth/src/test/java/org/pac4j/oauth/client/TestBenchClients.java
```diff
@@ -36,9 +36,9 @@
...
     
-    private static final int MAX_FACEBOOK = 10000;
+    private static final int MAX_FACEBOOK = 1000;
```
leleuj: Downsize the number of elements used for the bench to decrease the build time. Maybe this tests should bed moved out of the unit tests...

mmoayyed:
:+1:

PR #338, opened 2015-11-04
Add CheckHttpMethod Authorizer

leleuj:


pac4j-core/src/main/java/org/pac4j/core/authorization/authorizer/CheckHttpMethodAuthorizer.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public CheckHttpMethodAuthorizer(final List<HTTP_METHOD> methods) {
+        setMethods(methods);
+    }
+
+    public CheckHttpMethodAuthorizer(final Set<HTTP_METHOD> methods) {
+        this.methods = methods;
+    }
+
+    /**
```
mmoayyed: Use @Override instead. Make this block unnecessary.

mmoayyed:
:+1:

pac4j-core/src/main/java/org/pac4j/core/authorization/authorizer/CheckHttpMethodAuthorizer.java
leleuj: Good catch. Updated and other authorizers as well.

PR #347, opened 2015-11-07
InitializableWebObject

leleuj:


leleuj:
Will merge unless there is a concern

PR #349, opened 2015-11-09
Replace the contextualCallbackUrl by a specific resolver

leleuj:


leleuj:
Will merge tomorrow unless there is a concern

PR #350, opened 2015-11-09
Matcher to replace excludePath

leleuj:


leleuj:
Will merge tomorrow unless there is a concern

PR #351, opened 2015-11-10
New feature: Database configuration of SAML clients

jkacer:
Added a new Maven module with SAML clients able to be initialized from a relational database.
There is quite a lot of duplicate code between PAC4J-SAML and PAC4J-SAML-DB, further changes are assumed that should remove most duplicates.


jkacer:
Hello, here is the pull request that I promised during our discussion at https://groups.google.com/forum/#!topic/pac4j-users/aEVSx5FboKI

I don't expect that you will merge it right away; as I said before, there is a certain amount of duplicate code. Removing it would mean slightly refactoring the existing code and I wasn't sure I was allowed to do that. But this can done as a next step.

Originally I had this prepared for 1.7.1 but I have rewritten it against the current master.

leleuj:
Thank you very much. I don't think we can merge this right now as there is indeed too much duplicated code, but it's a good starting point. I will review.
I think we can target an integration for version 1.8.2 or 1.9.0 (next version planned in a few days is 1.8.1)

jkacer:
Sure, thank you for having a look.

In the meantime, I realized the CI build failed due to incorrect JavaDoc. I'll fix that. It was aimed at 1.7 and I forgot to update it for 1.8.

jkacer:
Fixed :-)

leleuj:
I would have prefered `pac4j-saml-config-db` for the name of the module.

pac4j-saml-db/pom.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
leleuj: Think of adding the license headers...

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClient.java
```diff
@@ -0,0 +1,369 @@
...
+import org.pac4j.saml.transport.Pac4jSAMLResponse;
+import org.pac4j.saml.util.Configuration;
+import org.w3c.dom.Element;
+
+import net.shibboleth.utilities.java.support.component.ComponentInitializationException;
+import net.shibboleth.utilities.java.support.resolver.ResolverException;
+
+
+/**
+ * An alternative to {@link SAML2Client}. It accepts {@link DbLoadedSamlClientConfiguration} instead of {@link SAML2ClientConfiguration}.
```
leleuj: The configuration is loaded once, being static or in your support, from the database.

So I guess we should only have one client: `SAML2Client` and two configurations: one being static and one from database.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClientConfiguration.java
```diff
@@ -0,0 +1,211 @@
...
+ * <li>IdP metadata - contained as a string, not a path to a resource</li>
+ * <li>SP metadata - dropped completely</li>
+ * <li>Client name - added</li>
+ * </ul>
+ *
+ * TODO: It would be nice to have a common interface and maybe a common abstract parent too.
+ * 
+ * @author jkacer
+ */
+public class DbLoadedSamlClientConfiguration implements Cloneable {
```
leleuj: You should have a `SAML2DatabaseConfiguration` inheriting from `SAML2Configuration` and this latter should become an `InitializableWebObject`.
In the regular `SAML2Configuration`, the `init(context)` should do nothing while in the `SAML2DatabaseConfiguration`, the `init(context)` method should load data from database.
This `init(context)` method should be of course called itself from the `init(context)` method of the `SAML2Client`.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClients.java
```diff
@@ -0,0 +1,198 @@
...
+ * called to find the right client according to the input context or type. The {@link #findAllClients(WebContext)} method returns all the
+ * clients.
+ * </p>
+ *
+ * TODO: We could have a common class (maybe abstract) that would contain common code of this class and {@link Clients}. Most code is
+ * duplicated. {@link Clients} cannot be extended because it is final.
+ * 
+ * @author jkacer
+ */
+public class DbLoadedSamlClients extends InitializableWebObject {
```
leleuj: We can keep this one: loading data from database should saty "hidden" in the `SAML2Client`.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/dao/api/SamlClientDao.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.List;
+
+import org.pac4j.saml.dbclient.DbLoadedSamlClientConfiguration;
+
+
+/**
+ * DAO to manipulate SAML Client Configurations stored in a data source.
+ * 
+ * @author jkacer
+ */
```
leleuj: Add a `@since` targeting the version 1.9.0.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/dao/api/SamlClientDao.java
```diff
@@ -0,0 +1,22 @@
+package org.pac4j.saml.dbclient.dao.api;
```
leleuj: Add license.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/dao/api/SamlClientDao.java
```diff
@@ -0,0 +1,22 @@
...
+ * @author jkacer
+ */
+public interface SamlClientDao {
+
+	/**
+	 * Loads all existing SAML Client Configurations.
+	 * 
+	 * @return A list of configurations.
+	 */
+	public List<DbLoadedSamlClientConfiguration> loadAllClients();
```
leleuj: While it's better to load all clients at the same time, I'd prefer to load each client's configuration separately to embed this logic inside the SAML database configuration component.
You could even keep this "load all" logic, cache the result and just provide configuration one by one for each `SAML2Client`.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbcrypto/KeyStoreCredentialProvider2.java
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Alternative to {@link KeyStoreCredentialProvider}. Unlike {@link KeyStoreCredentialProvider}, it does not take a path to the keystore but the keystore binary content.
+ *
+ * TODO: We cannot extend {@link KeyStoreCredentialProvider} because it needs a path in the constructor. It would be good to have
+ * a common ancestor to reuse some code. Most methods are just copied - duplicate code.
+ * 
+ * @author jkacer
+ */
+public class KeyStoreCredentialProvider2 implements CredentialProvider {
```
leleuj: We need a better name: `KeyStoreDatabaseCredentialProvider while the old one could be renamed as `KeyStoreFileCredentialProvider` and both should certainly share some code with a common ancestor.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbmetadata/SAML2IdentityProviderMetadataResolver2.java
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Alternative to {@link SAML2IdentityProviderMetadataResolver} that doesn't take a path to a metadata file but directly metadata content.
+ *
+ * TODO: We cannot extend {@link SAML2IdentityProviderMetadataResolver} because it needs a path in the constructor. It would be good to have
+ * a common ancestor to reuse some code.
+ * 
+ * @author jkacer
+ */
+public class SAML2IdentityProviderMetadataResolver2 implements SAML2MetadataResolver {
```
leleuj: Same comments as for the `KeyStoreCredentialProvider`.

pac4j-saml-db/src/test/java/org/pac4j/saml/dbclient/DbLoadedSamlClientIT.java
```diff
@@ -0,0 +1,189 @@
...
+import com.gargoylesoftware.htmlunit.html.HtmlSubmitInput;
+import com.gargoylesoftware.htmlunit.html.HtmlTextInput;
+
+
+/**
+ * Base for integration tests of {@link DbLoadedSamlClient}. Most code is taken from {@link SAML2ClientIT} and little adapted.
+ * 
+ * @author jkacer
+ */
+public abstract class DbLoadedSamlClientIT extends ClientIT implements TestsConstants {
```
leleuj: Great to have IT and unit tests.

leleuj:
Many thanks for this big work.

I've made many comments and indeed, many things need to be changed, though I think the biggest and thoughest part of the logic is here: we are close to have something that fits well in pac4j.

Just let us know when you'll update the pull request.

jkacer:
Hi Jerome, thanks for the review and feedback.
I'll go through it next week, I think. I cannot do this at work anymore (because we already have something working), so I'll do it at home -> slow progress.
I'll put my answers to your comments individually when I process them.

leleuj:
No problem. Do as you can, but open source should always be encouraged even at work -> virtuous circle (the more people contribute, the more people use contributions and save time for contributing themselves)

pac4j-saml-db/pom.xml
jkacer: Done

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/dao/api/SamlClientDao.java
jkacer: Done.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/dao/api/SamlClientDao.java
jkacer: Done. I have added a method to load a single client's configuration.

As for caching, I prefer not to have it in the base DAO implementation, so I created a very simple caching implementation of the DAO, which uses an underlying "real" DAO and only adds some trivial caching. Please see class CachingSamlClientDaoImpl.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/dao/api/SamlClientDao.java
jkacer: Done.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClient.java
jkacer: Yes, this was my initial idea but I had to introduce the extra client because of the constructor parameter typed to the existing configuration. I think we can get rid of the extra client after making those changes of the Configuration object you propose.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClientConfiguration.java
jkacer: I'm working on it right now and I've just encountered a little issue. When loading the configuration in init(context), we must already know the name for which we are loading the data - the names are not known in advance, unlike in case of static configuration. This means all names must be somehow read in advance and a particular name must be given to the configuration object. We can read all names using the DAO, no problem here. Is this how you meant it?

leleuj: I think the init method in the configuration object (called by the
init(context) of the client) should certainly take the client name as
input...

2015-11-16 14:28 GMT+01:00 Jaroslav KAČER <notifications@github.com>:

jkacer: Ah, thanks! I completely forgot I could use the web context for that.

leleuj:
@jkacer : a need for common configuration capabilities has arisen when working on pac4j integration in CAS and Knox (security gateway for Haddop) and it has lead me to: https://github.com/pac4j/pac4j/issues/357

This means that there will be a new `pac4j-config` module which gathers all the stuffs related to pac4j configuration outside the default basic configuration so your database configuration should definitely go into this new module. I let you update the PR when mine will be merged.

Thanks

jkacer:
@leleuj Thanks for the info. I will make my changes first and then move it into the new module, when it is in master.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClientConfiguration.java
jkacer: Done. I have introduced an abstract parent for both SAML configurations, both the classic one and the database-loaded one extend the abstract parent. This allowed me to get rid of DbLoadedSamlClient completely - hurray.
The configuration is initialized in iternalInit(), as you suggested. But there is one little detail about it to be noted: In the Clients class (in DbLoadedSamlClients now), we must fetch at least all existing client names - we must programatically create the clients and their (empty) configurations and add them to the list. So touching the DB is not completely deferred until configuration initialization. Without this, Clients.findClient(name) would not work because it would not simply know about a specific client at all.
Can you please have a look?

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClient.java
jkacer: Done too, see the comments below...

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClientConfiguration.java
leleuj: In the findClient(context, name), the first call in the method is to fire
the initialization: init(context) which calls internalInit(context) so
there should be no issue, right?

2015-11-18 13:12 GMT+01:00 Jaroslav KAČER <notifications@github.com>:

jkacer: Yes, exactly, that's why left the loading of client names from DB in internalInit(). Like this, findClient() will match the required name with an already-known client name and everything will be OK.

I just wanted to point out that the DB initialization code is not 100% placed in the configuration class itself but a small part of it is also in the clients class.

I'm just working on it deeper, as per another comment of yours...

pac4j-saml-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClients.java
jkacer: Done, although not exactly as you wished. I removed all duplicate code but kept the new class. It extends the existing Clients class, so it has all its capabilities. It only overrides the internalInit() method + a few setters. So I believe the new class can be used anywhere where the old one is used now, bringing the option to load SAML clients in addition to defining clients statically.

If you really do not like this approach, then I would suggest to add something like "Dynamic Client Contributors" to the Clients class. It would have a list of contributors, where each would be allowed to add some clients to the list of managed clients. The addition would happen in internalInit(), similarly to the current implementation in DbLoadedSamlClients.

pac4j-saml-db/src/main/java/org/pac4j/saml/dbcrypto/KeyStoreCredentialProvider2.java
jkacer: Done. So we have a base class (KeyStoreCredentialProvider) and two specialized sub-classes (KeyStoreFileCredentialProvider and KeyStoreByteArrayCredentialProvider). The two sub-classes only redefine the constructor, otherwise all code is inherited from the base class.

leleuj:
@jkacer : I finally changed my mind, I think both components have very different goals and different dependencies, so I think it should to keep different modules. Unless you think otherwise...

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClients.java
```diff
@@ -0,0 +1,197 @@
...
+
+
+	/**
+	 * Initializes all clients by loading them from the database.
+	 * 
+	 * @see org.pac4j.core.util.InitializableWebObject#internalInit(org.pac4j.core.context.WebContext)
+	 */
+	@Override
+	protected void internalInit(final WebContext context) {
+        // Remove all clients first if re-initialized
```
leleuj: I definitely think we should get rid of this class. The initialization of the clients from database should not happen in this class.

Let me be more specific:

- The `internalInit` method of `Clients` is called every time you try to find a client: it only computes the callback url of the client from the callback url of the `Clients`
- Every time you call a method of the client: `redirect`, `getCredentials` or `getUserProfile`,  the `internalInit` method of the client is called: this is where the initialization of the configuration from DB should happen.

The flow I see:
In the `SAML2Client`, the `internalInit(context)` would call `init(clientName, context)` from the `AbstractSAML2Configuration`. So maybe I misleaded you, but the `AbstractSAML2Configuration` should not be a `InitializableWebObject`, it should have a `init(String clientName, WebContext context)` method.
The default `SAML2Configuration` should have an empty `init` method while the `DatabaseSAML2Configuration` should use the DTO to load the configuration associated with the client defined by the `clientName` parameter.

Does it make sense?

pac4j-saml-db/src/main/java/org/pac4j/saml/dbmetadata/SAML2IdentityProviderMetadataResolver2.java
jkacer: Done. I added an abstract class (AbstractSAML2IdentityProviderMetadataResolver) and the two implementations now extend it, overriding/implementing only a few different methods. Impls: PathBasedSAML2IdentityProviderMetadataResolver and StringBasedSAML2IdentityProviderMetadataResolver.

Sorry for slow progress, I have had very little time to work on PAC4J recently.

I'll continue with those changes in Clients (discussion below) and with merging changes from PAC4J/master, I see some code conflicts now.

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClients.java
jkacer: @leleuj Thank you for the detailed explanation; I'll try to follow it and rework the code.

jkacer: @leleuj Sorry, but I have another question / comment.
The part about the configuration is clear, I have already done that.

However, a problem arises when I move to Clients and SAML2Client. What you wrote above would, in my opinion, work well if at the beginning we had the Clients object initialized with a list of SAML2Client-s and their not-yet-loaded DatabaseSAML2Configurations. Then, it would work as you suggest:
- Clients.findClient(name) - finds it OK by a known static name
- Clients.internalInit() - nothing important here
- SAML2Client.internalInit() - We try to initialize the configuration here, using a known name, and call DatabaseSAML2Configuration.init(clientName, context)
- Loaded OK --> everything OK

But it does not work if you have an unknown set of client definitions in the database and you don't know in advance:
- how many of them you have,
- what their names are.
You cannot create a new SAML2Client and its configuration in existing Clients because this class knows nothing about SAML clients etc. This means you will never find the client because Clients.findClient() iterates over a known list of clients. But you don't have the client in the list - Clients do not have this information because they don't know anything about SAML / SAML in DB.

That is why I first created the database-specific version of Clients. I agree with you that this is not nice and it should go away. That's why I proposed the dynamic "additional client contributor", which would tell Clients that there are additional clients that should be taken into account (and loaded later as you suggest).

Or was your intention to have the DB clients defined statically (as I described above, e.g. in Spring or simply Java) and just have all properties loaded dynamically?

Or maybe I haven't fully understood it yet?

jkacer: Hi Jerome, I have finally kicked out the DbLoadedSamlClients class in commit 9014980. There should be no duplicate code anymore. What exactly I did:
- Modified the configuration as you suggested: init(name, context), loading properties from DB here, no longer an InitializableWebObject.
- DbLoadedSamlClients is gone --> no duplicate code anymore.
- In order to let Clients know about some clients configured in the database, I introduced interface AdditionalClientContributor and an implementation that scans available SAML clients defined in the database (just names, no details) and contributes them to Clients.
- The rest then works as you described: Clients finds a client (either static or dynamic) and the chosen client then initializes its configuration inside internalInit().

I could not find an easier way how to accomplish what I want to do.
Please review and share your comments. Thank you!

leleuj: @jkacer :

First of all, thanks for your patience and tenacity on this.

Ahhh! I think I get it. I had in mind that the configuration could be
loaded from database for a SAML2Client: this is our first use case. For
that, we can keep the Clients object and have two configuration objects.

But what you also propose is another use case where the clients are not
known by default, they would be loaded from the database for example: am
getting it right?

I don't think it really makes sense to mix clients which are pre-defined
and others which are loaded from database.

I would keep the Clients class and change reference to this.clients by
getClients(). Then, I would create a new DatabaseSAML2Clients inherited
from Clients and overriding the getClients() method by a call to the
database to get all SAML2 clients. I would not create a new
concept: AdditionalClientContributor right now, but I may be mistaken.

What do you think?


2015-11-25 18:04 GMT+01:00 Jaroslav KAČER <notifications@github.com>:

leleuj:
@jkacer :

First of all, thanks for your patience and tenacity on this.

Ahhh! I think I get it. I had in mind that the configuration could be loaded from database for a SAML2Client: this is our first use case. For that, we can keep the Clients object and have two configuration objects.

But what you also propose is another use case where the clients are not known by default, they would be loaded from the database for example: am getting it right?

I don't think it really makes sense to mix clients which are pre-defined and others which are loaded from database.

I would keep the Clients class and change reference to this.clients by getClients(). Then, I would create a new DatabaseSAML2Clients inherited from Clients and overriding the getClients() method by a call to the database to get all SAML2 clients. I would not create a new concept: AdditionalClientContributor right now, but I may be mistaken.

What do you think?

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClients.java
jkacer: @leleuj Hi! Yes, now I see the two different perspectives: You assumed that (a) client(s) will be defined statically with a reference to a configuration that can load itself dynamically, while I assumed that you will know nothing at the beginning at the clients will be discovered at run-time. Now we are on the same frequency :-)

What you propose makes sense to me. So I will:
- Revert Clients back and enhance little to allow subclassing etc.
- Remove the Contributor interface + implementation.
- Re-introduce the new SAML Clients class that handles the discovery of SAML clients from the database.

And then we should have both cases covered, I believe.

I will most likely deliver it next Monday/Tuesday, depending on how much free time I have.

mmoayyed:
We'll need to pull this into a branch so that we can locally work on it. As such, it's slightly difficult to connect all the pieces.

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DbLoadedSamlClients.java
jkacer: Hi! It should be done now. The AdditionalClientContributor is gone, there is a new specialized clients class that subclasses Clients, overrides getClients(); there is one additional internal method.

jkacer:
@mmoayyed Hi, currently this PR goes against master but if you wish, I can create another one against a branch in PAC4J.

leleuj:
@mmoayyed : a branch would make sense if we want to work on it and not only @jkacer: is it what you have in mind? My feeling, as I reviewed several times this PR, is that we have reached a good state where we can make "final" comments and ask for the latest improvments...

mmoayyed:
Yes that is what I was thinking. I havent reviewed the code completely, but it seems like it's in a good enough shape now that we can take over and organize it internally. We might then be able to merge the feature in faster, but if @jkacer is happy with the current process, we can continue :)

jkacer:
@mmoayyed 
@leleuj 
If you still have some remarks, please tell me and I can incorporate them into the PR, otherwise feel free to merge it in. Or do you need a PR into another branch? If you do, then please fork master and tell me the branch name. I would slightly prefer if you merged in now because I cannot guarantee I will always have enough time to work on - I may have some delays, like last week.

mmoayyed:
I'll try to review the PR next week, but if you are ok with the current process we can continue on the same PR here. One thing I would ask, and you may have done this already, is that we need integration tests to pull metadata from an actual db like HSQL. This will fully confirm the changes in the long run. I'll have other comments likely when I start to review the pull. 

Could I also ask you to provide a short snippet of how the API is supposed to be used by a developer?

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -40,18 +39,19 @@
...
-public final class Clients extends InitializableObject {
-
-    private static final Logger logger = LoggerFactory.getLogger(Clients.class);
+public class Clients extends InitializableObject {
 
     public final static String DEFAULT_CLIENT_NAME_PARAMETER = "client_name";
 
     private String clientNameParameter = DEFAULT_CLIENT_NAME_PARAMETER;
 
+    /** Clients known in advance - set by the constructor or setter. */
```
leleuj: This comment should be cleaned.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -40,18 +39,19 @@
...
     private List<Client> clients;
-
+    
     private String callbackUrl = null;
 
+    // ------------------------------------------------------------------------------------------------------------------------------------
```
leleuj: This as well.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -82,34 +82,47 @@ public Clients(final Client client) {
...
     @Override
     protected void internalInit() {
-        CommonHelper.assertNotNull("clients", this.clients);
         final HashSet<String> names = new HashSet<>();
-        for (final Client client : this.clients) {
-            final String name = client.getName();
-            if (names.contains(name)) {
-                throw new TechnicalException("Duplicate name in clients: " + name);
+
+        final List<Client> loadedClients = loadClientsInternal();
```
leleuj: I don't think the `loadClientsInternal ` method should exist. Use `getClients`.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -170,8 +181,7 @@ public Client findClient(final String name) {
      * @return all the clients
      */
     public List<Client> findAllClients() {
-        init();
-        return this.clients;
+        return getClients();
```
leleuj: Not exactly. `find`means that the initialization should happen before.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -207,9 +217,19 @@ public void setClients(final List<Client> clients) {
...
-
+    
+    /**
+     * Returns initialized clients.
+     * 
+     * @return A list of clients ready to be used.
+     */
     public List<Client> getClients() {
-        return clients;
+    	init();
```
leleuj: While `get` has no initialization.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -217,4 +237,20 @@ public String toString() {
...
+	/**
+	 * An internal method that loads client definitions, if needed. The loaded clients should be returned as the result of this method and
+	 * also set into an internal variable that is later returned by {@link #getClients()}. These two methods should always be consistent
+	 * with each other.
+	 * 
+	 * @return A list of clients ready to be used.
+	 * 
+	 * @see #getClients()
+	 */
+    protected List<Client> loadClientsInternal() {
```
leleuj: To be removed.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -207,9 +217,19 @@ public void setClients(final List<Client> clients) {
...
+    
+    /**
+     * Returns initialized clients.
+     * 
+     * @return A list of clients ready to be used.
+     */
     public List<Client> getClients() {
-        return clients;
+    	init();
+    	if (this.clients == null) {
```
leleuj: What's the benefit of this test? Versus the original `CommonHelper.assertNotNull("clients", this.clients);`...

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DatabaseSAML2ClientConfiguration.java
```diff
@@ -0,0 +1,225 @@
...
+    /**
+	 * The constructor. Use setters to initialize properties.
+	 * 
+	 * @param dao
+	 *            DAO loading client configuration from DB.
+	 */
+    public DatabaseSAML2ClientConfiguration(final SamlClientDao dao) {
+    	super();
+    	if (dao == null) {
+    		throw new IllegalArgumentException("DAO must not be null.");
```
leleuj: Lets' throw `TechnicalException` to keep the pac4j spirit...

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DatabaseSAML2ClientConfiguration.java
```diff
@@ -0,0 +1,225 @@
...
+	public String getIdentityProviderMetadata() {
+		return identityProviderMetadata;
+	}
+
+	public void setIdentityProviderMetadata(String identityProviderMetadata) {
+		this.identityProviderMetadata = identityProviderMetadata;
+	}
+
+	
+	/* (non-Javadoc)
```
leleuj: What's the use of this? Why not `@Override` alone?

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DatabaseSAML2Clients.java
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	
+	
+	/* (non-Javadoc)
+	 * @see org.pac4j.core.client.Clients#getClients()
+	 */
+	@Override
+	public List<Client> getClients() {
+		init();
```
leleuj: No `init()` in `getClients()`

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/dao/impl/SpringJdbcTemplateSamlClientDaoImpl.java
```diff
@@ -0,0 +1,302 @@
...
+	 *            box. No need for OracleLobHandler.
+	 * @param tableName
+	 *            Name of the table containing configuration records.
+	 * @param environment
+	 *            Environment name; to distinguish records for different environments in a single table.
+	 */
+	public SpringJdbcTemplateSamlClientDaoImpl(final DataSource dataSource, final LobHandler lobHandler, final String tableName, final String environment) {
+		super();
+		
+		if (dataSource == null) {
```
leleuj: Use `Common.assertNotNull`

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
```diff
@@ -212,9 +225,14 @@ protected ChainingMetadataResolver initChainingMetadataResolver(final MetadataRe
         }
         return metadataManager;
     }
+
     @Override
     protected BaseClient<SAML2Credentials, SAML2Profile> newClient() {
-        return new SAML2Client(this.configuration.clone());
+        try {
+			return new SAML2Client(this.configuration.clone());
```
leleuj: Why is the `configuration` clones here?

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
```diff
@@ -15,285 +15,155 @@
...
-        this.identityProviderMetadataPath = identityProviderMetadataPath;
-        this.identityProviderEntityId = identityProviderEntityId;
-        this.serviceProviderEntityId = serviceProviderEntityId;
-
-        CommonHelper.assertNotBlank("keystorePath", this.keystorePath);
-        CommonHelper.assertNotBlank("keystorePassword", this.keystorePassword);
-        CommonHelper.assertNotBlank("privateKeyPassword", this.privateKeyPassword);
-        CommonHelper.assertNotBlank("identityProviderMetadataPath", this.identityProviderMetadataPath);
-
-        final BasicSignatureSigningConfiguration config = DefaultSecurityConfigurationBootstrap.buildDefaultSignatureSigningConfiguration();
```
leleuj: This is done in the abstract SAML2 configuration, right?

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
```diff
@@ -15,285 +15,155 @@
...
-                                    final String privateKeyPassword, final String identityProviderMetadataPath,
-                                    final String identityProviderEntityId, final String serviceProviderEntityId) {
-        this.keystorePath = keystorePath;
-        this.keystorePassword = keystorePassword;
-        this.privateKeyPassword = privateKeyPassword;
-        this.identityProviderMetadataPath = identityProviderMetadataPath;
-        this.identityProviderEntityId = identityProviderEntityId;
-        this.serviceProviderEntityId = serviceProviderEntityId;
-
-        CommonHelper.assertNotBlank("keystorePath", this.keystorePath);
```
leleuj: Where / why  are these checks gone?

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
```diff
@@ -1,145 +1,151 @@
...
+ * Responsible for loading a private key from a JKS keystore and returning the corresponding {@link Credential}
+ * opensaml object.
+ *
+ * It can be used as-is; or a subclass can be used instead, providing different constructor arguments for convenience.
+ * It is up to subclasses to provide an adequate input stream to the constructor, which should contain data of the JKS.
+ * 
+ * @author Misagh Moayyed
+ * @since 1.8.0
+ */
+public abstract class KeyStoreCredentialProvider implements CredentialProvider {
```
leleuj: What has been changed here, except the formating?

leleuj:
Several new comments. My review was really focused on the existing classes, especially `Clients` which should be changed as little as possible in its logic. And SAML ones as well. that being fixed, I think we could merge. After @mmoayyed 's code review.

I notice the indentation is often messy, but this is not a problem, I can fix it later.

jkacer:
Hi, I've seen the new comments. I'm quite busy at the moment and will process them later, maybe after Misagh's review. Sorry for the indentation, it looked just fine in my IDE, maybe it's a tab/spaces problem.

garpinc:
Strikes me that this is along the lines of [this pull request](https://github.com/pac4j/pac4j/pull/519
). I also initialize from a database (in my case a semantic database) and in addition I store the service provider metadata to database. From this experience I don't think pac4j really needs to know where the configuration data is coming from or where it's going to. The only property that was problematic previously was the metadata files and I think i've covered that with the abstraction.

jkacer:
Hi again!
Finally I can work on this again. I have brought my branch up-to-date with PAC4J master, which involved some more changes.
I'm going add my comments to yours, one by one. I've already fixed some smaller things, the big one with is Clients still to be discussed.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
jkacer: Done

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
jkacer: Done

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DatabaseSAML2ClientConfiguration.java
jkacer: Done, also on some more places.

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DatabaseSAML2ClientConfiguration.java
jkacer: These 4 methods are setter/getters for keystoreBinaryData and identityProviderMetadata. These two attributes are returned by getResolvedKeystoreData() and getResolvedIdentityProviderMetadata() - called in SAML2Client.

You are right that these setters/getters are not called anywhere. I just put them there to have setters/getters for the fields but I can remove them without any impact.

They are not annotated with @Override because they are introduced in this class, not a parent class.

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/dao/impl/SpringJdbcTemplateSamlClientDaoImpl.java
jkacer: Done

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
jkacer: I see the whole newClient() method has gone so it's not relevant anymore. However, cloning clone() was not from me.

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
jkacer: Only the last part - BasicSignatureSigningConfiguration and 5 related attributes loaded from it - are in the abstract configuration. The remaining 6 attributes stayed here. The deletion of the 6 assignments was a mistake - they are already back, together with the assertions. I am sorry for that.

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
jkacer: Please see the comment above - it was a mistake, already fixed.

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
jkacer: Changes in commit 033ee5303c0ea7c0f5c0e5445bff1e4c670f23a7:
The class is now abstract and has 2 sub-classes. The constructor has been changed and little JavaDoc has been added.

leleuj:
Cool to see you back!

For version 1.9, I've spent a lot of time cleaning the source code (-15%) so I come to the conclusion that the pac4j project should stay as small as possible (it's already more than 33k lines of source code). My feeling is that your specific modules should go into a specific extension you would maintain on your own and that pac4j would reference from its home page. Of course, some changes would be necessary in the pac4j project to make your extension possible. Currently, it's just a matter of taking out your new Maven modules from the pull request and add them in a new project of your own.

Sounds good?

By the way, I'm about to merge this pull request which may interest you: https://github.com/pac4j/pac4j/pull/519

jkacer:
Hi Jerome!

 I'm OK with this approach. There's just a small issue with the changes in PAC4J's own classes I have made in order to make the extension possible.

So I would propose the following:
- I would finish this MR as we initially planned
- You would review it and merge
- Right after the merge, you would remove the new PAC4J-SAML-DB-Config from it completely
- I would take the module and put on GitHub separately, referencing a stable (not snapshot) version of PAC4J, maybe I could do that for 1.8.x already

The purpose of this is not to lose the changes I have made outside PAC4J-SAML-DB-Config. What do you think? Thanks!

leleuj:
Having a separate extension will make our lifes easier because I will only focus on the changes targeting pac4j and you'll be a little more free to do what you want in your own repository.

I'm in line with the process. I have two remarks though: I'll focus my review on the pac4j changes and why merging a module to remove it just after? Instead of merging, I'll send you a GO and you'll remove the specific modules before I merge. Doesn't it make more sense?

jkacer:
Yes, it makes perfect sense, I agree.

I've just checked in a commit that should solve all your remaining comments from Dec 1:
The base Clients class is almost reverted to its original state. Removed loadClientsInternal() from Clients and only kept it in DatabaseSAML2Clients. Restored internalInit() in Clients and overrode this method in DatabaseSAML2Clients. Various findXxx() methods now call init() again, getClients() does not. DatabaseSAML2Clients does call init() in getClients() because it must assure the result is initialized - it is not just set via a setter or constructor.

Travis is still running but I expect it to pass; if not, I'll fix that.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
jkacer: Fixed in commit e16dc87f373a7c764c10829ae8b15d7cb0a5909b.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
jkacer: Fixed in commit e16dc87f373a7c764c10829ae8b15d7cb0a5909b.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
jkacer: Fixed in commit e16dc87f373a7c764c10829ae8b15d7cb0a5909b.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
jkacer: Fixed in commit e16dc87f373a7c764c10829ae8b15d7cb0a5909b.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
jkacer: Fixed in commit e16dc87f373a7c764c10829ae8b15d7cb0a5909b.

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DatabaseSAML2Clients.java
jkacer: Not changed. For the database-loaded clients, init() must be called here, IMO. Otherwise you would always return null here. Any called would get null, plus all inherited findXxx() methods would not find anything.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -179,14 +190,20 @@ public void setClients(final List<Client> clients) {
     public void setClients(final Client... clients) {
         this.clients = Arrays.asList(clients);
     }
-
+    
+    /**
+     * Returns initialized clients.
```
leleuj: In fact, this comment is misleading: getting clients via `getClients()` won't trigger any initialization while it will with the `findXX` methods (inline `init` call).

pac4j-saml-config-db/pom.xml
```diff
@@ -0,0 +1,144 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
leleuj: I definitely give up on headers and add a NOTICE file in pac4j.

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DatabaseSAML2ClientConfiguration.java
```diff
@@ -0,0 +1,224 @@
...
```
leleuj: I'm not sure this Javadoc really brings value: `@Override` already retrieves the "parent" Javadoc.

pac4j-saml-config-db/src/main/java/org/pac4j/saml/dbclient/DatabaseSAML2Clients.java
```diff
@@ -0,0 +1,133 @@
...
+     * 
+     * The clients are first loaded from the database, then set up as usual.
+     */
+    @Override
+    protected void internalInit() {
+    	final Set<String> names = new HashSet<>();
+        final List<Client> loadedClients = loadClientsInternal();
+        CommonHelper.assertNotNull("loadedClients", loadedClients);
+    	
+        for (final Client client : loadedClients) {
```
leleuj: We already have this check in the `Clients` class: I think we could share some code here.

pac4j-saml-config-db/src/test/java/org/pac4j/saml/dbclient/dao/impl/CachingSamlClientDaoImplTest.java
```diff
@@ -0,0 +1,167 @@
...
+import org.pac4j.saml.dbclient.dao.api.DbLoadedSamlClientConfigurationDto;
+import org.pac4j.saml.dbclient.dao.api.SamlClientDao;
+
+
+/**
+ * Unit test of {@link CachingSamlClientDaoImpl}.
+ * 
+ * @author jkacer
+ */
+public class CachingSamlClientDaoImplTest {
```
leleuj: Tests classes end with `Tests` in pac4j, you may or may not follow this convention on your own.

pac4j-saml/src/main/java/org/pac4j/saml/client/AbstractSAML2ClientConfiguration.java
```diff
@@ -0,0 +1,350 @@
+/*
```
leleuj: Please remove the header.

leleuj:
I just merged: https://github.com/pac4j/pac4j/pull/519, you should sync with master.

jkacer:
Ufff, this merge is a real nightmare but I'm working on it...

leleuj:
Thanks for your tenacity on this. Keep me posted...

jkacer:
Hi! It seems the last changes are a stop sign for my changes. There are many conflicts - on the Java source code level (that's solvable) but mainly on the conceptual level. I mean the changes in the configuration classes and various resolvers. I've come to a point where changes necessary to make it build and pass tests would be awful and not make much sense from the conceptual point of view.

So, I will rather close this pull request and wait for PAC4J 1.9 release. I will then prepare the extension as a separate project of mine and all changes necessary in PAC4J itself in a new pull request. I'm sorry about that but it doesn't make much sense to continue like this.

leleuj:
No problem. Feel free to reopen a new pull request whenever you want...

Issue #355, opened 2015-11-15
JWT  "Illegal key size"


gschrader:
I was trying to use the JwtGenerator however I'm getting an exception, even running the pac4j jwt unit tests I get the same exception:

```
Running org.pac4j.jwt.JwtTests
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.573 sec <<< FAILURE!
testGenerateAuthenticate(org.pac4j.jwt.JwtTests)  Time elapsed: 0.538 sec  <<< ERROR!
org.pac4j.core.exception.TechnicalException: Cannot generate JWT
	at org.pac4j.jwt.profile.JwtGenerator.generate(JwtGenerator.java:91)
	at org.pac4j.jwt.JwtTests.testGenerateAuthenticate(JwtTests.java:51)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:229)
[...]

PR #356, opened 2015-11-16
add support for unencrypted JWT

gschrader:
From https://github.com/pac4j/pac4j/issues/355, this basically is from the example [here](http://connect2id.com/products/nimbus-jose-jwt/examples/jwt-with-hmac) which does not encrypt the payload but still signs it.

You mention documentation in the issue, can you point to where the existing JWT documentation is, or are you only referring to javadoc?

### DIFF
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
@@ -60,13 +60,13 @@ public class JwtAuthenticator implements TokenAuthenticator {
         SignedJWT signedJWT = null;
 
         try {
+            // Parse the token
             JWT jwt = JWTParser.parse(token);
 
             if (jwt instanceof SignedJWT) {
                 signedJWT = (SignedJWT) jwt;
             } else if (jwt instanceof EncryptedJWT) {
-                // Parse the JWE string
-                JWEObject jweObject = JWEObject.parse(token);
+                JWEObject jweObject = (JWEObject) jwt;
 
                 // Decrypt with shared key
                 jweObject.decrypt(new DirectDecrypter(this.secret.getBytes("UTF-8")));
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
@@ -18,8 +18,7 @@ package org.pac4j.jwt.credentials.authenticator;
 import com.nimbusds.jose.JWEObject;
 import com.nimbusds.jose.crypto.DirectDecrypter;
 import com.nimbusds.jose.crypto.MACVerifier;
-import com.nimbusds.jwt.JWTClaimsSet;
-import com.nimbusds.jwt.SignedJWT;
+import com.nimbusds.jwt.*;
 import org.pac4j.core.exception.CredentialsException;
 import org.pac4j.core.exception.TechnicalException;
 import org.pac4j.core.profile.ProfileHelper;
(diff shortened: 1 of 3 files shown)


pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
```diff
@@ -39,9 +39,15 @@
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
     private final String secret;
+    private final boolean encrypted;
```
leleuj: Encryption should happen by default for security reasons and backward compatibility.

leleuj: You should have `= true`.

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
```diff
@@ -70,23 +75,27 @@ public String generate(final U profile) {
 
             // claims
             final JWTClaimsSet claims = builder.build();
+
             // signed
-            final SignedJWT signedJWT = new SignedJWT(new JWSHeader(JWSAlgorithm.HS256), claims);
+            final SignedJWT signedJWT = new SignedJWT(new JWSHeader(JWSAlgorithm.HS256, JOSEObjectType.JWT, null, null, null, null, null, null, null, null, null, null, null), claims);
```
leleuj: I don't believe a method with so many parameters exist ;-)

leleuj: From this example: http://connect2id.com/products/nimbus-jose-jwt/examples/jwt-with-rsa-signature, this would be the source code to just sign the token:

```
SignedJWT signedJWT = new SignedJWT(new JWSHeader(JWSAlgorithm.RS256), claims);
signedJWT.sign(signer);
return signedJWT.serialize();
````

So your algorithm looks good except the update of the `SignedJWT` constructor, I don't see the necessity.

File: pac4j-jwt/src/test/java/org/pac4j/jwt/JwtTests.java [hunk 3]
```diff
@@ -62,6 +72,14 @@ public void testGenerateAuthenticate() {
         assertEquals(profile.getVerified(), fbProfile.getVerified());
     }
 
+    private FacebookProfile createProfile() {
+        final FacebookProfile profile = new FacebookProfile();
```
leleuj: The `JwtAuthenticator` implements `TokenAuthenticator` which expects at least a `HttpProfile`, that's why it doesn't work with `CommonProfile` (which is an ancestor).

leleuj:
Nice try. See my comments for the required updates, but I think we'll be good for the merge after that. Thanks

pac4j-jwt/src/test/java/org/pac4j/jwt/JwtTests.java
gschrader: Ah ok I didn't see HttpProfile, that works for me.

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
gschrader: It is true by default, I'll rework the constructors so it's clear though.

leleuj:
+1 I'll merge tomorrow unless there is a concern

The "clients and authenticators" documentation is here: https://github.com/pac4j/pac4j/wiki/Clients, but I was thinking about dropping some line in the Javadoc: "JSON tokens are encrypted by default, but this may be disabled using this constructor..."

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -61,14 +60,22 @@ public void validate(TokenCredentials credentials) {
...
-
-            // Extract payload
-            signedJWT = jweObject.getPayload().toSignedJWT();
+            JWT jwt = JWTParser.parse(token);
+
+            if (jwt instanceof SignedJWT) {
+                signedJWT = (SignedJWT) jwt;
+            } else if (jwt instanceof EncryptedJWT) {
+                // Parse the JWE string
+                JWEObject jweObject = JWEObject.parse(token);
```
mmoayyed: Wasn't the token parsed once? can you not get the object directly from `jwt`?

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -61,14 +60,22 @@ public void validate(TokenCredentials credentials) {
...
+
+                // Decrypt with shared key
+                jweObject.decrypt(new DirectDecrypter(this.secret.getBytes("UTF-8")));
```
mmoayyed: Should this be hardcoded as UTF-8?

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
gschrader: I initially thought that as well, I however couldn't find any method on EncryptedJWT to get the jwt.  After more exploration, EncryptedJWT is a JWEObject so yeah it doesn't need to parsed again. I'll commit a fix for this.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
gschrader: Good question, I left that as it was before.

leleuj: @mmoayyed : what did you expect? A `encoding` property which can be set? What would be the use case not to use `UTF-8`?

PR #358, opened 2015-11-18
Add pac4j-config module and ConfigPropertiesFactory

leleuj:


leleuj:
Will merge unless there is a concern...

Issue #380, opened 2015-12-01
FacebookClient should have an appsecret_proof property


sbeitzel:
Facebook allows a developer to select an option "Require app secret for server API calls". If this option is selected, then Facebook will refuse a request for profile info with an error, e.g.:

25317 [http-nio-8080-exec-8] DEBUG org.pac4j.oauth.client.BaseOAuthClient  - response code : 400 / response body : {"error"
:{"message":"API calls from the server require an appsecret_proof argument","type":"GraphMethodException","code":100,"fbtra
ce_id":"FFioji\/LajQ"}}

leleuj:
"Require app secret for server API calls" means that the appplication secret needs to be passed along the access token to get the user profile, right?
Indeed, the `FacebookClient` should have an `appsecretProof` property and the appropriate getter and setter and send the application password when the property is `true`.

Would you mind submitting a pull request for this little change?

sbeitzel:
Yes, I will do that. I made the change and verified that Facebook is happy with it, but I cannot say for sure that it is a complete solution since I still have not managed to complete an end-to-end run successfully.

PR #383, opened 2015-12-02
Add appsecret_proof parameter to request if it is specified and needed.

sbeitzel:
This addresses issue #380 .

### DIFF
--- a/pac4j-oauth/src/main/java/org/pac4j/oauth/client/FacebookClient.java
+++ b/pac4j-oauth/src/main/java/org/pac4j/oauth/client/FacebookClient.java
@@ -15,9 +15,14 @@
  */
 package org.pac4j.oauth.client;
 
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+
+import com.fasterxml.jackson.databind.JsonNode;
 import org.apache.commons.lang3.StringUtils;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.exception.HttpCommunicationException;
+import org.pac4j.core.exception.TechnicalException;
 import org.pac4j.core.util.CommonHelper;
 import org.pac4j.oauth.client.exception.OAuthCredentialsException;
 import org.pac4j.oauth.profile.JsonHelper;
@@ -33,8 +38,6 @@ import org.scribe.model.Response;
 import org.scribe.model.SignatureType;
 import org.scribe.model.Token;
 import org.scribe.oauth.StateOAuth20ServiceImpl;
-
-import com.fasterxml.jackson.databind.JsonNode;
 
 /**
  * <p>This class is the OAuth client to authenticate users in Facebook.</p>
@@ -84,7 +87,7 @@ public class FacebookClient extends BaseOAuth20Client<FacebookProfile> {
     
     protected StateApi20 api20;
 
-    protected String appsecretProof = null;
+    protected String appsecret = null;
     
     public FacebookClient() {
     }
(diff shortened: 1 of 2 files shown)


pac4j-oauth/src/main/java/org/pac4j/oauth/client/FacebookClient.java
```diff
@@ -81,6 +83,8 @@
     protected boolean requiresExtendedToken = false;
     
     protected StateApi20 api20;
+
+    protected String appsecretProof = null;
```
leleuj: Why do we need a new value? Cannot we use the `secret` and just have a boolean `useAppsecretProof` to say if we must send it?

sbeitzel: Facebook defines three different bits of data: app ID, app secret, and client token. I am not conversant with the OAUTH protocol nor with the internals of pac4j. I am only just able to understand some of how Shiro works. None of the documentation is at all clear to one such as me; it may be plenty useful to someone who already has the whole model of OAUTH, pac4j, and Shiro loaded mentally and just needs a reminder but it is frustratingly vague to one for whom those systems are tangential to the work at hand.

That said, I looked at the code at line 145:
            url = CommonHelper.addParameter(url, OAuthConstants.CLIENT_SECRET, this.secret);
and since that was already in place I did not assume that it was incorrect. Is it, in fact, incorrect? OAuthConstants.CLIENT_SECRET is the string "client_secret" and if that parameter is not actually used by the Facebook OAUTH protocol, then the most straightforward change would be to add a new constant, OAuthConstants.FB_APPSECRET_PROOF for the "appsecret_proof" parameter and use that instead. I am unaware of whether any other OAUTH services use a parameter of that name; if not, it may make sense to define the constant within FacebookClient so as to reduce confusion, having a Facebook-only parameter sitting in a global table of OAUTH constants.

Are you aware of the answers to any or all of these questions?

leleuj: Reading that: https://www.sammyk.me/the-single-most-important-way-to-make-your-facebook-app-more-secure, I understand the appsec_proof is another value than the app secret.

Taking a closer look at your code, I see that you added this parameter on the exchange token url (it's not linked to the profile retrieval), which is good. But you should also add it to the profile url parameter https://github.com/narrative-technologies/pac4j/blob/develop/pac4j-oauth/src/main/java/org/pac4j/oauth/client/FacebookClient.java#L128

I guess you need some private method here to add the appsec_proof if it is not null to an url.

sbeitzel: Oh, I see. Okay. I have several other pressing tasks today, but may have something better for review in 8 to 12 hours.

pac4j-oauth/src/main/java/org/pac4j/oauth/client/FacebookClient.java
```diff
@@ -201,6 +212,50 @@ protected boolean hasBeenCancelled(final WebContext context) {
...
+     * @param token the application token we pass back and forth
+     * @return URL with the appsecret_proof parameter added
+     */
+    protected String computeAppSecretProof(String url, Token token) {
+        try {
+            Mac sha256_HMAC = Mac.getInstance("HmacSHA256");
+            SecretKeySpec secret_key = new SecretKeySpec(this.secret.getBytes("UTF-8"), "HmacSHA256");
+            sha256_HMAC.init(secret_key);
+            String proof = org.apache.commons.codec.binary.Hex.encodeHexString(sha256_HMAC.doFinal(token.getToken().getBytes("UTF-8")));
+            url = CommonHelper.addParameter(url, APPSECRET_PARAMETER, proof);
```
sbeitzel: This assignment is not really necessary and we could just return the String returned by CommonHelper.addParameter(...). This is actually a leftover from debugging.

leleuj:
Looks good to me. Were you able to test it?

sbeitzel:
I have successfully authenticated with Facebook, both with and without the appsecret_proof parameter.

shiro.ini thus looks like this:
  facebookClient = org.pac4j.oauth.client.FacebookClient
  facebookClient.key = <app ID, found in the app dashboard on Facebook>
  facebookClient.secret = <app secret, found in the app dashboard on Facebook>
  facebookClient.useAppSecretProof = true
  facebookClient.name = Facebook

leleuj:
Perfect. Many thanks.

Issue #386, opened 2015-12-04
pac4j-saml class loader issue


garpinc:
In the org.pac4j.core.util.CommonHelper class in method org.pac4j.core.util.CommonHelper.getInputStreamFromName(String) it calls CommonHelper.class.getResourceAsStream(path);

I am running this is an osgi environment and hence this doesn't return anything.

Can you change this to 
Thread.currentThread().getContextClassLoader().getResourceAsStream(path);

This is a blocker on my end so if you could get this in snapshot build and let me know it would be awesome.. While you're at it instead or in addition to "resource:" can you allow spring's "classpath:" for consistency with the rest of my spring configuration.

Thanks,
Garry

leleuj:
I plan to release pac4j v1.8.1 next week, so we need to do that before. And yes, we can add a "classpath:" support as well. Would you mind submitting a pull request for both changes? Thanks

garpinc:
I will try.. Right now my priority is getting saml working and that's not going well as per my other ticket.

leleuj:
Closing this one. Feel free to re-open it if necessary.

PR #387, opened 2015-12-08
#386 Fix classloading for osgi and add classpath: syntax

garpinc:
Here is the pull request


garpinc:
I never done a pull request before so not sure if I did it right

leleuj:
The build was ok before as you can see on Travis, in the "Build history" (https://travis-ci.org/pac4j/pac4j/builds).
You must take a closer look at the tests. I'm pretty sure they fail because of the change.

garpinc:
Added a new file.. will create a different pull request

PR #390, opened 2015-12-09
I modified resource: to act the same as it did before

garpinc:
I will leave it up to you the design decision. Basically class.getResourceAsStream and classloader.getResourceAsStream act differently as per http://www.javaworld.com/article/2077352/java-se/smartly-load-your-properties.html. Hence if people have previously used resource: they will expect it to remain the same so I left it the same. However that will mean classpath: syntax does something slightly different than resource:


pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -195,18 +195,33 @@ public static String toString(final Class<?> clazz, final Object... args) {
...
+    	int prefixEnd = name.indexOf(":");
+    	String prefix = null;
+    	String path = name;
+    	if (prefixEnd != -1) {
+    		prefix = name.substring(0,prefixEnd);
+    		path = name.substring(prefixEnd+1);
+    	}
+    	
+    	switch (prefix) {
+		case "resource":
```
leleuj: The indentation is a bit messy here: are you sure you use spaces only?

leleuj: Can we update the constant to reuse it? Or it is used somewhere else?

leleuj:
I'm in line with you to keep the old behaviour. So, for "resource:", adding a leading  slash if missing, means that all search are considered for absolute path, right?

On the other side, for "classpath:" (which uses ClassLoader.getResourceAsStream), we don't add a leading slash as path are absolute. What happens if we pass a path with a leading slash? Does it work?

Thanks for your work and tenacity on this.

leleuj:
Any update on this?

garpinc:
I have a big check in on my end so I'll do a pull request when that's done.. Aside from above.. I need to do a few more complicated things.. My file will not be on the class path or on the file system.. It will be in a service provided by my application. Also the service provider file also will not be put on the filesystem and instead it will be put also put in a file repository that my application provides as a service. Hence in my opinion simple paths are not enough.. What I've done is mirror http://docs.spring.io/spring-framework/docs/2.5.6/api/org/springframework/core/io/Resource.html
in pac4j. I would have used it directly but I'm assuming you want no spring dependencies in pac4j core. Hence I simply copies the classes and created a wrapper class for it in the spring-security-pac4j bundle which allows you to wrap spring resource objects as pac4j resource objects.. So then in my application I create Resource implementations that either read or write from the end points that my application provides. I hope that this is going to work for the code base.

leleuj:
Indeed, the idea is to limit the number of dependencies in the core. So no spring dependency here. Let's make your specific component work and we'll think about where it can go. Keep us posted.

leleuj:
Closing for now. Feel free to re-open when you are ready.

PR #397, opened 2015-12-15
Code Quality Improvement - @Override annotation should be used on any method overriding

civanyp:
This pull request is focused on resolving occurrences of Sonar rule
squid:S1161 - "@Override annotation should be used on any method overriding". You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S1161.

Please let me know if you have any questions.

Christian

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/http/DefaultAjaxRequestResolver.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/http/DefaultAjaxRequestResolver.java
@@ -27,9 +27,7 @@ import org.pac4j.core.context.WebContext;
  */
 public class DefaultAjaxRequestResolver implements AjaxRequestResolver, HttpConstants, Pac4jConstants {
 
-    /**
-     * {@inheritDoc}
-     */
+    @Override
     public boolean isAjax(final WebContext context) {
         final boolean xmlHttpRequest = AJAX_HEADER_VALUE.equalsIgnoreCase(context.getRequestHeader(AJAX_HEADER_NAME));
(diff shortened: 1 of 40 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/transport/Pac4jHTTPPostEncoder.java
```diff
@@ -92,6 +93,7 @@ protected void doDestroy() {
     }
 
     /** {@inheritDoc} */
```
mmoayyed: You may also want to remove `/** {@inheritDoc} */` as well, for overrides.

civanyp:
Done! Removed /** {@inheritDoc} */ for overrides change.

pac4j-saml/src/main/java/org/pac4j/saml/transport/Pac4jHTTPPostEncoder.java
civanyp: Done! Removed /** {@inheritDoc} */ for overrides change.

mmoayyed:
Great, :+1:

pac4j-core/src/main/java/org/pac4j/core/http/DefaultAjaxRequestResolver.java
```diff
@@ -30,6 +30,7 @@
     /**
```
leleuj: I think the use of `@Override` allows us not to add `{@inheritDoc}`, doesn't it? The Javadoc comment can be safely removed.

civanyp: Done! Removed all Javadoc comment {@inheritDoc} for @Override change.

leleuj:
Thanks

Issue #393, opened 2015-12-13
JwtAuthenticator uses the same secret for signing/encryption


mmoayyed:
I would caution pac4j to not use a shared secret when dealing with JWT. There should be separate keys for signing and encryption. Was there a specific reason to use a shared key?

leleuj:
"Making things easier", but it's not a good reason. I will add two secrets and deprecate the old one, for better security.

mmoayyed:
Thanks. I would vote that we keep using the same secret that we have for both signing and encryption, if the encryption key is undefined. Just to sort of keep the same behavior. Encrypted keys are not always a requirement.

PR #400, opened 2015-12-15
Allow jwt generator/authenticator to use separate secrets

mmoayyed:
Closes https://github.com/pac4j/pac4j/issues/393

### DIFF
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
@@ -40,20 +40,30 @@ public class JwtAuthenticator implements TokenAuthenticator {
 
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
-    private String secret;
+    private String signingSecret;
+    private String encryptionSecret;
 
     public JwtAuthenticator() {}
 
-    public JwtAuthenticator(final String secret) {
-        this.secret = secret;
+    public JwtAuthenticator(final String signingSecret) {
+        this(signingSecret, signingSecret);
+    }
+
+    /**
+     * @since 1.8.2
+     */
+    public JwtAuthenticator(final String signingSecret, final String encryptionSecret) {
+        this.signingSecret = signingSecret;
+        this.encryptionSecret = encryptionSecret;
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
-    public void validate(TokenCredentials credentials) {
-        CommonHelper.assertNotBlank("secret", secret);
+    public void validate(final TokenCredentials credentials) {
+        CommonHelper.assertNotBlank("signingSecret", signingSecret);
+
 
         final String token = credentials.getToken();
         boolean verified = false;
@@ -61,15 +71,20 @@ public class JwtAuthenticator implements TokenAuthenticator {
 
         try {
             // Parse the token
-            JWT jwt = JWTParser.parse(token);
+            final JWT jwt = JWTParser.parse(token);
 
             if (jwt instanceof SignedJWT) {
                 signedJWT = (SignedJWT) jwt;
             } else if (jwt instanceof EncryptedJWT) {
-                JWEObject jweObject = (JWEObject) jwt;
+                final JWEObject jweObject = (JWEObject) jwt;
 
                 // Decrypt with shared key
-                jweObject.decrypt(new DirectDecrypter(this.secret.getBytes("UTF-8")));
+                String secret = this.encryptionSecret;
+                if (CommonHelper.isBlank(secret)) {
+                    logger.warn("Encryption secret is not configured. Falling back to signing secret");
+                    secret = this.signingSecret;
+                }
+                jweObject.decrypt(new DirectDecrypter(secret.getBytes("UTF-8")));
 
                 // Extract payload
                 signedJWT = jweObject.getPayload().toSignedJWT();
@@ -77,7 +92,7 @@ public class JwtAuthenticator implements TokenAuthenticator {
                 throw new TechnicalException("unsupported unsecured jwt");
             }
 
-            verified = signedJWT.verify(new MACVerifier(this.secret));
+            verified = signedJWT.verify(new MACVerifier(this.signingSecret));
         } catch (final Exception e) {
             throw new TechnicalException("Cannot decrypt / verify JWT", e);
         }
@@ -96,11 +111,31 @@ public class JwtAuthenticator implements TokenAuthenticator {
         }
     }
 
-    public String getSecret() {
-        return secret;
+    /**
+     * @since 1.8.2
+     */
+    public String getSigningSecret() {
+        return signingSecret;
     }
 
-    public void setSecret(String secret) {
-        this.secret = secret;
+    /**
+     * @since 1.8.2
+     */
+    public void setSigningSecret(final String signingSecret) {
+        this.signingSecret = signingSecret;
+    }
+
+    /**
+     * @since 1.8.2
+     */
+    public String getEncryptionSecret() {
+        return encryptionSecret;
+    }
+
+    /**
+     * @since 1.8.2
+     */
+    public void setEncryptionSecret(final String encryptionSecret) {
+        this.encryptionSecret = encryptionSecret;
     }
 }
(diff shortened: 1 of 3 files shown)


pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -40,20 +40,30 @@
...
-    private String secret;
+    private String signingSecret;
+    private String encryptionSecret;
 
     public JwtAuthenticator() {}
 
-    public JwtAuthenticator(final String secret) {
-        this.secret = secret;
+    public JwtAuthenticator(final String signingSecret) {
+        this(signingSecret, signingSecret);
```
leleuj: My feeling is that we should output a warning to recommend using the new constructor which is safer with 2 secrets. What do you think?

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
```diff
@@ -38,27 +46,28 @@
...
-     * Initializes the generator that will create JWT tokens that are both signed and encrypted.
+     * Initializes the generator that will create JWT tokens that are signed but not encrypted.
      *
-     * @param secret The secret. Must be at least 256 bits long and not {@code null}
+     * @param signingSecret The signingSecret. Must be at least 256 bits long and not {@code null}
      */
-    public JwtGenerator(String secret) {
-        this.secret = secret;
+    public JwtGenerator(final String signingSecret) {
+        this(signingSecret, null);
```
leleuj: I guess there is a problem with the `null`: it should be `signingSecret`, right?

leleuj: I would expect a warning here also.

leleuj: OK. I see why you use `null` to replace `encrypted`, but this generates a problem: if someone upgrades to pac4j v1.8.2, data will stop being encrypted when keeping the same constructor (the one with only one parameter).
So we must keep the current constructors:

```
public JwtGenerator(final String secret) {
   this(secret, true);
}
public JwtGenerator(final String secret, final boolean encrypted) {
  this.signingSecret = secret;
  if (encrypted) {
    this.encryptionSecret = secret;
  }
}
```

and add a new one:

```
public JwtGenerator(final String signingSecret, final String encryptionSecret) {
  this.signingSecret = signingSecret;
  this.encryptionSecret = encryptionSecret;
}
```

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -96,11 +111,31 @@ public void validate(TokenCredentials credentials) {
...
-    public String getSecret() {
-        return secret;
+    /**
+     * @since 1.8.2
+     */
+    public String getSigningSecret() {
+        return signingSecret;
     }
 
-    public void setSecret(String secret) {
```
leleuj: I would keep the `setSecret` at least and make it set both the signing and encryption secrets.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -61,15 +71,20 @@ public void validate(TokenCredentials credentials) {
...
         try {
             // Parse the token
-            JWT jwt = JWTParser.parse(token);
+            final JWT jwt = JWTParser.parse(token);
 
             if (jwt instanceof SignedJWT) {
                 signedJWT = (SignedJWT) jwt;
             } else if (jwt instanceof EncryptedJWT) {
-                JWEObject jweObject = (JWEObject) jwt;
+                final JWEObject jweObject = (JWEObject) jwt;
```
leleuj: Here I would test that the `encryptionSecret` and not fallback to the `signingSecret` as the old constructor should have set both.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -61,15 +71,20 @@ public void validate(TokenCredentials credentials) {
...
 
                 // Decrypt with shared key
-                jweObject.decrypt(new DirectDecrypter(this.secret.getBytes("UTF-8")));
+                String secret = this.encryptionSecret;
+                if (CommonHelper.isBlank(secret)) {
```
leleuj: I would remove this fallback.

pac4j-jwt/src/test/java/org/pac4j/jwt/JwtTests.java
```diff
@@ -52,7 +52,15 @@ public void testGenerateAuthenticate() {
...
-        final JwtGenerator<FacebookProfile> generator = new JwtGenerator<FacebookProfile>(KEY, false);
+        final JwtGenerator<FacebookProfile> generator = new JwtGenerator<>(KEY);
+        final FacebookProfile profile = createProfile();
+        final String token = generator.generate(profile);
+        assertToken(profile, token);
+    }
+
+    @Test
+    public void testGenerateAuthenticateAndEncrypted() {
+        final JwtGenerator<FacebookProfile> generator = new JwtGenerator<>(KEY, KEY);
```
leleuj: Can we have a test with two different keys to be sure everything is ok?

leleuj:
I have several comments on this PR...

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
mmoayyed: Sure

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
mmoayyed: Note: old constructor only sets both if encrypt=true. Otherwise, the encryption key will be blank.

mmoayyed: Ah never mind. I see what you mean.

leleuj: I guess we should work in the same office, we would save time ;-)

mmoayyed:
Updated pull based on @leleuj's feedback

mmoayyed:
Let's finalize this first before switching back to #401

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -61,15 +80,13 @@ public void validate(TokenCredentials credentials) {
...
+            final JWT jwt = JWTParser.parse(token);
 
             if (jwt instanceof SignedJWT) {
                 signedJWT = (SignedJWT) jwt;
             } else if (jwt instanceof EncryptedJWT) {
-                JWEObject jweObject = (JWEObject) jwt;
-
-                // Decrypt with shared key
-                jweObject.decrypt(new DirectDecrypter(this.secret.getBytes("UTF-8")));
+                final JWEObject jweObject = (JWEObject) jwt;
```
leleuj: There should be a `CommonHelper.assertNotBlank("encryptionSecret", encryptionSecret);` here.

mmoayyed: Done

leleuj: In fact, I meant here, just on this line, because we should only check the existence of the `encryptionSecret` if we need to use it. Sorry if I was not clear...

leleuj: This is my last comment and we're good for a merge...

mmoayyed: You are right. Updated.

mmoayyed:
All feedback should be accounted for. Proceeding to merge so I can switch back #401

PR #401, opened 2015-12-15
Allowed a more generic format for JWTs

### LINKED ISSUE #394: JwtAuthenticator is strongly tied to JwtGenerator
Pac4j only accepts JWTs that are issued by its own generator. This should be more flexible. I should be allowed to use any JWT token I would want with any generator I would want. Pac4j, should just have to instruct me on how to structure the token, rather than requiring me to use its own component. 

The profile builder logic should be smart enough to even handle JWTs that are not created by Pac4j's.

mmoayyed:
Closes #394

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
@@ -64,12 +64,12 @@ public final class ProfileHelper {
 
         logger.info("Building user profile based on typedId {}", typedId);
         try {
-            if (!typedId.contains("#")) {
+            if (!typedId.contains(UserProfile.SEPARATOR)) {
                 final String completeName = determineProfileClassByName("HttpProfile");
                 return buildUserProfileByClassCompleteName(typedId, attributes, completeName);
             }
 
-            final String[] values = typedId.split("#");
+            final String[] values = typedId.split(UserProfile.SEPARATOR);
             final String className = values[0];
 
             final String completeName = determineProfileClassByName(className);
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
@@ -16,8 +16,10 @@
 package org.pac4j.core.profile;
 
 import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.util.Map;
 
+import org.pac4j.core.util.CommonHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


mmoayyed:
This pull also adds support for JWTs that are base64 encoded.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
```diff
@@ -56,55 +58,71 @@ public static boolean isTypedIdOf(final String id, final Class<? extends UserPro
...
-                        logger.error("Cannot build instance", e);
-                    }
-                }
+        if (CommonHelper.isBlank(typedId)) {
+            return null;
+        }
+
+        logger.info("Building user profile based on typedId {}", typedId);
+        try {
+            if (!typedId.contains("#")) {
```
leleuj: `#` is `UserProfile.SEPARATOR`.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
```diff
@@ -56,55 +58,71 @@ public static boolean isTypedIdOf(final String id, final Class<? extends UserPro
...
+                final String completeName = determineProfileClassByName("HttpProfile");
```
leleuj: I like these two new methods: `determineProfileClassByName` and `buildUserProfileByClassCompleteName`, it makes things cleaner!

But I don't think we should define here the default type (`HttpProfile`) as it' a specific need for the JWT mechanism.

pac4j-jwt/pom.xml
```diff
@@ -33,6 +33,10 @@
 			<artifactId>pac4j-core</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>commons-codec</groupId>
```
leleuj: I don't like adding new dependencies and some times ago, I learned that you can handle base64 encoding / decoding just with the JDK:

Example:
```
byte[] message = "hello world".getBytes("UTF-8");
String encoded = DatatypeConverter.printBase64Binary(message);
byte[] decoded = DatatypeConverter.parseBase64Binary(encoded);
```

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
leleuj: I think the code handling the `#` should be in the `JwtAuthenticator`.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -55,12 +60,16 @@ public JwtAuthenticator(final String secret) {
...
         CommonHelper.assertNotBlank("secret", secret);
 
-        final String token = credentials.getToken();
+        String token = credentials.getToken();
         boolean verified = false;
         SignedJWT signedJWT = null;
 
         try {
             // Parse the token
+            if (Base64.isBase64(token)) {
```
leleuj: An interesting idea, but I'm not very enthusiastic about magic "behaviour": don't you think we could add new methods for generic JWT tokens: `validateToken(final String token)` and `validateBase64Token(final String bas64Token)`?

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
mmoayyed: How? I am not sure I follow.

pac4j-jwt/pom.xml
mmoayyed: Yes, except that we might need to check to see if the message is base-64 encoded before we can decode it. This is related to the API change we are discussing below.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
mmoayyed: That depends. If you create the two new methods, there are several side effects. Two of which are:

1. I am forced to call the method on JWTAuthenticator, rather than a generic Authenticator. It would be best if we could keep things separate as much as possible.

I can do:

```
Authenticator a = new JwtAuthenticator()
a.validate()
```

The change would be:

```
JwtAuthenticator a = new JwtAuthenticator()
a.validateToken()

... or ...
a.validateBase64Token()
```


2. If you aim to make things easier for the developer, this should just work. Because otherwise, I'd have declare a dependency in my own project, and make that isBase64 call anyway, which means you will end up with that dependency anyway, and have forced me to call separate API methods. I think the framework should just handle this and protect the developer, but if you are OK with the above concerns, I'll add the methods.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
mmoayyed: Also, I don't think this is specific to JWT. The profile building logic expects all extended profiles to exist in `org.pac4j.oauth.profile.` or it wont work. You will get an exception somewhere down the road. So, I think you should account for the most generic case which is the HttpProfile, and maybe we could put out a warning to say the "typedId does not designate a profile. falling back to the default". This will work with all built-in profiles as well as profile classes I may create in "org.misagh.pac4j.this.that"

leleuj: I follow your reasoning but the `HttpProfile` is not the default user profile, it's the `UserProfile`. So either we change this method signature to allow providing a default user profile, either we handle that in the JWT module. It was towards the latest option as it's a 1.8.x, not a new 1.9.0. That said, the first option could do do the job if we change the method signature.

Does it make more sense?

mmoayyed: Thanks. I understand now and I would like this change to go to 1.8.x. So let's go with your second option. If I am following this correctly, the JwtAuthenticator must capture the subject field from the claims, and handle the separator there, right? 

```
final UserProfile profile = ProfileHelper.buildProfile(claimSet.getSubject(), claimSet.getClaims());
```

The subject needs to be changed to something here that ProfileBuidler can understand with the right separator. Makes sense?

pac4j-jwt/pom.xml
leleuj: Thinking more about this, we cannot add new methods as the `validate` method is the only one called in the authenticator.
I don't think we need base64 encoding / decoding as JWT tokens are already partially base64 encoded and thus URL safe.
So I would remove the `commons-codec` dependency and fix the test.

leleuj:
Notice the build is now broken as you don't decode base64 tokens any more.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
mmoayyed: Does this mean that we should also create a JwtProfile?

pac4j-jwt/pom.xml
mmoayyed: Sure. Copy that.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
leleuj: "the JwtAuthenticator must capture the subject field from the claims, and handle the separator there, right?": yes, you get it right.

Yes, it makes sense to create a specific `JwtProfile` (inheriting from `HttpProfile`) when we get a JWT token which was not previously a pac4j user profile. Indeed, you need to check the subject to see if there is a `#` otherwise you'll provide `JwtProfile.class.getSimpleName() + UserProfile.SEPARATOR`.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
leleuj: I don't thinking creating new methods is really an option if we want to use generic JWT tokens in a normal pac4j flow. Forget this suggestion.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
mmoayyed: Sounds good. I'll update the pull momentarily.

mmoayyed:
Pull is updated.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
```diff
@@ -56,55 +58,69 @@ public static boolean isTypedIdOf(final String id, final Class<? extends UserPro
...
+        }
+        return completeName;
+    }
+
+    private static UserProfile buildUserProfileByClassCompleteName(final String typedId, final Map<String, Object> attributes,
+                                                                   final String completeName)
+            throws Exception {
+        @SuppressWarnings("unchecked")
+        final Constructor<? extends UserProfile> constructor = (Constructor<? extends UserProfile>) Class
+                .forName(completeName).getDeclaredConstructor();
```
mmoayyed: I did not change this, but I don't this makes sense, performance-wise. It would be better if class instances were cached before they were newed.

leleuj: This is a legitimate concern: I don't think we rebuild so many profiles except in the JWT mechanism. What would you do? Store a map of constructors by classes? And invoke `newInstance` on these cached constructors?

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtProfile.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.pac4j.core.profile.CommonProfile;
+
+/**
+ * Represents a user profile based on a JWT.
+ *
+ * @author Misagh Moayyed
+ * @since 1.8.2
+ */
+public class JwtProfile extends CommonProfile {
```
leleuj: I'm wondering if we should inherit from `HttpProfile`, this is the case for all profiles built from data received in direct clients (LDAP, Mongo...)

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
mmoayyed: Yes exactly.

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtProfile.java
mmoayyed: I am for that idea. I don't see much difference between http and jwt profiles, which is why I thought originally to just return/use the http profile for jwt but I think it's best to keep the jwt profile separate and just have it extend http.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
mmoayyed: This is not so easy to do with the current design. In order to cache `Class.forName(completeName)` instances, you need the completeName for all known classes, which may not be available on the classpath. So we can't cache upfront, but we could cache on demand inside the same method via a map.

mmoayyed: You could on the other hand, scan the classpath and find all classes that extend UserProfile and cache those. That might also work, but it might be overkill. On demand caching is slightly better, and less complicated.

mmoayyed:
I updated the JwtProfile to use HttpProfile as its parent. I think that makes more sense.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
leleuj: OK. I just tracked that with the following issue: https://github.com/pac4j/pac4j/issues/405

leleuj:
Great! Thanks for this one.

PR #403, opened 2015-12-15
BasicAuthExtractor should work on Authorization: Basic by default

leleuj:


mmoayyed:
:+1

mmoayyed:
We should also clean up the CAS basic authn integration later

leleuj:
Can you elaborate a little more?

mmoayyed:
Sure.

CAS does this:

```java
 private static final String AUTHZ_HEADER = "Authorization";
...
final BasicAuthExtractor extractor = new BasicAuthExtractor(AUTHZ_HEADER, "Basic", this.getClass().getSimpleName());
```

We should use:

```java
final BasicAuthExtractor extractor = new BasicAuthExtractor(this.getClass().getSimpleName());
```

pac4j-core/src/main/java/org/pac4j/core/context/HttpConstants.java
```diff
@@ -42,6 +42,8 @@
 
     String AUTHORIZATION_HEADER = "Authorization";
 
+    String BASIC_HEADER_PREFIX = "Basic ";
```
mmoayyed: Do we need to check for case? BASIC? basic? Is "Basic" pretty much the standard?

leleuj: According to https://tools.ietf.org/html/rfc2617, it's "Basic ", but I think we could use a `equalsIgnoreCase` in the `BasicAuthExtractor`. No one has ever reported an issue on this...

mmoayyed: Sounds good.

mmoayyed: Let's get this merged. Looks pretty done. I ll work on the CAS side of things based on the snapshot here.

PR #407, opened 2015-12-17
Add an HttpActionAdapter

leleuj:


pac4j-core/src/main/java/org/pac4j/core/http/HttpActionAdapter.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.pac4j.core.context.WebContext;
+
+/**
+ * HTTP action adapter.
+ *
+ * @author Jerome Leleu
+ * @since 1.8.2
+ */
+public interface HttpActionAdapter {
```
mmoayyed: Do we have an impl? How is this used?

leleuj: In Play, see: https://github.com/pac4j/play-pac4j/pull/89/files#diff-0f312e3d8b53e5fb5a98dd7a144a7c2eR34 and with SparkJava as well. Though, the upgrade on Spark is in progress and I will only submit a PR in a few days...

PR #411, opened 2015-12-23
Improvements to pac4j-oidc

papegaaij:
This pull request contains 3 minor improvements to the OIDC implementation, vital for our implementation:
 - Upgrade to a the latest, stable SDK (5.1).
 - Allow setting custom timeouts for all socket operations. The defaults are quite high and can cause a DoS in case of a failing or slow OIDC provider.
 - Give the possibility to use a different ResourceRetriever. We need this to add caching and to work around a problem in the discovery metadata returned my Azure AD. https://login.microsoftonline.com/common/.well-known/openid-configuration contains invalid characters in the issuer URL. This is an error on the side of Azure AD, but at the moment cannot be easily worked around. In a custom ResourceRetriever, the invalid characters can be escaped.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/client/OidcClient.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/client/OidcClient.java
@@ -24,10 +24,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 
-import com.nimbusds.jose.JWSAlgorithm;
-import com.nimbusds.oauth2.sdk.http.DefaultResourceRetriever;
-import com.nimbusds.openid.connect.sdk.claims.IDTokenClaimsSet;
-
 import org.apache.commons.lang3.StringUtils;
 import org.pac4j.core.client.ClientType;
 import org.pac4j.core.client.IndirectClient;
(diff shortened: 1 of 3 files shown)


mmoayyed:
:+1:

pac4j-oidc/src/main/java/org/pac4j/oidc/client/OidcClient.java
```diff
@@ -134,6 +136,11 @@
     /* max clock skew in seconds */
     private int maxClockSkew = DEFAULT_MAX_CLOCK_SKEW;
 
+    /* timeouts for token and userinfo requests */
+    private int connectTimeout = 0;
```
leleuj: What is the meaning of these values? Can we extract these default timeout values in the following class: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/context/HttpConstants.java and reuse them here: https://github.com/pac4j/pac4j/blob/master/pac4j-oauth/src/main/java/org/pac4j/oauth/client/BaseOAuthClient.java#L58 ?

leleuj:
Excellent! Thanks. See my only question before I merge.

BTW, would you mind adding a `MicrosoftOidcClient` with the apprpriate discovery url and `ResourceRetriever` (https://github.com/pac4j/pac4j/issues/415)? It might be another PR.

It will be available in pac4j v1.8.3.

papegaaij:
I've moved the default timeouts to HttpConstants and set them to a more sensible 15s. 0, which was used before this PR means infinite, which is probably not what you want. Also, I've opened up IDTokenValidator to subclasses. The same issuer url is giving trouble with the validator as well.

I'll submit another PR next week with the AzureADOidcClient.

leleuj:
Good. I will merge "as is" and update the default timeouts later on. I think 15s for a connect time out is too much as it lets you reach a computer on the Sun from the Earth ;-) Paris - San Francisco ~ 100ms to connect

leleuj:
See: https://github.com/pac4j/pac4j/pull/419

PR #412, opened 2015-12-23
Updated Google2Client with updated scope values

ParthKolekar:
According the Google's documentation listed at https://developers.google.com/identity/protocols/googlescopes, the OAuth scopes are different from what was listed. 

Updated the file needed for this. 

The older values still work, but the newer values are what are documented further.


leleuj:
Did you successfully test the new values?

ParthKolekar:
Yes, the current values and the updated values are equivalent. 

The Google OAuth Demo on https://developers.google.com/oauthplayground/ does the requests in the same way. 

It also shows the raw headers for requests. The recommended way to request emails and profile is via `https://www.googleapis.com/auth/userinfo.email` and `https://www.googleapis.com/auth/userinfo.profile`.

leleuj:
Excellent! Thanks.

Issue #420, opened 2015-12-29
Issue with CAS server URL starting with "login"


dkefer:
When using a CAS server with URL starting with "login" (e.g. "login.example.com"), the library removes the string "login" when making the request to the CAS server in order to verify the token provided by client (i.e. it tries to connect to ".example.com").

Behaviour observed when using the following dependencies:

```
"org.pac4j" % "pac4j-cas" % "1.8.1-SNAPSHOT",
"org.pac4j" % "pac4j-core" % "1.8.1-SNAPSHOT",
"org.pac4j" % "play-pac4j-java" % "2.0.1-SNAPSHOT",
```

```
<dependency>
   <groupId>org.pac4j</groupId>
   <artifactId>spring-security-pac4j</artifactId>
   <version>1.3.0</version>
</dependency>
<dependency>
   <groupId>org.pac4j</groupId>
   <artifactId>pac4j-cas</artifactId>
   <version>1.7.1</version>
</dependency>
<dependency>
   <groupId>org.pac4j</groupId>
   <artifactId>pac4j-core</artifactId>
   <version>1.7.1</version>
</dependency>
```

leleuj:
Notice pac4j v1.8.1 has been released (no more SNAPSHOTs).

I guess the problem is here: https://github.com/pac4j/pac4j/blob/master/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java#L269

Would you mind submitting a PR for that?

dkefer:
Thanks, I created a PR with a fix which seems to work for me (I hope I understood the idea of that string replacement correctly).

leleuj:
OK. Can you thus a simple test to check it works to protect us against any regression? Using your example...

dkefer:
new PR...https://github.com/pac4j/pac4j/pull/421

PR #421, opened 2015-12-29
fix for FQDNs starting with "login"

dkefer:
My fix proposal for issue https://github.com/pac4j/pac4j/issues/420

### DIFF
--- a/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
@@ -266,7 +266,7 @@ public class CasClient extends IndirectClient<CasCredentials, CasProfile> {
             this.casPrefixUrl += "/";
         }
         if (CommonHelper.isBlank(this.casPrefixUrl)) {
-            this.casPrefixUrl = this.casLoginUrl.replaceFirst("/login$", "/$");
+            this.casPrefixUrl = this.casLoginUrl.replaceFirst("/login$", "/");
         } else if (CommonHelper.isBlank(this.casLoginUrl)) {
             this.casLoginUrl = this.casPrefixUrl + "login";
         }
--- a/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
@@ -266,7 +266,7 @@ public class CasClient extends IndirectClient<CasCredentials, CasProfile> {
             this.casPrefixUrl += "/";
         }
         if (CommonHelper.isBlank(this.casPrefixUrl)) {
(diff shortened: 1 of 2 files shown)


pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
```diff
@@ -266,7 +266,7 @@ protected void initializeClientConfiguration() {
             this.casPrefixUrl += "/";
         }
         if (CommonHelper.isBlank(this.casPrefixUrl)) {
-            this.casPrefixUrl = this.casLoginUrl.replaceFirst("/login", "/");
+            this.casPrefixUrl = this.casLoginUrl.replaceFirst("/login$", "/$");
```
leleuj: I'm surprised by the replacement String: why a dollar at the end?

leleuj:
Can you add a simple test to confirm the fix?

pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
dkefer: dollar in order to match just the end of the regex. Was that the original idea of the line or are there any other legitimate replacement cases?

mmoayyed: :-1: on this. Nobody will remember to do this. We'll need to be smarter about replacing bits in the path. 

Also, this only seems to be a problem if you don't set the casPrefix url? Can we simply set that value and move on?

mmoayyed: We'll need to be smart enough to only replace the "/login" path element and not the one in the actual domain. Asking folks to define their CAS login url with a $ at the end is not the way to go. Parse the path correctly, and take out the bit that is actually needed to be removed rather than relying on the deployer to correctly define it for you.

leleuj: The `$` is not a required convention, it's part of the regexp which is the first parameter of the `replaceFirst` method.

```
String url1 = "http://localhost:8080/cas/login";
System.out.println(url1.replaceFirst("/login$", "/end"));
```
Will output: `http://localhost:8080/cas/end`.

And it works for: `String url1 = "http://loginserver/cas/login";` which is the needed fix.

Though, the proposed fix: `this.casPrefixUrl = this.casLoginUrl.replaceFirst("/login$", "/$");` doesn't work and generates: `Illegal group reference: group index is missing`.

mmoayyed: Ah, gotcha. It certainly looks misleading :)

dkefer:
I had an extra dollar sign in the replacement string - corrected + test written

leleuj:
+1 @mmoayyed : ok for you?

mmoayyed:
Sure. :+1:

PR #426, opened 2016-01-02
Revert "Updated Google2Client with updated scope values"

ParthKolekar:
Reverts pac4j/pac4j#412

I was mistaken in the usage of these values. The old values were actually correct and depreciates the ones I provided. I did not read the documentation correctly.


leleuj:
No problem. One can be wrong. Reverted.

PR #428, opened 2016-01-02
Inject client SAML name into the credentials instead of the default.

peyo:
When the credentials are built by pac4j, it is not with the "real client name" but with the class name : SAML2.class.getSimpleName(). 

This is boring if you specify the name or if you use 2 SAML clients.

This is not the best solution to resolve my problem but it works.
The best solution is to inject the client name when the credentials are built, into the function : #SAML2DefaultResponseValidator.buildSAML2Credentials().

Bye.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
@@ -244,7 +244,12 @@ public class SAML2Client extends BaseClient<SAML2Credentials, SAML2Profile> {
     @Override
     protected SAML2Credentials retrieveCredentials(final WebContext wc) throws RequiresHttpAction {
         final SAML2MessageContext context = this.contextProvider.buildContext(wc);
-        return (SAML2Credentials) this.profileHandler.receive(context);
+        final SAML2Credentials credentials = (SAML2Credentials) this.profileHandler.receive(context);
+        final String clientName = getName();
+        if (CommonHelper.isBlank(clientName)) {
+            credentials.setClientName(clientName);
+        }
+        return credentials;
     }
 
     @Override


leleuj:
We don't plan any work on 1.7 (we are about to start the development of version 1.9). Can you upgrade to version 1.8 where this has been fixed?

peyo:
It's not possible for me to upgrade on 1.8 because I use pac4j with CAS server and it seems that version is not compatible with stable version of CAS. But, I may be wrong.

leleuj:
pac4j should be compatible with any CAS server version unless there is a bug. What's your CAS server version? What's the issue?

peyo:
I use the version 4.1.3 of CAS (with Tomcat 7).
If I use pac4j 1.8.2 it's impossible to start my server : NoClassDefFoundError: org/pac4j/core/client/Mechanism 
"Mechanism" class is referenced in the "cas-server-support-pac4j" jar (version 4.1.3 like CAS).
I make a mistake in my dependencies ?

leleuj:
OK. I get it now: yes, inside the CAS server 4.1, you need pac4j v1.7.x. I will merge your PR, let you test pac4j v1.7.2-SNAPSHOT and if you confirm the fix, I'll cut a release.

Sounds good?

peyo:
It's good (I use source not the SNAPSHOT Jar)
Do you know pac4j works with ADFS 3.0 (it seems not) ? 
Thx

leleuj:
I'm not sure, but I think so. There is a specific documentation for ADFS, but it might outdated: https://github.com/pac4j/pac4j/blob/master/README-ADFS.txt

What issue do you have? Is there a way for me to test your configuration (online ADFS)?

peyo:
Excuse me, it's just a bad ADFS configuration.
The ADFS don't send Subject NameID node and it cause NullPointerException in Pac4J.
I make a PullRequest for avoid it.

leleuj:
Would you mind updating the `README- ADFS` doc with your latest updates? It could help others.

Issue #415, opened 2015-12-26
Add a MicrosoftOidcClient


leleuj:
With the appropriate discovery url:  https://login.microsoftonline.com/common/.well-known/openid-configuration and the appropriate `ResourceRetriever`?

PR #430, opened 2016-01-04
Client for Microsoft Azure AD

papegaaij:
This PR contains the specialized OidcClient for Microsoft Azure AD, as requested in #415.

The main problem is the difference in issuer between what's returned by the discovery and what's in the ID token. Unfortunately, IdTokenValidator does not allow changing issuer validation, so I had to wrap the entire class.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/azuread/AzureAdClient.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/azuread/AzureAdClient.java
@@ -1,0 +1,45 @@
+package org.pac4j.oidc.azuread;
+
+import org.pac4j.oidc.client.OidcClient;
+
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.oauth2.sdk.http.ResourceRetriever;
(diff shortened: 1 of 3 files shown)


leleuj:
Great! Thanks.

leleuj:
@papegaaij : I updated the Azure AD client to have a specific profile returned: https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/main/java/org/pac4j/oidc/client/AzureAdClient.java#L65 and https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/AzureAdProfile.java

Notice there is also a manual test to confirm the correct behaviour: https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/test/java/org/pac4j/oidc/run/RunAzureAdClient.java

papegaaij:
Great work on this. In my PR I decided to only add the minimum needed for correct behavior. Good to see you are continuing development on this.

leleuj:
Offering an easy access to the authenticated user profile is a great strength of pac4j (at least, in the OAuth clients), so I did the same for OpenID Connect

PR #434, opened 2016-01-07
Clean deprecated elements

leleuj:


mmoayyed:
:+1: 

I assume we are also not using any deprecated APIs from libs?

leleuj:
Yep

PR #438, opened 2016-01-11
Disable processing of SAML external entities

### LINKED ISSUE #369: lack of XML External Entity Injection detection
in pac4j/pac4j-saml/src/main/java/org/pac4j/saml/transport/Pac4jHTTPPostDecoder.java, the method MessageDecodingException() does not detect potential "XML External Entity Injection" so that the hacker can easily uses the server to attack other computer, please throw SAXException when the injection is detected.

mmoayyed:
Closes  #369


leleuj:
Thanks. Will backport to 1.7.x and do the releases...

leleuj:
And to 1.8.x as well (master is in version 1.9.0 now).

PR #445, opened 2016-01-14
Remove incompatible xmltooling dependency

### LINKED ISSUE #444: pac4j-saml has incompatible OpenSAML v2 dependency on xmltooling
Hello,

I recently [opened an issue against OpenSAML v3](https://issues.shibboleth.net/jira/browse/OSJ-152), where I was getting the following stacktrace:

```
Caused by: java.lang.ClassCastException: org.opensaml.xml.schema.impl.XSAnyBuilder cannot be cast to org.opensaml.core.xml.XMLObjectBuilder
	at org.opensaml.core.xml.config.XMLConfigurator.initializeObjectProviders(XMLConfigurator.java:238)
	at org.opensaml.core.xml.config.XMLConfigurator.load(XMLConfigurator.java:203)
	at org.opensaml.core.xml.config.XMLConfigurator.load(XMLConfigurator.java:188)
	at org.opensaml.core.xml.config.XMLConfigurator.load(XMLConfigurator.java:162)
	at org.opensaml.core.xml.config.AbstractXMLObjectProviderInitializer.init(AbstractXMLObjectProviderInitializer.java:52)
	at org.opensaml.core.xml.config.XMLObjectProviderInitializer.init(XMLObjectProviderInitializer.java:45)
	at org.opensaml.core.config.InitializationService.initialize(InitializationService.java:56)
	at org.pac4j.saml.util.Configuration.bootstrap(Configuration.java:76)
	at org.pac4j.saml.util.Configuration.<clinit>(Configuration.java:49)
	... 8 more
```

The error is due to having conflicting jar resources at the same path (two jars have the same resource file in the same path, such as default-config.xml), so I presumed that it was an OpenSAML issue that only showed up if the classloader ordered jars in a specific way. However, [according to the response to my issue](https://issues.shibboleth.net/jira/browse/OSJ-152?focusedCommentId=22022&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22022), xmltooling is part of OpenSAML v2, and is not intended nor tested alongside OpenSAML v3.

The [pac4j-saml pom.xml has an explicit dependency on xmltooling](https://github.com/pac4j/pac4j/blob/6b090ac1bb466ca038de9ae20a44284da636d289/pac4j-saml/pom.xml#L166-L168), so it appears that needs to be removed. Is there any specific reason that xmltooling is an explicit dependency of pac4j-saml?

Note that when I remove the dependency of pac4j-saml on xmltooling, that leads to two classes getting NoClassDefFoundError on `org/opensaml/xml/util/XMLHelper`: `org.pac4j.saml.metadata.SAML2IdentityProviderMetadataResolver` and `org.pac4j.saml.metadata.SAML2MetadataGenerator`. They both appear to be using XMLHelper#nodeToString, and it appears that the method has moved in OpenSAML v3 to `net.shibboleth.utilities.java.support.xml.SerializeSupport` (with the same signature and javadoc) contained in the `net.shibboleth.utilities:java-support:7.1.1` transitive dependency of opensaml-core. So, unless I missed something, it may be a quick fix.

jhegg:
xmltooling is a part of OpenSAML v2, and is incompatible with OpenSAML v3 (see: https://issues.shibboleth.net/jira/browse/OSJ-152). It appears that it worked fine for most users in the demo projects, because the classloader order happened to place xmltooling after the other opensaml libraries (such as opensaml-core), causing the correct default-config.xml to be loaded, instead of the xmltooling version of the same file.

The xmltooling library is only used in the pac4j-saml module for a convenience method, which is also implemented in the OpenSAML v3 java-tooling library. This change replaces the call with that newer method, and also removes the xmltooling dependency from the pom.

This change resolves #444.

### DIFF
--- a/pac4j-saml/pom.xml
+++ b/pac4j-saml/pom.xml
@@ -163,21 +163,6 @@
             <version>${spring.extensions.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.opensaml</groupId>
-            <artifactId>xmltooling</artifactId>
-            <version>1.4.4</version>
-            <exclusions>
-                <exclusion>
-                    <artifactId>joda-time</artifactId>
-                    <groupId>joda-time</groupId>
-                </exclusion>
-                <exclusion>
-                    <artifactId>xmlsec</artifactId>
-                    <groupId>org.apache.santuario</groupId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
             <groupId>javax.servlet</groupId>
             <artifactId>javax.servlet-api</artifactId>
             <scope>provided</scope>
--- a/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2IdentityProviderMetadataResolver.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2IdentityProviderMetadataResolver.java
@@ -20,6 +20,7 @@ import net.shibboleth.utilities.java.support.component.ComponentInitializationEx
 import net.shibboleth.utilities.java.support.resolver.CriteriaSet;
 import net.shibboleth.utilities.java.support.resolver.ResolverException;
 import net.shibboleth.utilities.java.support.resource.Resource;
+import net.shibboleth.utilities.java.support.xml.SerializeSupport;
 import net.shibboleth.utilities.java.support.xml.XMLParserException;
 import org.opensaml.core.criterion.EntityIdCriterion;
 import org.opensaml.core.xml.XMLObject;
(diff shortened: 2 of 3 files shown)


mmoayyed:
:+1:

leleuj:
Removing this kind of dependency is always a good thing. Thanks

leleuj:
@jhegg : can you submit the same pull request for the master? Thanks

I'll release a pac4j v1.8.5 in a few days with your change.

jhegg:
@leleuj I'll do that shortly - thanks very much for your prompt response!

jhegg:
@leleuj Question: do you want the PR to be a merge of 1.8.x into master? Or do you want a cherry-pick of the my commit in the PR to master? I'm used to the merge workflow.

leleuj:
Whatever works for you...

PR #447, opened 2016-01-15
Remove incompatible xmltooling dependency

jhegg:
xmltooling is a part of OpenSAML v2, and is incompatible with OpenSAML v3. It
appears that it worked fine for most users in the demo projects, because the
classloader order happened to place xmltooling after the other opensaml
libraries (such as opensaml-core), causing the correct default-config.xml to be
loaded, instead of the xmltooling version of the same file.

The xmltooling library is only used in the pac4j-saml module for a convenience
method, which is also implemented in the OpenSAML v3 java-tooling library. This
change replaces the call with that newer method, and also removes the
xmltooling dependency from the pom.

This change is in reference to #444.

(This is a cherry-pick of bba2d40 from the 1.8.x branch)

### DIFF
--- a/pac4j-saml/pom.xml
+++ b/pac4j-saml/pom.xml
@@ -163,21 +163,6 @@
             <version>${spring.extensions.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.opensaml</groupId>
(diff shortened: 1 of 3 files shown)


leleuj:
Many thanks.

PR #450, opened 2016-01-17
Updating the ADFS documentation after my return of use

peyo:


leleuj:
Thanks for taking time to update the documentation.

PR #453, opened 2016-01-20
Remove or reduce commons-codec dependency

jasonaowen:
## Background

While looking into integrating pac4j-saml into our product, my team found conflicts between pac4j-saml's dependencies and existing product dependencies with different versions. One of these was commons-codec, and so I looked into how widely-used it was in pac4j. We were able to exclude the transitive dependency from our product to work around this, but I thought others might appreciate reduced dependencies.

## Details
I found a few classes (mostly tests) were using the Apache commons-codec library to encode and decode Base64 text. I recently found out that Java 8 introduced a [Base64 utility class](http://docs.oracle.com/javase/8/docs/api/java/util/Base64.html).  I found replacing the commons-codec-provided Base64 encoder/decoder with the JRE-provided encoder/decoder to be straightforward and I think it's a relatively easy way to minimize dependencies.

The commons-codec library is still used by pac4j-http and pac4j-oauth for sha512 and hexadecimal encoding, so this PR does not entirely remove commons-codec as a dependency. It does remove it as a dependency for pac4j-mongo; that did not require any code changes, so either the code that depended on it was previously removed or I don't understand why it was included. If it is still required, I can update this PR to reintroduce the dependency with an explanatory comment.

Along the way, I found that one of the test classes that used Base64 encoding had some repeated code, so I refactored it in a way that I found to be more readable. Does that look useful to you? Please let me know if you'd prefer that change separated into its own PR or removed entirely.

## Testing
I ran `mvn test` from the root project and all the tests still pass.

### DIFF
--- a/pac4j-http/src/main/java/org/pac4j/http/credentials/extractor/BasicAuthExtractor.java
+++ b/pac4j-http/src/main/java/org/pac4j/http/credentials/extractor/BasicAuthExtractor.java
@@ -15,7 +15,6 @@
  */
 package org.pac4j.http.credentials.extractor;
 
-import org.apache.commons.codec.binary.Base64;
 import org.pac4j.core.context.HttpConstants;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.exception.CredentialsException;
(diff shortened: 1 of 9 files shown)


jasonaowen:
I took a project-wide approach instead of a module-specific approach; I would be happy to split this into multiple, per-module pull requests if that would be easier to review.

leleuj:
In this new version 1.9, I'm cleaning useless mechanisms and components. So this pull request is really welcome. Many thanks.

PR #467, opened 2016-02-04
Clean and upgrade dependencies

leleuj:


pac4j-cas/pom.xml
```diff
@@ -27,6 +27,10 @@
     <packaging>jar</packaging>
     <name>pac4j for CAS protocol</name>
 
+    <properties>
+        <cas.version>3.4.1</cas.version>
```
mmoayyed: You might want to bump this to 3.6 at least.

pac4j-openid/pom.xml
```diff
@@ -35,18 +40,27 @@
...
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
-            <groupId>javax.servlet</groupId>
-            <artifactId>javax.servlet-api</artifactId>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
```
mmoayyed: This should be marked as provided.

pac4j-oauth/pom.xml
```diff
@@ -27,15 +27,16 @@
 	<packaging>jar</packaging>
 	<name>pac4j for OAuth protocol</name>
 
+	<properties>
+		<scribe.version>2.2.2</scribe.version>
+		<jackson.version>2.7.1-1</jackson.version>
```
mmoayyed: You likely should stay with 2.6 on jackson. 2.7 is not supported by Spring yet, and I imagine that might break a lot of existing uses.

pac4j-cas/pom.xml
leleuj: I'm referring to the CAS client: the latest version is the 3.4.1, isn't it?

pac4j-openid/pom.xml
leleuj: If I exclude the `commons-logging` dependency for `openid4java`, the `jcl-over-slf4j` is mandatory, isn't it? So I shouldn't mark it as `provided`, should I?

pac4j-oauth/pom.xml
leleuj: Yes, you're right. I will change the dependency version.

pac4j-cas/pom.xml
mmoayyed: True. Ignore this plz. 
--
Sent with Air Mail

leleuj:
@mmoayyed : are we good on this one?

mmoayyed:
Yup. Please proceed.

PR #470, opened 2016-02-09
Handle null BearerAccessToken/BearerAccessTokenBean in OidcProfile wh…

### LINKED ISSUE #458: Null pointer exception serializing OidcProfile
We have a problem using a shared session store with Pac4j oidc. There is a null pointer exception when Serializing OidcProfile.BearerAccessTokenBean,

The issue occurs in the toBean method, 

    public static BearerAccessTokenBean toBean(final BearerAccessToken token) {
            return new BearerAccessTokenBean(token.getValue(),
                    token.getLifetime(), token.getScope().toStringList());
    }

but at this time I am not sure if token is null or token.getScope() is null. I suspect the latter. @mmoayyed I know you worked on this previously, which do you think it might be? Would it be ok to use `null` in place of the null token / null scope list?

I will try to get this work planned in to our project, but in the mean time the issue is here if anyone else needs to work on it.

baxterj:
…en doing serialization and deserialization. fixes #458

The BearerAccessToken is more often than not null due to spring security’s clearing of credentials from the authentication.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/OidcProfile.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/OidcProfile.java
@@ -73,7 +73,10 @@ public class OidcProfile extends CommonProfile implements Externalizable {
     @Override
     public void writeExternal(final ObjectOutput out) throws IOException {
         super.writeExternal(out);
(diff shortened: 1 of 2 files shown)


leleuj:
Perfect. Many thanks.

Issue #480, opened 2016-02-11
Don't include default port on https


jonathanlabin:
In https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/http/RelativeCallbackUrlResolver.java#L36 the port number will be inserted into the URL even if it is the default HTTPS port.  This logic should really be of the form:
If ((http AND port != DEFAULT_PORT) OR (https AND port != DEFAULT_HTTP_PORT))

The corresponding DEFAULT_HTTP_PORT constant would also need to be created in https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/context/HttpConstants.java

leleuj:
You're right. I just saw your PR: https://github.com/pac4j/pac4j/pull/481 which I'm going to review carefully.

leleuj:
Your PR targets the master. I think it would be great to also fix this for the 1.8.x branch (in another pull request), so that we can include this fix in the 1.8.6 release.

In any case, many thanks for your help!

jonathanlabin:
Sorry, new to this.  To also fix 1.8.x branch do I branch the 1.8.x repo and manually re-apply the same changes there or is there a more correct way starting from this bugfix branch that is targeting master?

leleuj:
Generally, you can cherrypick (git command) the appropriate commit. That said, it might be tricky and we expect a slightly different solution in both branches. So the easiest way for you is to re-apply and update changes.

jonathanlabin:
I didn't notice the need for a different solution between the branches (unless I missed something) but I still re-applied the changes manually since there are just a few.

leleuj:
I was wrong indeed, same solution for both branches. Thanks.

It will go in the 1.8.6 release.

PR #481, opened 2016-02-11
Don't include default port on https

jonathanlabin:
For Issue: #480

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/http/RelativeCallbackUrlResolver.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/http/RelativeCallbackUrlResolver.java
@@ -33,7 +33,8 @@ public class RelativeCallbackUrlResolver implements CallbackUrlResolver {
 
             sb.append(context.getScheme()).append("://").append(context.getServerName());
 
-            if (context.getServerPort() != HttpConstants.DEFAULT_PORT) {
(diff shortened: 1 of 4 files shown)


leleuj:
The build fails because a test is broken. Can you fix it?

```
Failed tests: 
  RelativeCallbackUrlResolverTests.testCompute_whenRequestIsSecure:71 expected:<https://localhost[]/cas/login> but was:<https://localhost[:80]/cas/login>
```

pac4j-core/src/main/java/org/pac4j/core/http/RelativeCallbackUrlResolver.java
```diff
@@ -33,7 +33,8 @@ public String compute(final String callbackUrl, WebContext context) {
 
             sb.append(context.getScheme()).append("://").append(context.getServerName());
 
-            if (context.getServerPort() != HttpConstants.DEFAULT_PORT) {
+            if(("http".contentEquals(context.getScheme()) && context.getServerPort() != HttpConstants.DEFAULT_PORT) ||
```
leleuj: I would rewrite this test with an `equalsIgnoreCase` (instead of `contentEquals`) or maybe even better (for the master branch), add two new methods to the `ContextHelper`: `isHttp(WebContext context)` and `isHttps(WebContext context)`.

PR #486, opened 2016-02-15
Issue: 480 applied to branch 1.8.x: 

jonathanlabin:
#480 Only include the port in a URL when the port in use is not the default for the selected scheme.


leleuj:
Excellent! Thanks

PR #487, opened 2016-02-16
Keep roles / permissions using JWT

leleuj:


pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -138,9 +140,19 @@ private static void createJwtProfile(final TokenCredentials credentials, final S
...
         }
 
-        Map<String, Object> claims = new HashMap<>(claimSet.getClaims());
-        claims.remove(JwtConstants.SUBJECT);
-        final UserProfile profile = ProfileHelper.buildProfile(subject, claims);
+        final Map<String, Object> attributes = new HashMap<>(claimSet.getClaims());
+        attributes.remove(JwtConstants.SUBJECT);
+        final List<String> roles = (List<String>) attributes.get(JwtGenerator.INTERNAL_ROLES);
+        attributes.remove(JwtGenerator.INTERNAL_ROLES);
+        final List<String> permissions = (List<String>) attributes.get(JwtGenerator.INTERNAL_PERMISSIONS);
```
mmoayyed: Why remove them? What's the harm in keeping them both as roles/permissions and as attributes?

leleuj: No harm practically. It's more a choice from a theoretical point of view and to stay consistent with everything else where you get the roles / permissions via some methods and attributes from another one. They don't have the same meaning: roles / permissions are meant for authorization checks while attributes are not.

mmoayyed: Copy that. :+1:

PR #493, opened 2016-02-19
HTTP Digest Direct Authenticator

mirceac:
-implement RFC 2617 for digest authentication
Create DigestCredentials given Digest Authorization header
This has to be used similar as the Direct Basic Authenticator is used
Create new DigestAuthenticator intervace to validate DigestCredentials retrieved from the authenticator

The realm value is configurable:

 DirectDigestAuthClient digestAuthClient = new DirectDigestAuthClient(new SecurityDigestAuthenticator());
 digestAuthClient.setRealm("myRealm");

In order to validate that Digest authentication is successful is needed that:

digestCredentials.getResponse().equals(digestCredentials.calculateServerDigest(true, pwd))

where pwd is the user password kept in database.
According to RFC 2617 the user may keep HA1 encoded format
of the password in database: HA1=MD5(username-value + ":" + realm-value + ":" + passwd)
If the user keeps password in clear (not recommended) then the digest validation should be performed like this:

digestCredentials.getResponse().equals(digestCredentials.calculateServerDigest(false, pwd))

digestCredentials.getResponse() represents the client response digest
digestCredentials.calculcateServerDigest represents the server digest. These two must be equal for a successful digest authentication

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
@@ -90,6 +90,22 @@ public class RequiresHttpAction extends Exception {
         context.setResponseStatus(HttpConstants.UNAUTHORIZED);
         return new RequiresHttpAction(message, HttpConstants.UNAUTHORIZED);
     }
+
+    /**
+     * Build a digest auth popup credentials.
+     *
+     * @param message message
+     * @param context context
+     * @param realmName realm name
+     * @return a digest auth popup credentials
+     */
+    public static RequiresHttpAction unauthorizedDigest(final String message, final WebContext context, final String realmName, final String qop, final String nonce) {
+        if (CommonHelper.isNotBlank(realmName)) {
+            context.setResponseHeader(HttpConstants.AUTHENTICATE_HEADER, "Digest realm=\"" + realmName + "\", qop=\"" + qop + "\", nonce=\"" + nonce + "\"");
+        }
+        context.setResponseStatus(HttpConstants.UNAUTHORIZED);
+        return new RequiresHttpAction(message, HttpConstants.UNAUTHORIZED);
+    }
     
     /**
      * Build a forbidden response.
--- a/pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
+++ b/pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
@@ -1,0 +1,77 @@
+package org.pac4j.http.client.direct;
+
+import org.pac4j.core.context.WebContext;
+import org.pac4j.core.exception.CredentialsException;
+import org.pac4j.core.exception.RequiresHttpAction;
+import org.pac4j.http.credentials.CredentialUtil;
+import org.pac4j.http.credentials.DigestCredentials;
+import org.pac4j.http.credentials.authenticator.DigestAuthenticator;
+import org.pac4j.http.credentials.authenticator.UsernamePasswordAuthenticator;
+import org.pac4j.http.credentials.extractor.BasicAuthExtractor;
+import org.pac4j.http.credentials.extractor.DigestAuthExtractor;
+import org.pac4j.http.profile.creator.ProfileCreator;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Random;
+
+/**
+ * <p>This class is the client to authenticate users directly through HTTP digest auth.</p>
+ * <p>It returns a {@link org.pac4j.http.profile.HttpProfile}.</p>
+ *
+ * @see org.pac4j.http.profile.HttpProfile
+ * @author Mircea Carasel
+ */
+public class DirectDigestAuthClient extends DirectHttpClient<DigestCredentials> {
+    private String realm = "pac4jRealm";
+
+    public DirectDigestAuthClient() {
+    }
+
+    public DirectDigestAuthClient(final DigestAuthenticator digestAuthenticator) {
+        setAuthenticator(digestAuthenticator);
+    }
+
+    public DirectDigestAuthClient(final DigestAuthenticator digestAuthenticator,
+                                 final ProfileCreator profileCreator) {
+        setAuthenticator(digestAuthenticator);
+        setProfileCreator(profileCreator);
+    }
+
+    @Override
+    public DigestCredentials getCredentials(WebContext context) throws RequiresHttpAction {
+        DigestCredentials credentials = super.getCredentials(context);
+        if (credentials == null) {
+            String nonce = calculateNonce();
+            /* Per RFC 2617
+            If a server receives a request for an access-protected object, and an
+            acceptable Authorization header is not sent, the server responds with
+            a "401 Unauthorized" status code, and a WWW-Authenticate header
+             */
+            RequiresHttpAction.unauthorizedDigest("Digest required", context, realm, "auth", nonce);
+        }
+        return credentials;
+    }
+
+    @Override
+    protected void internalInit(final WebContext context) {
+        extractor = new DigestAuthExtractor(getName());
+        super.internalInit(context);
+    }
+
+    public void setRealm(String realm) {
+        this.realm = realm;
+    }
+
+    /**
+     * A server-specified data string which should be uniquely generated each time a 401 response is made (RFC 2617)
+     */
+    private String calculateNonce() {
+        Date d = new Date();
+        SimpleDateFormat f = new SimpleDateFormat("yyyy:MM:dd:hh:mm:ss");
+        String fmtDate = f.format(d);
+        Random rand = new Random(100000);
+        Integer randomInt = rand.nextInt();
+        return CredentialUtil.H(fmtDate + randomInt.toString());
+    }
+}
--- a/pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
+++ b/pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
@@ -1,13 +1,11 @@
 package org.pac4j.http.client.direct;
 
 import org.pac4j.core.context.WebContext;
-import org.pac4j.core.exception.CredentialsException;
 import org.pac4j.core.exception.RequiresHttpAction;
+import org.pac4j.core.util.CommonHelper;
 import org.pac4j.http.credentials.CredentialUtil;
 import org.pac4j.http.credentials.DigestCredentials;
 import org.pac4j.http.credentials.authenticator.DigestAuthenticator;
-import org.pac4j.http.credentials.authenticator.UsernamePasswordAuthenticator;
-import org.pac4j.http.credentials.extractor.BasicAuthExtractor;
 import org.pac4j.http.credentials.extractor.DigestAuthExtractor;
 import org.pac4j.http.profile.creator.ProfileCreator;
 
@@ -70,8 +68,6 @@ public class DirectDigestAuthClient extends DirectHttpClient<DigestCredentials> 
         Date d = new Date();
         SimpleDateFormat f = new SimpleDateFormat("yyyy:MM:dd:hh:mm:ss");
         String fmtDate = f.format(d);
-        Random rand = new Random(100000);
-        Integer randomInt = rand.nextInt();
-        return CredentialUtil.H(fmtDate + randomInt.toString());
+        return CredentialUtil.H(fmtDate + CommonHelper.randomString(10));
     }
 }
--- a/pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
+++ b/pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
@@ -9,9 +9,8 @@ import org.pac4j.http.credentials.authenticator.DigestAuthenticator;
 import org.pac4j.http.credentials.extractor.DigestAuthExtractor;
 import org.pac4j.http.profile.creator.ProfileCreator;
 
-import java.text.SimpleDateFormat;
-import java.util.Date;
-import java.util.Random;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 
 /**
  * <p>This class is the client to authenticate users directly through HTTP digest auth.</p>
@@ -36,16 +35,16 @@ public class DirectDigestAuthClient extends DirectHttpClient<DigestCredentials> 
         setProfileCreator(profileCreator);
     }
 
+    /** Per RFC 2617
+     * If a server receives a request for an access-protected object, and an
+     * acceptable Authorization header is not sent, the server responds with
+     * a "401 Unauthorized" status code, and a WWW-Authenticate header
+     */
     @Override
-    public DigestCredentials getCredentials(WebContext context) throws RequiresHttpAction {
+    public DigestCredentials getCredentials(final WebContext context) throws RequiresHttpAction {
         DigestCredentials credentials = super.getCredentials(context);
         if (credentials == null) {
             String nonce = calculateNonce();
-            /* Per RFC 2617
-            If a server receives a request for an access-protected object, and an
-            acceptable Authorization header is not sent, the server responds with
-            a "401 Unauthorized" status code, and a WWW-Authenticate header
-             */
             RequiresHttpAction.unauthorizedDigest("Digest required", context, realm, "auth", nonce);
         }
         return credentials;
@@ -57,17 +56,18 @@ public class DirectDigestAuthClient extends DirectHttpClient<DigestCredentials> 
         super.internalInit(context);
     }
 
-    public void setRealm(String realm) {
+    public void setRealm(final String realm) {
         this.realm = realm;
     }
 
     /**
      * A server-specified data string which should be uniquely generated each time a 401 response is made (RFC 2617)
+     * Based on current time including nanoseconds
      */
     private String calculateNonce() {
-        Date d = new Date();
-        SimpleDateFormat f = new SimpleDateFormat("yyyy:MM:dd:hh:mm:ss");
-        String fmtDate = f.format(d);
-        return CredentialUtil.H(fmtDate + CommonHelper.randomString(10));
+        LocalDateTime time = LocalDateTime.now();
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy:MM:dd:HH:mm:ss.SSS");
+        String fmtTime = formatter.format(time);
+        return CredentialUtil.encryptMD5(fmtTime + CommonHelper.randomString(10));
     }
 }
--- a/pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
+++ b/pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
@@ -20,12 +20,11 @@ public class CredentialUtil {
      * @param data data
      * @return MD5(data)
      */
-    public static String H(String data) {
+    public static String encryptMD5(String data) {
         try {
             MessageDigest digest = MessageDigest.getInstance("MD5");
             return copyValueOf(Hex.encodeHex(digest.digest(data.getBytes())));
         } catch (NoSuchAlgorithmException ex) {
-            // shouldn't happen
             throw new RuntimeException("Failed to instantiate an MD5 algorithm", ex);
         }
     }
@@ -37,7 +36,7 @@ public class CredentialUtil {
      * @param secret secret
      * @return H(concat(secret, ":", data));
      */
-    public static String KD(String secret, String data) {
-        return H(secret + ":" + data);
+    public static String encryptMD5(String secret, String data) {
+        return encryptMD5(secret + ":" + data);
     }
 }
--- a/pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
+++ b/pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
@@ -1,5 +1,8 @@
 package org.pac4j.http.credentials;
 
+import static java.lang.String.copyValueOf;
+
+import org.apache.commons.codec.binary.Hex;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
@@ -11,10 +14,6 @@ import java.security.NoSuchAlgorithmException;
  */
 public class CredentialUtil {
 
-    private static final char[] toHex = {
-            '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'
-    };
-
     /**
      * Defined in rfc 2617 as H(data) = MD5(data);
      *
@@ -24,28 +23,11 @@ public class CredentialUtil {
     public static String H(String data) {
         try {
             MessageDigest digest = MessageDigest.getInstance("MD5");
-
-            return toHexString(digest.digest(data.getBytes()));
+            return copyValueOf(Hex.encodeHex(digest.digest(data.getBytes())));
         } catch (NoSuchAlgorithmException ex) {
             // shouldn't happen
             throw new RuntimeException("Failed to instantiate an MD5 algorithm", ex);
         }
-    }
-
-    /**
-     * Converts b[] to hex string.
-     *
-     * @param b the bte array to convert
-     * @return a Hex representation of b.
-     */
-    public static String toHexString(byte b[]) {
-        int pos = 0;
-        char[] c = new char[b.length * 2];
-        for (int i = 0; i < b.length; i++) {
-            c[pos++] = toHex[(b[i] >> 4) & 0x0F];
-            c[pos++] = toHex[b[i] & 0x0f];
-        }
-        return new String(c);
     }
 
     /**
--- a/pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
+++ b/pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
@@ -1,0 +1,61 @@
+package org.pac4j.http.credentials;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * This class contains utility methods related to credential encryption algorithms
+ * (MD5 for http digest)
+ *
+ * @author Mircea Carasel
+ */
+public class CredentialUtil {
+
+    private static final char[] toHex = {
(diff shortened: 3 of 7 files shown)


mirceac:
Please have a look and merge with original pac4j project if you find it useful

I haven't write any tests, but I plan to write tests also and make new pull request. However I tested the new Direct Digest Authenticator with my web app that uses pac4j (play-pac4j) and works fine

Thanks!
Mircea Carasel

pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
```diff
@@ -90,6 +90,22 @@ public static RequiresHttpAction unauthorized(final String message, final WebCon
...
+    /**
+     * Build a digest auth popup credentials.
+     *
+     * @param message message
+     * @param context context
+     * @param realmName realm name
+     * @return a digest auth popup credentials
+     */
+    public static RequiresHttpAction unauthorizedDigest(final String message, final WebContext context, final String realmName, final String qop, final String nonce) {
+        if (CommonHelper.isNotBlank(realmName)) {
```
mmoayyed: Doesn't look like this actually belongs here. You should likely move this to where these values make sense as an extension.

pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
```diff
@@ -0,0 +1,61 @@
...
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * This class contains utility methods related to credential encryption algorithms
+ * (MD5 for http digest)
+ *
+ * @author Mircea Carasel
+ */
+public class CredentialUtil {
```
mmoayyed: Isn't there a library that would handle all of this? Why are we manually handling encryption?

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
```diff
@@ -0,0 +1,136 @@
...
+    @Override
+    public int hashCode() {
+        int result = username != null ? username.hashCode() : 0;
+        result = 31 * result + (response != null ? response.hashCode() : 0);
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return CommonHelper.toString(this.getClass(), "username", this.username, "response", "[PROTECTED]",
```
mmoayyed: What is [PROTECTED]?

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
```diff
@@ -0,0 +1,136 @@
...
+                "clientName", getClientName());
+    }
+
+    private Map<String, String> parseTokenValue(String token) {
```
mmoayyed: This looks like duplicate code to me, based on what the token authenticator does, right?

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
```diff
@@ -0,0 +1,136 @@
...
+            ha1 = CredentialUtil.H(username + ":" + realm + ":" +password);
+        }
+
+        String digest;
+
+        if (qop == null) {
+            // as per RFC 2069 compliant clients (also reaffirmed by RFC 2617)
+            digest = CredentialUtil.KD(ha1, nonce + ":" + ha2);
+        }
+        else if ("auth".equals(qop)) {
```
mmoayyed: You should likely move else clauses to be on the same like as the preceding }

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * A server-specified data string which should be uniquely generated each time a 401 response is made (RFC 2617)
+     */
+    private String calculateNonce() {
+        Date d = new Date();
+        SimpleDateFormat f = new SimpleDateFormat("yyyy:MM:dd:hh:mm:ss");
+        String fmtDate = f.format(d);
+        Random rand = new Random(100000);
```
mmoayyed: Shouldnt this be SecureRandom?

mmoayyed: 100000 is the seed right? Why is it static?

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
```diff
@@ -0,0 +1,77 @@
...
+        Integer randomInt = rand.nextInt();
+        return CredentialUtil.H(fmtDate + randomInt.toString());
```
mmoayyed: Would be make sense to simply generate UUID instead?

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/DigestAuthenticator.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.pac4j.http.credentials.DigestCredentials;
+import org.pac4j.http.credentials.UsernamePasswordCredentials;
+
+/**
+ * This interface represents the contract to validate digest credentials.
+ *
+ * @author Mircea Carasel
+ */
+public interface DigestAuthenticator extends Authenticator<DigestCredentials> {
```
mmoayyed: I don't see an impl for this?

mmoayyed: In your example, you are using SecurityDigestAuthenticator(). Where is that? and how does that work?

pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
leleuj: Yes, it looks specific to the digest auth, but at the same time, it fits well here.

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
leleuj: To protect the response value I guess. Consistent with: https://github.com/pac4j/pac4j/blob/master/pac4j-http/src/main/java/org/pac4j/http/credentials/UsernamePasswordCredentials.java#L68

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/DigestAuthenticator.java
mirceac: That is an implementation of the DigestAuthenticator interface that should be done in the user application. Example:
public class SecurityDigestAuthenticator implements DigestAuthenticator {

    @Override
    public void validate(DigestCredentials credentials) {
        if (credentials == null) {
            this.throwsException("No credential");
        }
        String username = credentials.getUsername();
        if (CommonHelper.isBlank(username)) {
            this.throwsException("Username cannot be blank");
        }

        HttpProfile profile = new HttpProfile();
        profile.setId(username);
        credentials.setUserProfile(profile);
    }
    protected void throwsException(String message) {
        throw new CredentialsException(message);
    }

I followed the same model as for the Basic direct authenticator.  The purpose is to inject the http user profile here

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
leleuj: Something I created and used in OAuth: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java#L230

mirceac: based on RFC 2617 it is said:

nonce
     A server-specified data string which should be uniquely generated
     each time a 401 response is made. It is recommended that this
     string be base64 or hexadecimal data. Specifically, since the
     string is passed in the header lines as a quoted string, the
     double-quote character is not allowed.

     The contents of the nonce are implementation dependent. The quality
     of the implementation depends on a good choice. A nonce might, for
     example, be constructed as the base 64 encoding of

         time-stamp H(time-stamp ":" ETag ":" private-key)

So rfc recommends that nonce to be calculated given timestamp. 
Related to static Random(100000) you are correct, we better use SecureRandom here

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
mirceac: SecureRandom is better, correct. 1000000 is put just to follow RFC 2617 recommendation:

time-stamp H(time-stamp ":" ETag ":" private-key)

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
mirceac: will do the change as recommended

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/DigestAuthenticator.java
leleuj: So we should be able to use the `SimpleTestTokenAuthenticator` for example (https://github.com/pac4j/pac4j/blob/master/pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/test/SimpleTestTokenAuthenticator.java), right?

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
mirceac: right.. I will fix this

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
mirceac: that is the client response, not to write the client response in clear. I will remove, that is duplicated code as outlined on a different comment -thanks

pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
mirceac: I am using: java.security.MessageDigest
for other hexadecimal conversion, probably there is... I will search more

pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
mirceac: unauthorized related to basic authentication is here, so I thought a good place to put the digest related here too

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
mirceac: Yes, this is what I thought, I followed the model for UsernamePasswordCredentials. Let me know if  we should consider this duplicate or to keep the equals and hash functions. thanks

pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
leleuj: If you find some library, it would be an `optional` dependency (only required for Digest).

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
leleuj: I have no problem with "[PROTECTED]", but other information would be more interesting in the `toString`, like the `username`.

leleuj: About the `equals` and `hashCode` value, I would rely on the parent ones.

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
leleuj: I'm not sure to follow: the `TokenAuthenticator` does not do any parsing...

mirceac: Basically what follows after "Digest " in the Authorization HTTP header is a token composed of different values like: nonce, nc, cnonce, qop, response...this is why I thought that DigestCredentials should extend TokenCredentials

Further, the DigestAuthExtractor retrieves the generic TokenCredentials from the Authorization header, just like in BasicAuthExtractor. then it converts it to DigestCredentials (token parsing incapsulated in DigetCredentials). 

I can change this so DigestCredential to extend HttpCredentials and enhance token parsing inside DigestAuthExtractor. What do you think ? thanks!

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
mirceac: correct, I better use CommonHelper instead of static Random(1000000) - thanks for the tip!

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/DigestAuthenticator.java
mirceac: yes, correct

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
leleuj: No, I think it makes sense to inherit from `TokenCredentials`. I'm even wondering if `DigestAuthExtractor ` could inherit from `BasicAuthExtractor`...

mirceac: We could inherit from BasicAuthExtractor, not much code to save, just we would get read of duplicating:
    public DigestAuthExtractor(final String headerName, final String prefixHeader, final String clientName) {
        this.extractor = new HeaderExtractor(headerName, prefixHeader, clientName);
        this.clientName = clientName;
    }

but we wound anyway need to overwrite the extract method.

pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
mirceac: There is: org.apache.commons.codec.binary.Hex;... and there is already a dependency. I will use it instead my manual conversion

pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
mmoayyed: OK. Maybe something to refactor later.

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
```diff
@@ -0,0 +1,73 @@
...
+        setAuthenticator(digestAuthenticator);
+        setProfileCreator(profileCreator);
+    }
+
+    @Override
+    public DigestCredentials getCredentials(WebContext context) throws RequiresHttpAction {
+        DigestCredentials credentials = super.getCredentials(context);
+        if (credentials == null) {
+            String nonce = calculateNonce();
+            /* Per RFC 2617
```
mmoayyed: Could we make this an actual javadoc?

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
```diff
@@ -0,0 +1,73 @@
...
+import org.pac4j.core.util.CommonHelper;
+import org.pac4j.http.credentials.CredentialUtil;
+import org.pac4j.http.credentials.DigestCredentials;
+import org.pac4j.http.credentials.authenticator.DigestAuthenticator;
+import org.pac4j.http.credentials.extractor.DigestAuthExtractor;
+import org.pac4j.http.profile.creator.ProfileCreator;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Random;
```
mmoayyed: This import used anywhere?

I imagine you have many other unnecessary imports. Make sure you clean up everything that is not used.

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
```diff
@@ -0,0 +1,73 @@
...
+        return credentials;
+    }
+
+    @Override
+    protected void internalInit(final WebContext context) {
+        extractor = new DigestAuthExtractor(getName());
+        super.internalInit(context);
+    }
+
+    public void setRealm(String realm) {
```
mmoayyed: Not a big deal for the time being, but you have certain parameters marked as final and some are not. Please ensure you are consistent.

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
```diff
@@ -0,0 +1,73 @@
...
+        this.realm = realm;
+    }
+
+    /**
+     * A server-specified data string which should be uniquely generated each time a 401 response is made (RFC 2617)
+     */
+    private String calculateNonce() {
+        Date d = new Date();
```
mmoayyed: Pac4j is on Java8 correct? Should u really be using Date?

pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Defined in rfc 2617 as H(data) = MD5(data);
+     *
+     * @param data data
+     * @return MD5(data)
+     */
+    public static String H(String data) {
```
mmoayyed: Is there a reason this method is called H? I understand that's the spec's name, but surely, we could pick a name that is more sensible and leave the javadocs to explain the spec ref?

pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
```diff
@@ -0,0 +1,43 @@
...
+        try {
+            MessageDigest digest = MessageDigest.getInstance("MD5");
+            return copyValueOf(Hex.encodeHex(digest.digest(data.getBytes())));
+        } catch (NoSuchAlgorithmException ex) {
+            // shouldn't happen
```
mmoayyed: Remove this line please.

pac4j-http/src/main/java/org/pac4j/http/credentials/DigestCredentials.java
```diff
@@ -0,0 +1,92 @@
...
+        if (passwordAlreadyEncoded) {
+            ha1 = password;
+        } else {
+            ha1 = CredentialUtil.H(username + ":" + realm + ":" +password);
+        }
+
+        String digest;
+
+        if (qop == null) {
+            // as per RFC 2069 compliant clients (also reaffirmed by RFC 2617)
```
mmoayyed: Turn these into actual javadoc comments please. Avoid inline comments.

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/DigestAuthenticator.java
mmoayyed: Is there a reason then to create a separate specific authenticator interface? Can the credentials not be casted down?

pac4j-http/src/main/java/org/pac4j/http/credentials/extractor/DigestAuthExtractor.java
```diff
@@ -0,0 +1,80 @@
...
+            return null;
+        }
+
+        String token = credentials.getToken();
+        Map<String, String> valueMap = parseTokenValue(token);
+        String username = valueMap.get("username");
+        String response = valueMap.get("response");
+
+        if (CommonHelper.isBlank(username) || CommonHelper.isBlank(response)) {
+            throw new CredentialsException("Bad format of the digest auth header");
```
mmoayyed: You should probably explain why this is a bad format.

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
leleuj: Yes, we are on Java 8. I think it's better to move to the new Time API, although we have some `Date` left, here and there.

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/DigestAuthenticator.java
leleuj: Good question: I'm not sure there is a use case where you want to use `DigestCredentials` really. They don't play any specific role for username / password validation, do they?

mirceac: That is correct, they do not play any specific role for username/password validation. I think I can get read of DigestAuthenticator interface completely.

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
mirceac: yes

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
mirceac: sure, thanks

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
```diff
@@ -0,0 +1,73 @@
...
+        SimpleDateFormat f = new SimpleDateFormat("yyyy:MM:dd:hh:mm:ss");
```
mirceac: will do

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
mirceac: sure - thanks

pac4j-http/src/main/java/org/pac4j/http/client/direct/DirectDigestAuthClient.java
mirceac: I will move to Time API

pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
mirceac: sure, I will think to a better name and add javadoc

pac4j-http/src/main/java/org/pac4j/http/credentials/CredentialUtil.java
mirceac: sure

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/DigestAuthenticator.java
mirceac: Not sure I can do this because DirectDigestAuthClient extends DirectHttpClient, which has a validate call, after digest Authorization token is parsed and data extracted:

    @Override
    public C getCredentials(WebContext context) throws RequiresHttpAction {
        init(context);
        try {
            C credentials = this.extractor.extract(context);
            if (credentials == null) {
                return null;
            }
            getAuthenticator().validate(credentials);
            return credentials;
        } catch (CredentialsException e) {
            logger.error("Failed to retrieve or validate credentials", e);
        }
        return null;
    }

Here is a indirect cast to DigestCredentials (the generic class C becomes DigestCredentials in DirectDigestAuthClient

mirceac: Also... my understanding is that the validate call has the purpose to create HttpProfile and set it in Credentials...

leleuj: The `validate` call validates credentials and should throw a `CredentialsException` if they don't match. The creation of the profile (and attachement to the credentials) is optional but recommended as the default `ProfileCreator` retrieves the profile attached to the credentials. The `HttpProfile` is some default profile, a specific profile must be created by the `Authenticator` if the need arises...

leleuj:
@mmoayyed I think we're good on this one. Is it ok for you?

I'd like to merge it before (https://github.com/pac4j/pac4j/pull/497) or it will be a nightmare for @mirceac . I'll make a few adjustements afterwards.

mmoayyed:
I'll do a final review later this evening. My main viewpoint in reviewing this would bes to ensure we can use this functionality easily in CAS via https://github.com/Jasig/cas/issues/1384

pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
mmoayyed: One final reminder that I do think we should move this out in the future. Both for basic and digest authn functionality.

mmoayyed:
Yeah this looks fine. Still a couple of questions on whether we need that extra authenticator, but it looks good to me. We can always fix things later once we get to an impl.

pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
leleuj: Conversely, I think we should keep these methods here, `RequiresHttpAction` is the component to use to trigger new HTTP action and break the flow. Moreover, moving them out would be a breaking change. That said, it's not such a big deal and we can revisit this choice later on.

leleuj:
I will merge, but remove the interface. We'll add it if we need it later on.

Many thanks @mirceac for your patience and tenacity on this!

mirceac:
Thank you so much for merging my changes!
Regarding to removing DigestAuthenticator interface. Let me know if you want me now to remove it and maybe using the existing TokenAuthenticator interface instead. I can give it a try today, just let me know

leleuj:
No problem. I'm currently refactoring your wok with my PR: https://github.com/pac4j/pac4j/pull/497

After its merge, you'll be able to see the changes I made and submit some tests for quality.

mirceac:
ok, great! Thanks again!

leleuj:
It's updated, you can sync with master and work on tests if you want.

PR #497, opened 2016-02-22
Refactor to internally use Authenticator / ProfileCreator

leleuj:


leleuj:
@millross @mmoayyed 

This giant pull request prepares the support of reactive clients. There are many changes which make things hard to review. Though, here are the main pointers:

1) Specific concepts embedded in the pac4j-http module have been moved to the core: `Authenticator`, `Extractor`, `ProfileCreator` and the new `Redirector` to build the new `DirectClient2` (https://github.com/pac4j/pac4j/pull/497/files#diff-8c047b5ddb3861190fb8261a69bf9e49R38) and  `IndirectClient2` (https://github.com/pac4j/pac4j/pull/497/files#diff-014d2903083def7b2236f4918076e842R40)
The idea is to replace heavy inheritance by inheritance + delegation and offer smallest components that could be reused to build asynchronous clients and components.

2) With small components, overriding behaviours will become easier and allow almost possible use cases as you can define your own `Authenticator` and `ProfileCreator` and even `Redirector` and `Extractor`.

3) Don't be surprised to see empty constructors in most public components: it is so to allow their creation in Shiro (shiro.ini file), which makes initialization occurred in the `internalInit` method in most cases.

@millross : `DirectClient2` and `IndirectClient2`  should be your starting point for reactive clients: I'm looking forward to your feedbacks on this with the main question: does it help?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
```diff
@@ -0,0 +1,148 @@
...
+        return CommonHelper.toString(this.getClass(), "name", getName(), "redirector", this.redirector, "extractor", this.extractor,
+                "authenticator", this.authenticator, "profileCreator", this.profileCreator);
+    }
+
+    public Redirector getRedirector() {
+        return redirector;
+    }
+
+    public void setRedirector(final Redirector redirector) {
+        if (this.redirector == null) {
```
mmoayyed: So once you set it, you can never reset it? Should this be turned into a constructor parameter?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
```diff
@@ -0,0 +1,148 @@
...
+        CommonHelper.assertNotNull("authenticator", this.authenticator);
+        CommonHelper.assertNotNull("profileCreator", this.profileCreator);
+        if (authenticator instanceof InitializableWebObject) {
+            ((InitializableWebObject) this.authenticator).init(context);
+        }
+    }
+
+    @Override
+    protected RedirectAction retrieveRedirectAction(final WebContext context) {
+        return redirector.redirect(context);
```
mmoayyed: This is confusing a bit. A redirector does NOT redirect apparently. It builds a RedirectAction, and that thing then does redirect. I think you should opt for better names here. A redirector sounds like something that actually does the redirecting but that's now what's happening here right?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
```diff
@@ -0,0 +1,148 @@
...
+import org.pac4j.core.util.InitializableWebObject;
+
+/**
+ * New indirect client type using the {@link Redirector}, {@link Extractor}, {@link Authenticator}
+ * and {@link ProfileCreator} concepts.
+ * 
+ * @author Jerome Leleu
+ * @since 1.9.0
+ */
+public abstract class IndirectClient2<C extends Credentials, U extends CommonProfile> extends IndirectClient<C, U> {
```
mmoayyed: This is also confusing. What does 2 mean? If you already have an IndirectClient, why don't we just use that rather than using a "2" suffix?

mmoayyed: If you wanted to keep BC, I suggest you should then opt for a 2.0 release and keep this change for that. You will eventually have to get rid of that 2 I think, and it's best to do it early without introducing that type of noise.

mmoayyed: But that wont matter, since sem ver indicates that you can change Impl API across minor versions. So I'd rather we get rid of the old client, and replace it with this "2" one :)

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
leleuj: Yes, that's the idea. A default one is generally created in the `internalInit` method; if you have set your specific one, the default one is not set.
I always prefer a setter as Shiro does not support constructors with parameters.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
leleuj: Would you prefer a `RedirectActionBuilder` or `RedirectBuilder`?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
leleuj: We will have two implementations which coexist and remain for a long time (wa can rewrite everything in one version). Most clients will stay with the current version 1.0, but we will migrate clients as opportunities arise (maybe driven by what @millross will implement for Vertx).
As for the name, "2" is the best I can think of. It's better than "NewModel", isn't it?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
mmoayyed: I like the 1st better, since RedirectAction is really what you are building.

leleuj: OK. I will change that.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
mmoayyed: Why not merge everything into one client and then move all subclasses to use the new client? Bigger effort sure, but I see no reason to maintain two separate hierarchies. It does take a while to convert, but we should do so before the final release. Devs would have a difficult time understanding the difference between IndirectClient and IndirectClient2. I would in fact pick 2, since my gut feeling by reading that name says: it's newer, it's better :)

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
mmoayyed: So if I have set a specific one, I cant reset it? Why? Trying to understand the reasoning behind this.

I guess my point is, it may not make sense to set and rest redirectors, but leave that decision to the developer.

leleuj:
Renamed `Redirector` as `RedirectActionBuilder` and `Extractor` as `CredentialsExtractor`.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
leleuj: I see your point, but it's a huge work: why should we put ourselves under pressure as things are already working pretty well? In version 1.8, we already had two main types, but things could be improved. Now we have two types coexisting in harmony. I don't think we need to hurry in the migration. Plus we really want feedbacks from Jez to see how it fits for Vertx.

This leads me to a more global comment for our release strategy. People do not upgrade pac4j version as easily as we could wish. So the idea is to release as soon as possible the version 1.9.0 with the breaking changes and then add what's missing in the same stream...

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
leleuj: Why you would reset your specific one? If you have a specific one, you set
it and it will be used instead of the default one. If you don't have a
specific one, the default one will be used.



2016-02-22 19:10 GMT+01:00 Misagh Moayyed <notifications@github.com>:

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
mmoayyed: That's fair, but I would also argue that if you are going to release breaking changes, you might as well do it "correctly" once. There is no pressure for the release, right? There is an expectation of a release timeline, but nobody can expect to instantaneously receive a fix via a release. There are snapshots, and there are patch releases. 

I envision, you would basically have to do this:

- Have IndirectClient and IndirectClient2 
- Slow start migrating everything to IndirectClient2 (many releases possibly in between)
- Deprecated IndirectClient
- Remove IndirectClient

Consider that your every release (and patch release) would potentially present breaking changes...or you would have to take lots of maneuvers to not breaking BC, which is arguably a lot more work! 

Alternatively:

You do it once, and you do it correctly, and sure it takes longer, but you could always have patch releases for minor small fixes that people might need sooner. As a developer, I'd rather deal with breaking changes once, as global as they are, than deal with minor breaking changes on every upgrade, or deal with what seems to be a temporary API. 

But, all and all, if you think this will work in the long run, that's fine. I am always for incremental progress.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
mmoayyed: Thought about this some more. I don't have a particular use case in mind yet, but I guess I find a one-time setter confusing most of all. A setter is always able to set. If you want to set something only once, then ctor is the way to do. But if you want to have the setter available, then it better set the thing it wants to set, every time. 

An alternative would be for the setter to throw an exception, when it can't. I like that better at least, so the developer knows what's going on. Otherwise, you're presenting the impression that the setter can be called as many times as needed, for whatever reason.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
leleuj: In fact, it really depends on what we call "backward compatibility" and "breaking changes". For me, it's nothing more than the `Client` interface, the `ProfileManager` and `UserProfile` elements. That's why I plan to do all changes related to these classes before the 1.9.0 release.

Then, if we move a client from version 1.0 to version 2.0, some customizations may be broken. I think it's perfectly acceptable as it will be marginal case.

There is no pressure for the release indeed, but I'd like we stop adding features on 1.8.x and only work on 1.9.x.

mmoayyed: OK, understood. Let's progress. I had in mind that you might have a wider definition of breaking changes, but as long as the release policy is clear on those items, I'm perfectly fine with your approach. 

P.S: I think public APIs somehow needs to be marked somewhere. Maybe markdown, maybe even annotations in code.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
leleuj: If you throw an exception, you won't be able to override a
`CredentialsExtractor` or `RedirectActionBuilder` because as soon as the
default one will be built (in addition to yours), an exception will be
thrown.

I have another option I can propose: reset `public` setters to be what they
should be and create new `protected` specific setters like
`setCredentialsExtractorIfNotDefined` which would be used internally in
`internalInit` methods:  what do you think?

2016-02-23 16:10 GMT+01:00 Misagh Moayyed <notifications@github.com>:

mmoayyed: Could you explain that a little more please? 

So something like:

```
if this.redirector == null
    then set it
else 
    throws new YouCantSetThisAnymore()
```

is going to cause issues? 

As for `setCredentialsExtractorIfNotDefined`, I am not sure if I like that option any more than the first. This does seem like it's something we should perhaps think about in the future. So, let's keep it this way as it is, and I"ll see perhaps I can come up with a reasonable proposal.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
leleuj: I was thinking about creating a `pac4j-api` module like for CAS, but I'm not sure it brings much value.

I think there is one point left for which I answered you with a new proposal (setXIfNotDefined) and then, we'll be good for the merge, aside from any later change required by "reactive feedbacks".

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient2.java
leleuj: Let's take an example:
FormClient client = new FormClient("http://myapp/login", new
SimpleTestUsernamePasswordAuthenticator());
client.setRedirectActionBuilder(new MyRedirectActionBuilder());

When calling this:
https://github.com/pac4j/pac4j/pull/497/files#diff-06461338984333e163b57825f2b2747aR82
, it would fail as this client has already a default RedirectActionBuilder.

I will proceed with the merge and let the issue open for a few weeks. We'll
be able to revisit it with improvements before the release...

2016-02-24 9:40 GMT+01:00 Misagh Moayyed <notifications@github.com>:

leleuj:
@millross : I'm proceeding with the merge, to be able to address other issues remaining for 1.9.0. Though, integrating your feedbacks and new ones from Misagh will be done afterwards.

PR #509, opened 2016-02-28
Tests for HTTP Digest Authenticator

mirceac:
Hi,

I created some tests for the HTTP Digest Authenticator contribution.
Please have a look!

Thanks,
Mircea


leleuj:
Thanks for the tests. There is a failure during the build on a SAML test (!), but I don't see how this could be related. Will merge and see.

leleuj:
Build fixed. In fact, the `URI` constant was the "culprit": https://github.com/pac4j/pac4j/commit/7ec522cc7179c63014edfe1c0af3b2db1f006a3a

PR #519, opened 2016-03-08
Removed all formatting issues. Left in the spring converted file abstraction for now.

garpinc:
Not sure how to make the spring abstractions optional. The code as it stands requires it.. I'd need to do some refactoring and it wouldn't be clean since i'd be repeating code in places.. It does work the way it is now but if you have some ideas let me know.


leleuj:
pac4j is made for being multi-frameworks, it means it cannot have specific dependencies on Spring (or Play or Undertow...) So the main target should be to change pac4j to become open to extensions required for these dependencies.

You can try to go further by having the Spring dependency as optional. Example for Kryo we already have:
```
<dependency>
  <groupId>com.esotericsoftware</groupId>
  <artifactId>kryo-shaded</artifactId>
  <scope>compile</scope>
  <optional>true</optional>
</dependency>
```
That said, there is another caveat: classes using the Spring dependency should be new ones separated from the existing ones to prevent pac4j users from being tied to the Spring dependency. So for example, no Spring package in the imports of the `CommonHelper` class.

This is maybe not achievable, though.

garpinc:
So in that case what's the issue with copied classes?
________________________________

leleuj:
Copying classes is not clean. Imagine someone else does the same for another framework. pac4j would become a patch work of copied classes.

What will happen when you effectively use the Spring dependency? Which classes will be used? What happens when you upgrade the dependency?

pac4j being an engine, the best you can have is extensible classes you can easily use or specific classes with an optional dependency.

garpinc:
The problem for the most part is that the pac4j code affected are interfaces.. Unless I make the objects being passed around just a java Object there is going to have to some pac4j interface that abstracts the file handling. If we can’t use the spring interfaces directly because you don’t want the dependency then I think the only real way is the way I implemented it.. A mirror of the interfaces (because the interfaces are good representations of the abstraction) in the pac4j world.. I could make the implementation FileSystemResource use the actual spring implementation in an optional way. But I think we need the 2 interfaces Resource and WritableResource in the pac4j world..

 

Would that suffice?

garpinc:
Hey.. So I think really the best way going forward is to have an abstraction on these files but as you say it doesn't have to be a copy of spring files.. So basically what I did was make bare minimal interfaces and one implementation that are pac4j abstractions.. I hate continuing to merge with master.. I hope you will accept my pull request

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -104,7 +105,10 @@ protected void internalInit() {
      */
     public Client findClient(final WebContext context) {
         final String name = context.getRequestParameter(this.clientNameParameter);
-        CommonHelper.assertNotBlank("name", name);
+		if (name == null && defaultClient != null) {
```
leleuj: What is this `defaultClient` for? It doesn't seem related to the resource changes...

pac4j-core/src/main/java/org/pac4j/core/io/FileSystemResource.java
```diff
@@ -0,0 +1,40 @@
+package org.pac4j.core.io;
+
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public class FileSystemResource implements WritableResource {
```
leleuj: Some Javadoc will be appreciated.

pac4j-core/src/main/java/org/pac4j/core/io/FileSystemResource.java
```diff
@@ -0,0 +1,40 @@
...
+
+	private String path;
+
+	public FileSystemResource(String path) {
+		this.path = path;
+	}
+
+	@Override
+	public boolean exists() {
+		// TODO Auto-generated method stub
```
leleuj: I'd prefer not to keep any `TODO`. My feeling is that either you throw some exception to say it is not implemented or you implement the real logic.

pac4j-core/src/main/java/org/pac4j/core/io/InputStreamSource.java
```diff
@@ -0,0 +1,24 @@
...
+import java.io.InputStream;
+
+/**
+ * Simple interface for objects that are sources for an {@link InputStream}.
+ *
+ * <p>This is the base interface for @link Resource} interface.
+ *
+ * @see java.io.InputStream
+ * @see Resource
+ */
```
leleuj: I like to have the `@since` annotation for history.

pac4j-core/src/main/java/org/pac4j/core/io/WritableResource.java
```diff
@@ -0,0 +1,40 @@
+/*
```
leleuj: Please remove the header.

pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java
```diff
@@ -79,23 +79,36 @@ protected UsernamePasswordCredentials retrieveCredentials(final WebContext conte
             credentials = getCredentialsExtractor().extract(context);
             logger.debug("usernamePasswordCredentials: {}", credentials);
             if (credentials == null) {
-                String redirectionUrl = CommonHelper.addParameter(this.loginUrl, this.usernameParameter, username);
-                redirectionUrl = CommonHelper.addParameter(redirectionUrl, ERROR_PARAMETER, MISSING_FIELD_ERROR);
-                logger.debug("redirectionUrl: {}", redirectionUrl);
-                final String message = "Username and password cannot be blank -> return to the form with error";
-                logger.debug(message);
-                throw RequiresHttpAction.redirect(message, context, redirectionUrl);
+				// it's an AJAX request -> unauthorized (instead of a
```
leleuj: What is this change? You shouldn't mix different changes in the same pull request.

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
```diff
@@ -150,24 +150,24 @@ protected void initSAMLContextProvider(final MetadataResolver metadataManager) {
     }
 
     protected MetadataResolver initServiceProviderMetadataResolver(final WebContext context) {
-        this.spMetadataResolver = new SAML2ServiceProviderMetadataResolver(this.configuration.getServiceProviderMetadataPath(),
-                computeFinalCallbackUrl(context),
-                this.configuration.getServiceProviderEntityId(),
+		String finalCallbackUrl = computeFinalCallbackUrl(context);
+		this.spMetadataResolver = new SAML2ServiceProviderMetadataResolver(this.configuration.getServiceProviderMetadataPath(),
+				this.configuration.getServiceProviderMetadataResource(), finalCallbackUrl,
+				(this.configuration.getServiceProviderEntityId().equals("CALLBACK_URL") ? finalCallbackUrl
```
leleuj: What is this test for?

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
```diff
@@ -59,17 +69,45 @@ public SAML2ClientConfiguration() {}
...
     public SAML2ClientConfiguration(final String keystorePath, final String keystorePassword,
-                                    final String privateKeyPassword, final String identityProviderMetadataPath,
-                                    final String identityProviderEntityId, final String serviceProviderEntityId) {
+                                    final String privateKeyPassword, final String identityProviderMetadataPath) {
+        this(null, null, null, null, keystorePath, keystorePassword, privateKeyPassword, null,
+                identityProviderMetadataPath, null, null);
+    }
+
+	private SAML2ClientConfiguration(final KeyStore keyStore, final String keyStoreAlias, final String keyStoreType,
+			                         final Resource keystoreResource, final String keystorePath, final String keystorePassword,
```
leleuj: Why do we have a `keystoreResource` and a `keystorePath`?

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
```diff
@@ -674,7 +674,12 @@ protected final void validateSignature(final Signature signature, final String i
     private boolean isDateValid(final DateTime issueInstant, final int interval) {
         final DateTime before =  DateTime.now().plusSeconds(acceptedSkew);
         final DateTime after =  DateTime.now().minusSeconds(acceptedSkew + interval);
-        return issueInstant.isBefore(before) && issueInstant.isAfter(after);
+        boolean isDateValid = issueInstant.isBefore(before) && issueInstant.isAfter(after);
+		if (logger.isTraceEnabled() && !isDateValid) {
```
leleuj: We don't need a `logger.isTraceEnabled` here.

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -226,20 +242,83 @@ public static String toString(final Class<?> clazz, final Object... args) {
...
-        }
-    }
+		int prefixEnd = name.indexOf(":");
+		String prefix = null;
+		String path = name;
+		if (prefixEnd != -1) {
+			prefix = name.substring(0, prefixEnd);
+			path = name.substring(prefixEnd + 1);
+		}
+		if (prefix == null) {
```
leleuj: It's not backward compatible, we support the case where we have no prefix.

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -226,20 +242,83 @@ public static String toString(final Class<?> clazz, final Object... args) {
...
+			// My decision was to keep legacy working the same.
+			return CommonHelper.class.getResourceAsStream(path);
+		case CLASSPATH_PREFIX:
+			return Thread.currentThread().getContextClassLoader().getResourceAsStream(path);
+		case HTTP_PREFIX:
+			logger.warn("IdP metadata is retrieved from an insecure http endpoint [{}]", path);
+			return getInputStreamViaHttp(name);
+		case HTTPS_PREFIX:
+			return getInputStreamViaHttp(name);
+		default:
```
leleuj: And you have a `default` case...

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -226,20 +242,83 @@ public static String toString(final Class<?> clazz, final Object... args) {
...
+			} catch (FileNotFoundException e) {
+				throw new TechnicalException(e);
+			}
+		}
+
+	}
+
+	private static InputStream getInputStreamViaHttp(String name) {
+		URLConnection con = null;
+		try {
```
leleuj: Very strange: no `finally` to close the connection?

pac4j-core/src/test/java/org/pac4j/core/util/CommonHelperTests.java
```diff
@@ -146,6 +146,11 @@ public void testAreEqualsIgnoreCaseAndTrimFails() {
         assertFalse(CommonHelper.areEqualsIgnoreCaseAndTrim(VALUE, null));
         assertFalse(CommonHelper.areEqualsIgnoreCaseAndTrim(NAME, VALUE));
     }
+    
+    @Test
+    public void testClassPathPrefix() {
```
leleuj: Just one test for all these changes.

pac4j-core/src/main/java/org/pac4j/core/io/Resource.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Interface for a resource descriptor that abstracts from the actual
+ * type of underlying resource, such as a file or class path resource.
+ *
+ * <p>An InputStream can be opened for every resource if it exists in
+ * physical form, but a URL or File handle can just be returned for
+ * certain resources. The actual behavior is implementation-specific.
+ *
+ */
+public interface Resource extends InputStreamSource {
```
leleuj: I don't think we use `InputStreamSource` solely: can we merge it with `Resource`?

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -226,20 +242,83 @@ public static String toString(final Class<?> clazz, final Object... args) {
...
+			// The choice here was to keep legacy behavior and remove / prior to
+			// calling classloader.getResourceAsStream.. or make it work exactly
+			// as it did before but have different behavior for resource: and
+			// classpath:
+			// My decision was to keep legacy working the same.
+			return CommonHelper.class.getResourceAsStream(path);
+		case CLASSPATH_PREFIX:
+			return Thread.currentThread().getContextClassLoader().getResourceAsStream(path);
+		case HTTP_PREFIX:
+			logger.warn("IdP metadata is retrieved from an insecure http endpoint [{}]", path);
```
leleuj: This SAML log has nothing to do here, in this generic `CommonHelper` method.

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -226,20 +242,83 @@ public static String toString(final Class<?> clazz, final Object... args) {
...
+		} catch (IOException ex) {
+			// Close the HTTP connection (if applicable).
+			if (con instanceof HttpURLConnection) {
+				((HttpURLConnection) con).disconnect();
+			}
+			throw new TechnicalException(ex);
+		}
+	}
+
+	public static Resource getResource(final String idpMetadataPath) {
```
leleuj: `idMetadataPath`:  this seems specific to SAML.

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
```diff
@@ -19,6 +22,9 @@
  * @since 1.7
  */
 public final class SAML2ClientConfiguration implements Cloneable {
+	private KeyStore keyStore;
+
+	private Resource keystoreResource;
 
     private String keystorePath;
```
leleuj: Do we still really need a `keystorePath`? Cannot it be some encapsulated data of a file resource for the `keystoreResource`?

leleuj:
I started reviewing the changes and stopped as there are already many comments. It's getting better though.

First of all, this pull request should only target the introduction of the `Resource` concept. Nothing more. No changes in `Clients` or `FormClient`. I see many side-effects changes as well. These must be justified or removed.

Regarding the core changes, I'm in line with adding new interfaces. It seems that `InputStreamResource` could be merged in `Resource` as it's not used anywhere, but maybe you need it for Spring extensions. See my different comments...

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
garpinc: Mainly I wanted all your original code paths to still work so you could pass either the original keystorePath or the new keystoreResource. I figured someone in the wild may have made a call to the old methods and I didn't want my changes to break it. Is this not desirable?

pac4j-core/src/main/java/org/pac4j/core/io/Resource.java
garpinc: WriteableResource extends Resource so I don't think they should be merged for that reason. Resource is the common interface, input is for read and writeable is for write

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
garpinc: Caller is responsible for closing inputstream. I added that to comments.

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
garpinc: I added the if null for findbugs which told me that the prefix maybe null.. I'll take it out and see if travis will build

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
garpinc: I added the if null for findbugs which told me that the prefix maybe null.. I'll take it out and see if travis will build

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
garpinc: I always do that for performance reasons: http://learnertobeginner.blogspot.com/2010/06/how-to-use-isdebugenabled-effectly.html. Shall I remove it as requested?

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
garpinc: Mainly I wanted all your original code paths to still work so you could pass either the original keystorePath or the new keystoreResource. I figured someone in the wild may have made a call to the old methods and I didn't want my changes to break it. Is this not desirable?

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
garpinc: In your https://github.com/pac4j/spring-security-pac4j-demo/blob/master/src/main/resources/securityContext.xml you had 

> <property name="serviceProviderEntityId" value="http://localhost:8080/callback?client_name=SAML2Client" />

I basically thought instead I could just refer to CALLBACK_URL and in those cases it would just put the callback url rather than hardcoding the config defined somewhere else.

pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java
garpinc: This change addresses a bug I saw when a login to the authentication URL was accessed via an ajax call. When there is an ajax call you can't do a redirect because the browser will redirect before the ajax caller get's control. Hence this fixes that and for ajax it doesn't do a redirect.

pac4j-core/src/main/java/org/pac4j/core/io/InputStreamSource.java
garpinc: Added it for next checkin

pac4j-core/src/main/java/org/pac4j/core/io/FileSystemResource.java
garpinc: Agreed.. implemented for next checkin

pac4j-core/src/main/java/org/pac4j/core/io/FileSystemResource.java
garpinc: Added some javadoc

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
garpinc: basically if on callback url there is no clientName specified you can configure in a default client., It's not related to resource changes.. I'm just not sure how to make that change separate at this point.

pac4j-core/src/test/java/org/pac4j/core/util/CommonHelperTests.java
garpinc: Mainly I thought your other tests were covering most of my code changes already especially since the paths are now converted to resources in code.

garpinc:
Also regarding InputStreamResource. I do need it for spring extensions

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
leleuj: Sure. But, I imagine you can remove the internal `keystorePath ` property and keep the setter (and getter) to initialize a specific file resource in the `keystoreResource` based on the provided path: it would be better: backward compatible and cleaner a the same time.

pac4j-core/src/main/java/org/pac4j/core/io/Resource.java
leleuj: But no other interface extends from `InputStreamSource` directly, doesn't it? So in our case, a `Resource` is necessarily a `InputStreamSource` isn't it? We never distinguish between both interfaces.
Or do you directly use the `InputStreamResource` for your Spring extension?
That said, it might be cleaner to keep both for clarity.

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
leleuj: OK. How the connection is closed if no exception occurs?

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
leleuj: Findbugs may be "wrong" or too picky sometimes and we can exclude applied rules if it makes sense.
Here, we must be backward compatible for all users using this without a prefix and this is certainly the majority of the pac4j users.

I'm surprised that no test fails on this: we lack tests on this (change).

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
leleuj: Yes, you should remove it since slf4j version 1.7 I think, but you also need to change the way you log things: from `logger.trace("myvalue is: " + myvalue)` to `logger.trace("myvalue: {}", myvalue);`. Following this rule, there will be no performance impact (no String building, these are just parameters).

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
leleuj: It depends. Here, I don't understand why we have **at the same time** both parameters: isn't it one or the other?

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
leleuj: I get the idea, but can you remove this for now? Let's try to keep one change at a time, there are already many things to account for.

pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java
leleuj: OK. If there is a problem, we'll fix it, but in another pull request. Just recommit the original current master file and the change will go away from the pull request. We'll deal with that problem later on. You can already open a Github issue so we don't forget about it.

pac4j-core/src/main/java/org/pac4j/core/io/InputStreamSource.java
leleuj: After the hard work you are doing, I guess you don't want to forget the `author` annotation as well. I think I saw the problem in several other files.

pac4j-core/src/main/java/org/pac4j/core/io/FileSystemResource.java
leleuj: It might be the case for other classes.

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
leleuj: For a different need, we need a different pull request. Recommit the master file version.
Notice this must be discussed before submitting a pull request: do we really want to bring in that behaviour? I'm really not sure.

pac4j-core/src/test/java/org/pac4j/core/util/CommonHelperTests.java
leleuj: It should, but nothing is perfect. I think there is a lack of tests here: we should have ones for all prefixes and for the use case with no prefix.

leleuj:
OK. So let's keep `InputStreamResource`. I replied to all your comments, I see that some of my remarks have not been accounted for so far...

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
garpinc: It's a private method used by both external public methods that only allow application of one or other.. Then in the private method I make sure that one or the other is supplied.

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
garpinc: Once again the job of this is not to close the stream.. the caller has to enclose the call to this method with the finally to close the stream.

garpinc:
I refactored my end and removed InputStreamResource per your request, removed various checkins to re-add in a future pull request, and added the test cases for the prefixes. let me know how this iteration is

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
leleuj: I was mistaken, I meant: how the connection is supposed to be disconnected in regular use case (no error)? By the caller?

leleuj:
Thanks for your tenacity on this. I will review it shortly.

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -19,7 +26,16 @@
  */
 public final class CommonHelper {
 
-    public static final String RESOURCE_PREFIX = "resource:";
+    private static final Logger logger = LoggerFactory.getLogger(CommonHelper.class);
+
+	public static final String RESOURCE_PREFIX = "resource";
+	public static final String CLASSPATH_PREFIX = "classpath";
+	public static final String HTTP_PREFIX = "http";
```
leleuj: In fact, we already these constants in the `HttpConstants` interface.

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
```diff
@@ -59,25 +67,56 @@ public SAML2ClientConfiguration() {}
...
+		this.keyStore = keyStore;
+        this.keyStoreAlias = keyStoreAlias;
+        this.keyStoreType = keyStoreType;
+        this.keystoreResource = keystoreResource;
+        if (this.keystoreResource == null) {
+        	this.keystoreResource = CommonHelper.getResource(keystorePath);
+        }
         this.keystorePassword = keystorePassword;
         this.privateKeyPassword = privateKeyPassword;
+		this.identityProviderMetadataResource = identityProviderMetadataResource;
```
leleuj: Cannot we do like for the `keystorePath` and `keystoreResource` regarding the `identityProviderMetadataPath` and `identityProviderMetadataResource`?

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
```diff
@@ -43,14 +48,43 @@
...
+				"Either keyStore, keyStoreResource or keyStorePath must be provided");
+
+		final KeyStore keyStoreToUse;
+		if (keyStore != null) {
+			keyStoreToUse = keyStore;
+		} else {
+			final InputStream inputStream;
+			if (keyStoreResource != null) {
+				try {
+					inputStream = keyStoreResource.getInputStream();
```
leleuj: I don't see where this stream is closed...

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
```diff
@@ -43,14 +48,43 @@
...
+				} catch (IOException e) {
```
leleuj: No `finally` to close the stream?

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
```diff
@@ -43,14 +48,43 @@
...
+					throw new TechnicalException(e);
+				}
+			} else {
+				inputStream = CommonHelper.getInputStreamFromName(keyStorePath);
```
leleuj: No `try/catch IOException` here?

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
```diff
@@ -114,12 +148,17 @@ private KeyStore loadKeyStore(final InputStream inputStream, final String storeP
...
-    private String getPrivateKeyAlias(final KeyStore keyStore) {
+	private String getPrivateKeyAlias(final KeyStore keyStore, final String keyStoreAlias) {
         try {
             final Enumeration<String> aliases = keyStore.aliases();
-            if (aliases.hasMoreElements()) {
-                return aliases.nextElement();
-            }
+			while (aliases.hasMoreElements()) {
+	            String currentAlias = aliases.nextElement();
+				if (keyStoreAlias == null) {
```
leleuj: I'm not sure to understand this use case.

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
```diff
@@ -114,12 +148,17 @@ private KeyStore loadKeyStore(final InputStream inputStream, final String storeP
...
+					return currentAlias;
+				} else if (currentAlias.equals(keyStoreAlias)) {
```
leleuj: I think this condition could be merge with the previous one.

leleuj:
+1 Looks good to me with a few comments to address before the merge

Waiting also for @mmoayyed 's feedback on the SAML changes...

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
garpinc: From http://docs.spring.io/autorepo/docs/spring/3.2.x/spring-framework-reference/html/resources.html  "It is expected that each invocation returns a fresh InputStream. It is the responsibility of the caller to close the stream.". Here are some examples in regular use case closing the stream: http://www.mkyong.com/spring/spring-resource-loader-with-getresource-example/

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
garpinc: It's closed in the loadKeyStore method.. I didn't change that.

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
garpinc: I agree there should be a finally I'll add it..

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
garpinc: Previously if the keystore had more than one alias you just picked the first one. I have a scenario where my keystore contains more than one alias and the one i need for saml isn't always the first one so if you specify which alias then it uses the correct one.

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
garpinc: Not sure I understand what you mean here.

garpinc:
I have addressed the issues you mentioned.. Also re-implemented these changes on 1.8.x pull request. Will keep those in sync with your comments here.

pac4j-core/src/main/java/org/pac4j/core/io/Resource.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Interface for a resource descriptor that abstracts from the actual
+ * type of underlying resource, such as a file or class path resource.
+ *
+ * <p>An InputStream can be opened for every resource if it exists in
+ * physical form, but a URL or File handle can just be returned for
+ * certain resources. The actual behavior is implementation-specific.
+ * 
+ *  @since 1.9.0
```
leleuj: I'd like to see the `@author` annotation here also.

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
leleuj: I agree: `inputStream.close();` must be handled by the caller. But how the `con.disconnect()` is called?

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
leleuj: In both cases, you return the `currentAlias`, you could have one condition with `||`.

pac4j-saml/src/main/java/org/pac4j/saml/crypto/KeyStoreCredentialProvider.java
leleuj: OK

leleuj:
A few comments more, but it looks good globally.

@mmoayyed : is it ok for you?

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
garpinc: http://kingori.co/minutae/2013/04/httpurlconnection-disconnect/ and http://stackoverflow.com/questions/4767553/safe-use-of-httpurlconnection.. Basically this code is the same as it is in spring.. When the inputstream is close it can free up resources and depending on keep-alive setting will do the disconnect as far as I understand it.

leleuj: OK

pac4j-core/pom.xml
```diff
@@ -45,6 +45,12 @@
         </dependency>
         <dependency>
             <groupId>org.springframework</groupId>
+            <artifactId>spring-core</artifactId>
+            <scope>compile</scope>
+            <version>${spring.version}</version>
```
leleuj: It should be marked with `<optional>true</optional>`  because I don't want to pull the Spring dependency for everybody.

pac4j-core/src/main/java/org/pac4j/core/io/SpringResourceWrapper.java
```diff
@@ -0,0 +1,34 @@
+package org.pac4j.core.io;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * A wrapper that adapts a spring {@link org.springframework.core.io.Resource} to a pac4j {@link Resource}.
+ * 
+ * @author Keith Garry Boyce
+ * @since 1.4.3
```
leleuj: Version 1.9.0.

pac4j-core/src/main/java/org/pac4j/core/io/SpringWritableResourceWrapper.java
```diff
@@ -0,0 +1,25 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+
+import org.springframework.core.io.Resource;
+
+/**
+ * A wrapper that adapts a spring {@link org.springframework.core.io.WritableResource} to a pac4j {@link WritableResource}.
+ * 
+ * @author Keith Garry Boyce
+ * @since 1.4.3
```
leleuj: Version 1.9.0

leleuj:
Let's fix my latest comments before I merge.

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
```diff
@@ -150,7 +150,7 @@ protected void initSAMLContextProvider(final MetadataResolver metadataManager) {
     }
 
     protected MetadataResolver initServiceProviderMetadataResolver(final WebContext context) {
-        this.spMetadataResolver = new SAML2ServiceProviderMetadataResolver(this.configuration.getServiceProviderMetadataPath(),
+		this.spMetadataResolver = new SAML2ServiceProviderMetadataResolver(this.configuration.getServiceProviderMetadataPath(),
```
leleuj: I think that passing the `configuration` could be sufficient, couldn't it? To mirror what's done for the identity provider...

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2IdentityProviderMetadataResolver.java
```diff
@@ -41,16 +38,33 @@
     protected final static String HTTP_PREFIX = "http";
```
leleuj: These two constants are no longer necessary.

leleuj:
Thank your very much for your work. Thanks for your patience and tenacity through this long merging process.

PR #521, opened 2016-03-11
Use idp descriptor to retrieve signature signing parameters

hmerat:
In our project which uses pac4j to connect to a saml idp, we had an issue with the DigestMethod used by pac4j to talk to the idp (it's always sha512). The used idp doesn't support sha512 as digest method.

According to the documentation (https://github.com/pac4j/pac4j/wiki/Clients in SAML support/Additional configuration) the signature parameters are determined based on the idp metadata and the pac4j configuration. So we tried to configure pac4j to change the digest method, but the request still used sha512. We also tried to change the idp metadata in order to add information about the supported digest methods but it still failed.

When looking at the code, it seems to us that pac4j was using the sp metadata (which is generated with many supported digest methods, sha512 being the first in the list) instead of the idp metadata. We changed it to use the idp and then it fixed our issue (we can now change the idp metadata to specify the supported digest method or configure pac4j).

This pull request contains the fix we did.

Please tell us if we misunderstood the pac4j code.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2WebSSOMessageSender.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2WebSSOMessageSender.java
@@ -74,7 +74,7 @@ public class SAML2WebSSOMessageSender implements SAML2MessageSender<AuthnRequest
         outboundContext.getSAMLPeerEntityContext().setEntityId(context.getSAMLPeerEntityContext().getEntityId());
         outboundContext.getSAMLProtocolContext().setProtocol(context.getSAMLProtocolContext().getProtocol());
         outboundContext.getSecurityParametersContext()
-                .setSignatureSigningParameters(this.signatureSigningParametersProvider.build(spDescriptor));
+                .setSignatureSigningParameters(this.signatureSigningParametersProvider.build(idpssoDescriptor));
 
         if (relayState != null) {
             outboundContext.getSAMLBindingContext().setRelayState(relayState.toString());


leleuj:
Thanks. @mmoayyed ?

mmoayyed:
I don't think this will work. 

The message that is sent to the IdP is supposed to explain "to the IdP": here is the digest method I support and care about. It's the IdP's job to determine that and fail if needed. If your IdP does not support Sha512, then your SP should not ask for Sha512. Change your SP metadata instead?

hmerat:
The SP metadata is generated by pac4j. The code generating the metadata extension is in SAML2MetadataGenerator#generateMetadataExtensions. It doesn't seem possible to change the extensions in the generated metadata.
DefaultSignatureSigningParametersProvider#build(SSODescriptor) uses the extensions to choose which DigestMethod to use. So it will always choose sha512.

Maybe I'm missing something

mmoayyed:
Correct. That's the fix we'd want to make there inside `SAML2MetadataGenerator`. You can for the time being override that class and specify the extensions you  want, but we likely should expose config options for this sort of thing.

mmoayyed:
You notice that `generateMetadataExtensions` is marked as protected. So for now, you can override it but this PR should turn into a few additional opts by the SAML2 config here that let you choose what extensions should be supported in the metadata.

mmoayyed:
Also note that the saml2 config allows you to whitelist and blacklist signature signing params right in the config. Have you tried that?

hmerat:
I thought we tried the blacklist params but I just tried again and indeed it seems to work now. Maybe we made a mistake during our previous try.
Concerning the whitelist params, is SAML2ClientConfiguration#setSignatureReferenceDigestMethods the method you're talking about? I tried it and it didn't seem to work. I think that the metadata extensions, if defined, are used instead of the configuration params.
I found a whitelist in BasicSignatureSigningConfiguration but It seems we can't set it through the SAML2ClientConfiguration.

The edition of the blacklist seems to solve our issue, thank you for your help

mmoayyed:
The metadata extension defined what is supported and requested by the SP. You can via config decide which of those is appropriate and which can be black/white listed. For the exact behavior of what those settings mean, consider looking up the javadocs for opensaml.

PR #524, opened 2016-03-16
Add isEmpty / isNotEmpty for collections methods

leleuj:


pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -86,6 +87,26 @@ public static boolean areNotEquals(final String s1, final String s2) {
...
+    }
+
+    /**
+     * Return if a collection is not empty.
+     *
+     * @param coll a collection
+     * @return whether it is not empty
+     */
+    public static boolean isNotEmpty(final Collection coll) {
+        return coll != null && !coll.isEmpty();
```
mmoayyed: You could simply turn this into !isEmpty()

leleuj: I want to be able to check that without worrying about the fact that the collection can be null. Hence the first check.

mmoayyed: To clarify, my suggestion was not so much:

```
return !coll.isEmpty()
```

But more like:

```
return !isEmpty(coll)
```

I believe this is what commons does too.

leleuj: OK. I get it now. Will change.

PR #527, opened 2016-03-22
Support for storing refresh token in OIDC profile

### LINKED ISSUE #526: Support for storing Refresh Token in OIDC profile
Nowadays, access_token and id_token are stored in OIDC profile. Some identity providers, like Google, give you short-lived tokens and offer a refresh token service. It would be nice to support [refresh tokens](http://openid.net/specs/openid-connect-core-1_0.html#RefreshTokens).

rsoletob:
At now, refresh_token are stored in profile if it is retured by the identity provider.

Fixes #526

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/OidcProfile.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/OidcProfile.java
@@ -28,6 +28,7 @@ public class OidcProfile extends CommonProfile implements Externalizable {
 
     private transient static final String ACCESS_TOKEN = "access_token";
     private transient static final String ID_TOKEN = "id_token";
+    private transient static final String REFRESH_TOKEN = "refresh_token";
 
     public OidcProfile() {
     }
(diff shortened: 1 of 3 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/profile/OidcProfile.java
```diff
@@ -55,6 +56,14 @@ public JWT getIdToken() throws ParseException {
         return null;
     }
 
+    public String getRefreshTokenString() {
```
mmoayyed: Can a refresh token become a JWT? If so, do we need an extra getter here?

rsoletob: I think it can't become a JWT. An example of refresh_token is `13/FuJRLB-4xn_4rd9iJPAUL0-gApRRtpDYuXH5ub5uW5Ne0-oSohI6jUTnlb1cYPMIHq0Ne63h8HdZjAidLFlgNg==` by [Dex](https://github.com/coreos/dex)  and `25/3Uv_Sp3komLpvJTPHUr4bnTE3J-HkpECwOdXmpG-hk8MEudVrK5jSpoR30zcRFq6` by Google. I tried both in [jwt.io](http://jwt.io/) without luck.

mmoayyed: Fair enough. Thanks for the explanation

leleuj: Yes, I think refresh tokens are like access tokens: opaque string, by opposition to the id token.

leleuj:
Looks good to me. I think we can merge as is.
The next question would be how to get a new access token from a refresh token, but I guess it may depend on identity providers.

PR #532, opened 2016-03-29
Ability to load roles from a list value as well as strings.

ikogan:
Currently, `FromAttributesAuthorizationGenerator` only loads roles from string attributes with a comma separated list of roles. Sometimes, service (notably CAS), return roles as a list (notably `memberOf` from LDAP). This allows attribute values to be lists or arrays.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/authorization/generator/FromAttributesAuthorizationGenerator.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/authorization/generator/FromAttributesAuthorizationGenerator.java
@@ -1,6 +1,7 @@
 package org.pac4j.core.authorization.generator;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.StringTokenizer;
 
 import org.pac4j.core.profile.CommonProfile;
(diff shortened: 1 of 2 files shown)


pac4j-core/src/main/java/org/pac4j/core/authorization/generator/FromAttributesAuthorizationGenerator.java
```diff
@@ -43,20 +44,36 @@ private void generateAuth(final U profile, final String[] attributes, final bool
...
-                            profile.addRole(st.nextToken());
-                        } else {
-                            profile.addPermission(st.nextToken());
+                if (value != null) {
+                    if(value instanceof String) {
+                        final StringTokenizer st = new StringTokenizer((String) value, this.splitChar);
+                        while (st.hasMoreTokens()) {
+                            setAuth(profile, st.nextToken(), isRole);
+                        }
+                    } else if(value.getClass().isArray() && value.getClass().getComponentType().isAssignableFrom(String.class)) {
```
leleuj: I think we could even remove the test to check that elements are String.

leleuj:
Thanks for your contribution. Will merge.

ikogan:
I don't suppose this something small enough that it could be backported to 1.8.x? It'd be nice to be able to have this in something that would retain compatibility with CAS 4.2.x.

leleuj:
Sure. Feel free to open a new pull request targeting branch 1.8.

PR #534, opened 2016-03-30
Ability to load roles from a list value as well as strings. Backport from 1.9: 8ad89468f5a8133188fde1d0eb5243f6d3a755ba

ikogan:
Backporting the following change from 1.9:

Ability to load roles from a list value as well as strings.

Currently, `FromAttributesAuthorizationGenerator` only loads roles from string attributes with a comma separated list of roles. Sometimes, service (notably CAS), return roles as a list (notably `memberOf` from LDAP). This allows attribute values to be lists or arrays.


leleuj:
Thanks for the backport. I'll release a 1.8.8 in a few weeks (waiting for other improvements / bug fixes). Just remind me if it doesn't show up.

PR #538, opened 2016-03-31
Don't swallow CredentialsException in DirectClientV2#getCredentials

YannRobert:
Any CredentialsException occuring while extracting or validating the credentials should be thrown up by org.pac4j.core.client.DirectClientV2#getCredentials

In order to be able to present an relevant message to the user, we would like the CredentialsException to be thrown a level up, instead of returning null. It will be up to the caller to decide what to do with it.

For instance, if the CredentialsException indicated 'the password does not match' or 'the account is disabled', and should we want to display that message to the user, then we should be able to.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
@@ -3,7 +3,6 @@ package org.pac4j.core.client;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.credentials.authenticator.LocalCachingAuthenticator;
 import org.pac4j.core.credentials.extractor.CredentialsExtractor;
-import org.pac4j.core.exception.CredentialsException;
 import org.pac4j.core.exception.TechnicalException;
 import org.pac4j.core.profile.CommonProfile;
 import org.pac4j.core.credentials.authenticator.Authenticator;
@@ -41,17 +40,12 @@ public abstract class DirectClientV2<C extends Credentials, U extends CommonProf
     @Override
     public C getCredentials(final WebContext context) throws RequiresHttpAction {
         init(context);
-        try {
-            final C credentials = this.credentialsExtractor.extract(context);
-            if (credentials == null) {
-                return null;
-            }
-            this.authenticator.validate(credentials);
-            return credentials;
-        } catch (CredentialsException e) {
-            logger.error("Failed to retrieve or validate credentials", e);
+        final C credentials = this.credentialsExtractor.extract(context);
+        if (credentials == null) {
             return null;
         }
+        this.authenticator.validate(credentials);
+        return credentials;
     }
 
     @Override
--- a/pac4j-http/src/test/java/org/pac4j/http/client/direct/DirectFormClientTests.java
+++ b/pac4j-http/src/test/java/org/pac4j/http/client/direct/DirectFormClientTests.java
@@ -4,6 +4,7 @@ import org.junit.Test;
 import org.pac4j.core.context.MockWebContext;
 import org.pac4j.core.context.Pac4jConstants;
 import org.pac4j.core.credentials.authenticator.LocalCachingAuthenticator;
+import org.pac4j.core.exception.CredentialsException;
 import org.pac4j.core.exception.RequiresHttpAction;
 import org.pac4j.core.profile.CommonProfile;
 import org.pac4j.core.profile.ProfileHelper;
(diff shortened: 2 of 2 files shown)


leleuj:
I understand your need and it makes sense. Though, the proposed solution will break things.

An `Authenticator` is expected to throw a `CredentialsException` in case of authentication failures. It's handled by the client. If you remove that, `CredentialsException` will "break" the CallbackFilter, like for example: https://github.com/pac4j/j2e-pac4j/blob/master/src/main/java/org/pac4j/j2e/filter/CallbackFilter.java#L78

So I'm in line with the change, but it requires another solution: do you want to perform a redirection if the authentication fails with a specific error message? Or keep that error message in session?

YannRobert:
> I understand your need and it makes sense. Though, the proposed solution will break things.

I understand the proposed solution may not be correct. It's more a starting point to the discussion on the feature I need. It's a good point you understand the need and want to help find a solution.

> do you want to perform a redirection if the authentication fails with a specific error message? Or keep that error message in session?

None of the above. The web application is state-less and session-less, and in that case I want a direct response to the client.
Maybe instead of keeping that error message in session like you proposed, it could be saved in a request attribute instead? That may work.

Imagine a ```curl``` client requesting a JSON webservice with an ```"Autorization: Bearer "``` header based authentication.
If the authentication fails, I want to send a HTTP 401 Unauthorized status (that's easy) but I want to help the end-user with a JSON body containing a detailed message (that's hard) like 
```
{
  "message" : "Invalid credentials : invalid token authentication : TOKEN_HAS_EXPIRED"
}
```

In order to do that, I have 
1) extended ```org.pac4j.http.client.direct.HeaderClient``` (and every other ```DirectClient```s I use) to override ```#getCredentials(WebContext)``` and make it throw the ```CredentialsException``` instead of swallowing it (like in the current Pull Request)
2) rewritten the ```org.pac4j.springframework.security.web.ClientAuthenticationFilter``` (because it's a ```final class``` by the way, and makes multiple use of the ```HttpSession)``` so that it catches ```CredentialsException``` a then ```throw new AuthenticationCredentialsException("Invalid credentials : " + ce.getMessage(), ce)```

So I have already managed to do what I need. Now I would like to avoid having to extend every ```DirectClient``` classes to change the ```#getCredentials(WebContext)``` behavior.

Any thoughts?

leleuj:
OK. It's clearer.

In fact, the `ClientAuthenticationFilter` is not meant to be used with direct clients (only indirect clients). It's defined on a callback url and should be called after a successful authentication at a remote identity provider.
It's not the first time I have requests to use direct clients with Spring Security. As I will rebuild the spring-security-pac4j library on pac4j v1.9, it's maybe time to do something.
In other implementations, in new versions, you'll be able to break the flow by returning a `RequiresHttpAction` from the `Authenticator`: it could do the job for you: `throw Requires.unauthorized(...)`

But first of all, why do you use spring-security-pac4j and not spring-webmvc-pac4j for example?

YannRobert:
Yes I realized ```ClientAuthenticationFilter``` was only meant for indirect clients.

I think it is possible to have a Filter that works for both indirect and direct clients.
If you're interested I can contribute my Filter.

My project already use spring-security since a long time ago. And I only recently introduced pac4j in the mix. So spring-security-pac4j was a natural fit I supposed.
First, I used pac4j in order to add a form-based authentication (```indirect.FormClient```) and ```ClientAuthenticationFilter``` was almost OK for that (exception dependencies on the ```HttpSession```).
Then I needed to add a token-based authentication (```direct.HeaderClient``` and ```direct.ParameterClient```).

I created a Custom ```org.springframework.security.web.util.matcher.RequestMatcher``` so that my ```org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter``` matches both the callbackURI and the Authentication Header and Parameter.

By the way, it would be nice to have my Custom RequestMatcher replaced by a RequestMatcher provided by the ```org.pac4j.core.client.Clients``` instance, that checks if a ```HttpServletRequest``` matches one of the ```Client```'s ```Extractor```, and find the appropriate Client to handle that request.

leleuj:
In fact, in the pac4j security model, we have two filters: the `CallbackFilter` (here, the `ClientAuthenticationFilter`)  which is only available on the /callback url to finish the authentication process for indirect clients. The dependency on the HTTP session seems natural for saving/restoring the originally requested url
- the `RequiresAuthenticationFilter` now named `SecurityFilter` which protects any url it is defined on and performs authentication for direct clients (the reference implementation is: https://github.com/pac4j/j2e-pac4j) and redirects to the identity provider fro indirect clients.

If you think you can propose something that matches the pac4j guidelines, feel free to submit a pull request.

PR #543, opened 2016-04-06
isAnonymous and isAuthenticated authorizers

leleuj:


pac4j-core/src/main/java/org/pac4j/core/authorization/authorizer/IsAuthenticatedAuthorizer.java
```diff
@@ -0,0 +1,45 @@
...
+    public boolean isProfileAuthorized(final WebContext context, final U profile) throws RequiresHttpAction {
+
+        if (profile != null && !(profile instanceof AnonymousProfile)) {
+            return true;
+        } else {
+            if (this.redirectionUrl != null) {
+                throw RequiresHttpAction.redirect("user should be authenticated", context, this.redirectionUrl);
+            } else {
+                return false;
+            }
```
mmoayyed: Looks like the above block is repeated. Could we extract this bit out?

leleuj: Yes, there won't be a huge benefit, but indeed, some code is duplicated. I will fix the build as well.

Issue #536, opened 2016-03-31
Handle ajax requests to not redirect


garpinc:
change in garpinc fork commit ba3c351

diff --git a/pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java b/pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java
index e70528f..edfd5cf 100644
--- a/pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java
+++ b/pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java
@@ -79,12 +79,6 @@
             credentials = getCredentialsExtractor().extract(context);
             logger.debug("usernamePasswordCredentials: {}", credentials);
             if (credentials == null) {
-				// it's an AJAX request -> unauthorized (instead of a
-				// redirection)
-				if (getAjaxRequestResolver().isAjax(context)) {
-					logger.info("AJAX request detected -> returning 401");
-					throw RequiresHttpAction.unauthorized("AJAX request -> 401", context, null);
-				} else {
 					String redirectionUrl = CommonHelper.addParameter(this.loginUrl, this.usernameParameter, username);
 					redirectionUrl = CommonHelper.addParameter(redirectionUrl, ERROR_PARAMETER, MISSING_FIELD_ERROR);
 					logger.debug("redirectionUrl: {}", redirectionUrl);
@@ -92,15 +86,9 @@
 					logger.debug(message);
 					throw RequiresHttpAction.redirect(message, context, redirectionUrl);
 				}
-            }
             // validate credentials
             getAuthenticator().validate(credentials);
         } catch (final CredentialsException e) {
-			// it's an AJAX request -> forbidden (instead of a redirection)
-			if (getAjaxRequestResolver().isAjax(context)) {
-				logger.info("AJAX request detected -> returning 403");
-				throw RequiresHttpAction.forbidden("AJAX request -> 403", context);
-			} else {
 				String redirectionUrl = CommonHelper.addParameter(this.loginUrl, this.usernameParameter, username);
 				String errorMessage = computeErrorMessage(e);
 				redirectionUrl = CommonHelper.addParameter(redirectionUrl, ERROR_PARAMETER, errorMessage);
@@ -109,7 +97,6 @@
 				logger.debug(message);
 				throw RequiresHttpAction.redirect(message, context, redirectionUrl);
 			}
-        }
 
         return credentials;
     }

leleuj:
[...]

PR #545, opened 2016-04-14
Git issue #536

garpinc:


pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java
```diff
@@ -79,23 +79,36 @@ protected UsernamePasswordCredentials retrieveCredentials(final WebContext conte
...
             if (credentials == null) {
-                String redirectionUrl = CommonHelper.addParameter(this.loginUrl, this.usernameParameter, username);
-                redirectionUrl = CommonHelper.addParameter(redirectionUrl, ERROR_PARAMETER, MISSING_FIELD_ERROR);
-                logger.debug("redirectionUrl: {}", redirectionUrl);
-                final String message = "Username and password cannot be blank -> return to the form with error";
-                logger.debug(message);
-                throw RequiresHttpAction.redirect(message, context, redirectionUrl);
+				// it's an AJAX request -> unauthorized (instead of a
+				// redirection)
+				if (getAjaxRequestResolver().isAjax(context)) {
```
leleuj: This is where I realized that both blocks (line 82 - 94 and line 99 - 111) of source code looks very close: can we extract that in a method? Thanks

garpinc:
Pushing this on another push request

Issue #535, opened 2016-03-31
Add the ability to set a default client


garpinc:
In garry fork ba3c351:

	private Client defaultClient;
In findClients:
		if (name == null && defaultClient != null) {
			return defaultClient;
		}

	public void setDefaultClient(Client defaultClient) {
		this.defaultClient = defaultClient;
	}

leleuj:
What use cases do you want to address with that change?

garpinc:
The use case is a client that wants to authorize at a particular url with a username and password. The client already exists and i'm modifying pac4j to suit but the client doesn't know about the clientName parameter and I don't want to have to change all the clients.

leleuj:
I'm sorry but I don't understand. Currently, the `findClient(context)` is only used in the callback filter and the callback url is called with a `clientName` parameter after authenticating at a remote provider.
I don't see when this `findClient` method is called without any `clientName`.

garpinc:
Although within the framework that is true nothing prevents an external application from making a request to the callback url. In my case I have a third party app that attempts to authenticate at /someurl with some_username and some_password and I want the one callback url to service authentications from both within the framework an this type of authentication. That third party app doesn't know about clientName but obviously I want it to use the formClient.

leleuj:
The callback url must be called with a `client_name` otherwise, an error is returned. If you have a third app which posts credentials on your app, you certainly shouldn't use the app login page but provide a specific endpoint.

garpinc:
Isn't the whole idea of security to reduce number of attack vectors?

leleuj:
I'm not sure to understand what the security issue you try to highlight here. The callback url should always be called with a `client_name` parameter. It's the behaviour. All other use cases lead to an error.
In your case, I think there is a better way to define your security configuration.

garpinc:
I'm trying to highlight that single sign on means single sign on.. i.e: one place where the credentials can  be provided.. If I add another place it's another attack vector. If a default is provided then there would be no error and if not the behavior would be as it is now hence nothing would change for most people who currently don't supply the default. What's your reservation?

mmoayyed:
> The use case is a client that wants to authorize at a particular url with a username and password. The client already exists and i'm modifying pac4j to suit but the client doesn't know about the clientName parameter and I don't want to have to change all the clients.

Sorry, but I also have no idea what you are trying to accomplish. Could you elaborate on your use case and requirement? This is not clear to either of us. Start from the top. 

- You have an application
[...]

PR #546, opened 2016-04-14
Git issue #535

garpinc:
Unfortunately I still haven't mastered this git stuff and this incorporated the stuff I'm doing on other pull request. Guess I need to branch for each feature so this won't happen.. I can also wait until you've merged the other pull request and then perhaps this will be sorted.. Let me know.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
@@ -32,6 +32,7 @@ public class Clients extends InitializableObject {
     private List<Client> clients;
 
     private String callbackUrl = null;
+	private Client defaultClient;
 
     public Clients() {
     }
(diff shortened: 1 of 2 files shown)


leleuj:
OK. We'll deal with it (both changes in the same pull request).

pac4j-core/src/main/java/org/pac4j/core/client/Clients.java
```diff
@@ -188,5 +192,9 @@ public void setClients(final Client... clients) {
     public String toString() {
         return CommonHelper.toString(this.getClass(), "callbackUrl", this.callbackUrl, "clientNameParameter",
                 this.clientNameParameter, "clients", getClients());
-    }
+	}
+
```
leleuj: Can we have a getter as well?

PR #551, opened 2016-04-18
Fix DB Authenticator password encoder timing attack

samgurtman:
Moved expected password encoding to before user not found error handling as otherwise password would only be encoded if user was found. This may be in the realm of detectable timing variations, allowing a timing attack to determine if a given user exists.


leleuj:
Thanks for your help.

PR #554, opened 2016-04-21
Pac4j 1.8.8 branch

garpinc:
Corrected source jars for osgi to contain right manifest

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -375,17 +375,44 @@
 					<encoding>UTF-8</encoding>
 				</configuration>
 			</plugin>
+			<!-- Build helper maven plugin sets the parsedVersion.osgiVersion property -->
+			<plugin>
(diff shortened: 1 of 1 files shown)


leleuj:
Thanks.

How does it fit with the `maven-bundle-plugin` we have in all submodules? Both are required?

garpinc:
Basically it doesn't modify any of that. I'm using [this](http://rajakannappan.blogspot.com/2010/03/automating-eclipse-source-bundle.html) as reference. And Just verified after a few iterations that it works for me. I modified the pom.xml at root of pac4j doing a regex replace to be able to get the correct module name in the manifest. You may want to ultimately follow that pattern rather than hard coding module name in every pom. As it turns out you strip the "pac4j-" to form the last part of the module name.

leleuj:
OK. Thanks

PR #560, opened 2016-04-28
upgrade to Scribe 2.5

zawn:


leleuj:
Wow! What an incredible PR! I was not expecting so many changes. Before diving into a thoroughfull code review, I have a few general comments: I see false versions (like 1.2.0) for the `@since` annotation, unappropriate header class Javadocs (bad copy/paste) as well as chinese comments (I guess) which must be replaced by english ones...

My main concern is the changes you made on the `BaseOAuthClient` class in the generics definition: can you elaborate on this? Was it absolutely necessary? Thanks

zawn:
I am helpless to do so. 

But after  Scribe  2.2.2 they removed   Api.java. #https://github.com/scribejava/scribejava/commit/bd2d84ffdd7dc24ea4d482dd3e12026d2d775383

Fortunately they later added a generic BaseApi.java
#https://github.com/scribejava/scribejava/commit/22f5c38e7b08af4fb805ae64ae9ac58d6dbb0599

zawn:
Since BaseApi have generic parameters OAuthService, and OAuthService.signRequest () method has been removed, so you must add the generic parameter S is BaseOAuthClient 

OAuthService.java#https://github.com/scribejava/scribejava/commit/d734a4df

This commit remove OAuthService generic parameter token, Token subclass hierarchy has also changed, in order to avoid the ibject type cast,Add generic parameters T

zawn:
QQ and weibo client is wrong I committed to the master branch, has been removed

leleuj:
Thanks for the explanations.

Why did you remove the Weibo and QQ clients? These were very interesting new supports which, I'm sure, many pac4j users could use. **Can you add them back?**

I will take a closer look a the PR.

pac4j-oauth/src/main/java/org/pac4j/oauth/client/BaseOAuth20StateClient.java
```diff
@@ -19,11 +17,10 @@
 
     private static final String STATE_PARAMETER = "#oauth20StateParameter";
 
-    private String stateData;
```
leleuj: Why did you remove the `stateData`?

pac4j-oauth/src/main/java/org/pac4j/oauth/client/BaseOAuth20StateClient.java
```diff
@@ -33,18 +30,13 @@ protected String getStateParameter(final WebContext context) {
     }
 
     @Override
-    protected String retrieveAuthorizationUrl(final WebContext context) throws RequiresHttpAction {
+    protected void internalInit(WebContext context) {
```
leleuj: The state may be a random string, so we cannot have only one instance of the service to generate the authorization url. We need to re-create a new service every time (for each new state parameter).

I don't think this will work.

leleuj:
A few comments to address, but the result is good. Expecting the Weibo and QQ clients back to finish the code review...

pac4j-oauth/src/main/java/org/pac4j/oauth/client/BaseOAuth20StateClient.java
zawn: For single instance, this is a problem, it has been modified.

zawn:
After this the pull request, Weibo and QQ clients  will be submitted as a separate pull request.

Thanks.

pac4j-oauth/src/main/java/org/pac4j/oauth/client/BaseOAuth20StateClient.java
leleuj: OK. Now I see: the `state` parameter is part of the `OAuthConfig`.

leleuj:
Excellent! Merging...

leleuj:
@jotunskij : we have updated the OAuth support on pac4j. We now have manual tests in the https://github.com/pac4j/pac4j/tree/master/pac4j-oauth/src/test/java/org/pac4j/oauth/run package. I haven't been able to setup something for Orcid: could you take care of that?

@indvd00m : I have the same issue for Vk support, it has been updated, but it would be great to have a manual test to confirm everything is ok?

Thanks

PR #561, opened 2016-04-28
Fixes NPE in OidcClient when token endpoint auth methods are undefined.

pauloriordan:
The OIDCProviderMetadata class from nimbus/oauth-oidc returns null, rather than an empty list when these are not specified.

	/*
	 * @return The supported token endpoint authentication methods,
	 *         {@code null} if not specified.
	 */
	public List<ClientAuthenticationMethod> getTokenEndpointAuthMethods() {


pac4j-oidc/src/main/java/org/pac4j/oidc/client/OidcClient.java
```diff
@@ -283,7 +283,7 @@ protected void internalInit(final WebContext context) {
         final List<ClientAuthenticationMethod> methods = getProviderMetadata().getTokenEndpointAuthMethods();
         final ClientAuthenticationMethod method;
 
-        if (!methods.isEmpty()) {
+        if (methods != null && !methods.isEmpty()) {
```
leleuj: Good catch! Can you use the `CommonHelper.isNotEmpty` method?

leleuj:
Thanks for your contribution.

PR #563, opened 2016-05-02
split OAuth1.0,2.0 from OAuthCredentials

zawn:
oauth1,2 term is not the same


zawn:
After the split, the flexibility to customize the credentials used to add  custom OAuth grant flow, without affecting other existing protocol

leleuj:
Looks good to me!

Issue #565, opened 2016-05-05
Add more clients into PropertiesConfigFactory


mmoayyed:
Add the following:

1. DropBox
2. Github
3. Yahoo
4. Google
5. FourSquare
6. Windows Live

leleuj:
Adding more clients would be great. But I think we need a better way to be able to configure almost anything like a `LdapAuthenticator`. My best thought so far was to define the configuration as a Groovy script (greatly inspired by what you did in CAS ;-)

mmoayyed:
Help me understand this better please: how do you envision a groovy script doing client registration?

leleuj:
Sure. Currently, we have the `ConfigFactory` interface to implement to build a `Config`:  https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/config/ConfigFactory.java This is used for the J2E implementation.

So the idea would be to create some Groovy class extending the `ConfigFactory` to build the `Config` and indicate the path to this Groovy class. That said, I was thinking of Groovy as it feels more dynamic, but it could maybe be in Java or in Javascript using the capabilities of the JVM.

mmoayyed:
OK, I think I understand. Few concerns there:

- A groovy implementation of the config factory, as you noted, may point to some custom groovy script that builds and loads clients. Problem with that approach (regardless of whether it's groovy or JS, etc) is that someone has to write that script! I wouldn't want that, ever :) I'd much rather we ship "preconfigured" clients that folks can easily configure with key=value type of settings.

- Let developers write the groovy implementation and everything that follows afterwards. 

- Doing this via properties and such has the advantage of leveraging spring boot/cloud to proxy those beans and make them dynamic. A groovy/JS script basically amounts to the same thing, but on a more complicated level.

- If as we expect developers to write a groovy script that is loaded by the groovy ConfigFactory, why wouldnt they just write the thing in Java to begin with?! Specially if your target audience are developers with the know-how, why introduce a level of indirection for folks who are expected to use APIs directly?

leleuj:
OK. I understand. So how do we handle advanced use cases with properties? Like setting a bean as the property of another bean. I think we would be very close to the INI configuration system of Shiro, wouldn't we?

mmoayyed:
On a case by case works best I think. Two approaches:

[...]

PR #571, opened 2016-05-05
Added more clients to PropertiesConfigFactory

mmoayyed:
Closes https://github.com/pac4j/pac4j/issues/565


leleuj:
OK. Let's merge this change, but I'm thinking more and more that we should start building something more generic: "github.id" means I must build the GithubClient object and set (by reflection) the id property...

PR #572, opened 2016-05-07
Added a configurable response RequiresHttpAction

garpinc:
Different error code should be passed through depending on what happened

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
@@ -92,6 +92,19 @@ public class RequiresHttpAction extends Exception {
     }
     
     /**
+     * Build a configurable response..
+     * 
+     * @param message message
+     * @param errorCode status code of response
+     * @param context context
+     * @return a configurable response
+     */
+    public static RequiresHttpAction configurable(final String message, int errorCode, final WebContext context) {
+        context.setResponseStatus(errorCode);
+        return new RequiresHttpAction(message, errorCode);
+    }
+    
+    /**
      * Build a forbidden response.
      * 
      * @param message message
--- a/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
@@ -35,7 +35,20 @@ public class RequiresHttpAction extends Exception {
         super(message);
         this.code = code;
     }
-    
+
+    /**
+     * Build a response with message and status.
+     *
+     * @param message message
+     * @param status the HTTP status
+     * @param context context
+     * @return an HTTP response
+     */
+    public static RequiresHttpAction status(final String message, final int status, final WebContext context) {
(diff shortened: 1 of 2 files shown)


pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
```diff
@@ -92,6 +92,19 @@ public static RequiresHttpAction unauthorized(final String message, final WebCon
...
     
     /**
+     * Build a configurable response..
+     * 
+     * @param message message
+     * @param errorCode status code of response
+     * @param context context
+     * @return a configurable response
+     */
+    public static RequiresHttpAction configurable(final String message, int errorCode, final WebContext context) {
```
leleuj: This already exists in 1.9: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java#L32

pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java
```diff
@@ -124,7 +124,7 @@ private RequiresHttpAction handleInvalidCredentials(final WebContext context, fi
 		// it's an AJAX request -> unauthorized (instead of a redirection)
 		if (getAjaxRequestResolver().isAjax(context)) {
 			logger.info("AJAX request detected -> returning " + errorCode);
-			return RequiresHttpAction.unauthorized("AJAX request -> " + errorCode, context, null);
+			return RequiresHttpAction.configurable("AJAX request -> " + errorCode, errorCode, context);
```
leleuj: I think it's clear to have an explicit `unauthorized` here.

leleuj: And I think the `handleInvalidCredentials` should no longer have an `errorCode` parameter.

garpinc: You want to be able to differentiate between CredentialsException and and extract of the credentials being null. i.e 403 vs 401..

leleuj: OK. I get it.

pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
leleuj: Can you copy/paste the `status` method from the master instead of this `configurable` one?

garpinc: I did already

leleuj:
Good

garpinc:
Cool when will this make it to maven central?

leleuj:
Snapshots are available in the Sonatype snapshots repository...

garpinc:
I understand about the snapshots but when are u planning the 1.8.9 release?

leleuj:
No planned release date yet, it can be anytime we want. The latest commit is from 3 days, I propose to wait till next week before I'll do the release. Ping me if you don't see it coming.

PR #573, opened 2016-05-07
Added a configurable response RequiresHttpAction

garpinc:
Pass through error code to differential ajax error

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
@@ -108,19 +108,6 @@ public class RequiresHttpAction extends Exception {
     }
     
     /**
-     * Build a configurable response..
-     * 
-     * @param message message
-     * @param errorCode status code of response
-     * @param context context
-     * @return a configurable response
-     */
-    public static RequiresHttpAction configurable(final String message, int errorCode, final WebContext context) {
-        context.setResponseStatus(errorCode);
-        return new RequiresHttpAction(message, errorCode);
-    }
-    
-    /**
      * Build a forbidden response.
      * 
      * @param message message
--- a/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
@@ -108,6 +108,19 @@ public class RequiresHttpAction extends Exception {
     }
     
     /**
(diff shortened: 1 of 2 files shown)


pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
```diff
@@ -108,6 +108,19 @@ public static RequiresHttpAction unauthorizedDigest(final String message, final
...
     
     /**
+     * Build a configurable response..
+     * 
+     * @param message message
+     * @param errorCode status code of response
+     * @param context context
+     * @return a configurable response
+     */
+    public static RequiresHttpAction configurable(final String message, int errorCode, final WebContext context) {
```
leleuj: This already exists: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java#L32

pac4j-http/src/main/java/org/pac4j/http/client/indirect/FormClient.java
```diff
@@ -94,7 +94,7 @@ private RequiresHttpAction handleInvalidCredentials(final WebContext context, fi
 		// it's an AJAX request -> unauthorized (instead of a redirection)
 		if (getAjaxRequestResolver().isAjax(context)) {
 			logger.info("AJAX request detected -> returning " + errorCode);
-			return RequiresHttpAction.unauthorized("AJAX request -> " + errorCode, context, null);
+			return RequiresHttpAction.configurable("AJAX request -> " + errorCode, errorCode, context);
```
leleuj: I think it's clear to have an explicit `unauthorized` here. And I think the `handleInvalidCredentials` method should no longer have an errorCode parameter.

garpinc: you want do be able to distinguish between credentials being null and invalid credentials.. i.e 401 vs 403

leleuj: OK

pac4j-core/src/main/java/org/pac4j/core/exception/RequiresHttpAction.java
leleuj: Please remove this one.

garpinc: I did already

PR #575, opened 2016-05-09
Upgrade oauth-oidc-sdk to 5.8

papegaaij:
This pull request upgrades com.nimbusds:oauth2-oidc-sdk to version 5.8.

In this version, the old ResourceRetriever is deprecated and replaced by a new one. Also, IDTokenValidator can now take a ResourceRetriever in its constructor, allowing it to follow the timeouts set in the client.

I've also updated the AzureADClient's javadoc to reflect the fact that v2 is no longer in preview.

### DIFF
--- a/pac4j-oidc/pom.xml
+++ b/pac4j-oidc/pom.xml
@@ -13,7 +13,7 @@
 	<name>pac4j for OpenID Connect protocol</name>
 
 	<properties>
-		<oauth-oidc-sdk.version>5.5</oauth-oidc-sdk.version>
+		<oauth-oidc-sdk.version>5.8</oauth-oidc-sdk.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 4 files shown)


leleuj:
Thanks for your contribution.

leleuj:
@papegaaij : did you test the upgrade? I have a strange issue with the OpenID Connect support now: for Google, the `client_secret` is added to the authorize endpoint (`https://accounts.google.com/o/oauth2/v2/auth`), which should not happen and generates a 400 error page...

papegaaij:
Yes, I did test it, but not with Google, only with Azure AD and our own OpenID Connect implementations. I'll see what I can find.

papegaaij:
I think I've found the problem: OidcClient:249 adds the secret to the authParams, which get added to the AuthenticationRequest since 5.8:

```
version` 5.8 (2016-04-21)
  * Adds support for custom OAuth 2.0 authorization request parameters.
  * Adds support for custom OpenID Connect authentication request parameters.
```
I think the fix should be as easy as removing line 249.

papegaaij:
I've checked our own applications, and we have the same issue with Azure AD: the client_secret is rendered in the public facing authorization request. Removing 'client_secret' from authParams fixes the issue.

leleuj:
Can you submit a PR for that? Thanks

papegaaij:
I'll submit a PR monday.

PR #577, opened 2016-05-13
request.getCookies() maybe null according to spec

garpinc:


pac4j-core/src/main/java/org/pac4j/core/context/J2EContext.java
```diff
@@ -198,8 +198,7 @@ public String getFullRequestURL() {
 
     @Override
     public Collection<Cookie> getRequestCookies() {
-        final javax.servlet.http.Cookie[] cookies = this.request.getCookies();
-        final Collection<Cookie> pac4jCookies = new LinkedHashSet<>(cookies.length);
+        final Collection<Cookie> pac4jCookies = new LinkedHashSet<>();
         for (javax.servlet.http.Cookie c : this.request.getCookies()) {
```
mmoayyed: If request.getCookies() can be null, how is that this line is working correctly?! Shouldnt we fix this too?

mmoayyed: ` for (javax.servlet.http.Cookie c : this.request.getCookies()) {` to exact.

garpinc: agreed.. with send a fix

Issue #454, opened 2016-01-21
Handle RSA and Eliptic curve signing for JSON Web Tokens


leleuj:
I propose the following implementation:
- a `JwtSignature` interface with the `getSigner()` and `getVerifier()` methods
- a `HMACSignature` for HMAC signature
- a `ECDSASignature`  for Eliptic Curve signature
- a `RSASignature` for RSA signature

`*Signature` classes being used to build the `JwtGenerator` and `JwtAuthenticator`.

garpinc:
Oh no.. I just ran into this error: Unsupported JWS algorithm ES256, must be HS256, HS384 or HS512

How soon before it is implemented?

leleuj:
Planned for 1.9.1 so after the 1.9 release for which I must first finish: https://github.com/pac4j/pac4j/issues/558. That said, it could perfectly go into 1.9.

There are some other issues related to JWT support as well: https://github.com/pac4j/pac4j/issues/567, https://github.com/pac4j/pac4j/issues/568, https://github.com/pac4j/pac4j/issues/569 and https://github.com/pac4j/pac4j/issues/570.

The signers are different for the different family of encryption methods (ES, HS). The nimbusds jose library already supports all, but the pac4j JWT support must be changed to take that into account, which is feasible with various configurations (which I called `Signature` in my proposal above).

This is something you could perfectly work on and contribute to pac4j, if you find some free time.

leleuj:
PR: https://github.com/pac4j/pac4j/pull/621

Signature and encryption configurations are now available: https://github.com/leleuj/pac4j/tree/pac4j-619/pac4j-jwt/src/main/java/org/pac4j/jwt/config

PR #579, opened 2016-05-13
Use underlying library to decide which verifier to use based on header.

garpinc:
#454 On the authenticator side I think the better approach is to use the underlying api to decide which verifier to use based on the header. On the generator side you are in control so I don't think you need an abstraction. Now except for the original MacVerifier I'm not sure how to create a Key but I think that's up to whoever injects the key into the authenticator.. I preserved existing functionality allowing secret key to be converted to SecretKey.


garpinc:
Opened another pull request that allows keys to be specified in PEM format

PR #580, opened 2016-05-13
Fix osgi classloading issues

garpinc:


leleuj:
Did you test it also in a non-OSGI environment?

garpinc:
No I don't have that environment but wouldn't test cases already cover that?

leleuj:
I hope so but classloader issues can be tricky. I'll give it a try later on...

PR #583, opened 2016-05-18
Use underlying library to decide which verifier to use based on header.

garpinc:
Allow secret to be a PEM with associated algorithm
#454

### DIFF
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
@@ -215,7 +215,7 @@ public class JwtAuthenticator implements TokenAuthenticator {
     	// from http://stackoverflow.com/questions/35739932/how-do-i-decode-a-jwt-token-using-an-rsa-public-key-in-pem-format
     	// decode to its constituent bytes
     	String publicKeyPEMToUse = publicKeyPEM;
-    	publicKeyPEMToUse = publicKeyPEMToUse.replace("-----BEGIN PUBLIC KEY-----\n", "");
+    	publicKeyPEMToUse = publicKeyPEMToUse.replaceAll("-----BEGIN PUBLIC KEY-----[\\r\\n]+", "");
     	publicKeyPEMToUse = publicKeyPEMToUse.replace("-----END PUBLIC KEY-----", "");
     	byte[] publicKeyBytes = javax.xml.bind.DatatypeConverter.parseBase64Binary(publicKeyPEMToUse);
 
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
@@ -55,6 +55,9 @@ import javax.crypto.spec.SecretKeySpec;
 /**
  * Authenticator for JWT. It creates the user profile and stores it in the credentials
  * for the {@link org.pac4j.http.profile.creator.AuthenticatorProfileCreator}.
+ * 
+ *  If the secret is a pem you need to specify the algorithm which will create the right Key object for underlying validator. 
+ *  In addition you can create the key object of the right type yourself and pass it in.
  *
  * @author Jerome Leleu
  * @since 1.8.0
@@ -206,7 +209,7 @@ public class JwtAuthenticator implements TokenAuthenticator {
     /**
      * @throws NoSuchAlgorithmException 
      * @throws InvalidKeySpecException 
-     * @since 1.8.2
+     * @since 1.8.9
      */
     public void setSigningPem(final String publicKeyPEM, final String algorithm) throws NoSuchAlgorithmException, InvalidKeySpecException {
     	// from http://stackoverflow.com/questions/35739932/how-do-i-decode-a-jwt-token-using-an-rsa-public-key-in-pem-format
@@ -243,10 +246,16 @@ public class JwtAuthenticator implements TokenAuthenticator {
         warning();
     }
 
+    /**
+     * @since 1.8.9
+     */
 	public Key getKey() {
 		return key;
 	}
 
+    /**
+     * @since 1.8.9
+     */
 	public void setKey(Key key) {
 		this.key = key;
 	}
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
@@ -16,8 +16,11 @@
 package org.pac4j.jwt.credentials.authenticator;
 
 import com.nimbusds.jose.JWEObject;
+import com.nimbusds.jose.JWSVerifier;
 import com.nimbusds.jose.crypto.DirectDecrypter;
-import com.nimbusds.jose.crypto.MACVerifier;
+import com.nimbusds.jose.crypto.factories.DefaultJWSVerifierFactory;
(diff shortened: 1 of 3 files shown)


garpinc:
#454 On the authenticator side I think the better approach is to use the underlying api to decide which verifier to use based on the header. On the generator side you are in control so I don't think you need an abstraction. Now except for the original MacVerifier I'm not sure how to create a Key but I think that's up to whoever injects the key into the authenticator.. I preserved existing functionality allowing secret key to be converted to SecretKey.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -160,14 +188,38 @@ private static void createJwtProfile(final TokenCredentials credentials, final S
...
+    		this.key = new SecretKeySpec(signingSecret.getBytes(Charset.forName("UTF-8")), "AES");
+    	} else {
+    		this.key = cert.getPublicKey();
+    	}
+    }
+    
+    /**
+     * @throws NoSuchAlgorithmException 
+     * @throws InvalidKeySpecException 
+     * @since 1.8.2
```
leleuj: Not exactly. 1.8.9 I suppose...

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -70,7 +84,17 @@ private void warning() {
      * @since 1.8.2
      */
     public JwtAuthenticator(final String signingSecret, final String encryptionSecret) {
-        this.signingSecret = signingSecret;
+    	setSigningSecret(signingSecret);
```
leleuj: The possible values of the `signingSecret` should be documented somewhere.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
garpinc: Not following you here..

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -114,8 +138,10 @@ public void validate(final TokenCredentials credentials) {
             } else {
                 throw new TechnicalException("unsupported unsecured jwt");
             }
+            
+    		JWSVerifier verifier = factory.createJWSVerifier(signedJWT.getHeader(), key);
```
leleuj: It's a good idea indeed to create the verifier from the signedJWT header, the problem is that the key must be of the right format.

leleuj:
Although this solution works, I think that the `setSigningSecret` and `setSigningPem` methods are a bit too specific. I'd like to see a PR which considers the `JwtGenerator` as well.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
garpinc: I think that's documented mostly in the underlying libraries I think..

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
garpinc: A gave the possibility to pass in a Key (i.e: be explicit)

garpinc:
As I said.. A person has the option to generate what kind of JWT key they like with whatever library they like. I'm not really clear on the purpose of the JwtGenerator since i don't myself use it.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
leleuj: I was expected the `@since 1.8.9` javadoc as it's a new method and we are at version 1.8.9

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
leleuj: Certainly, but we must help pac4j adopters to use the library: so there must be somewhere (in the Javadoc header class or in this method Javadoc) a explanation and example. So we can use simple strings as we use to but also the value of a certificate, right?

leleuj:
Most pac4j deployers use the `JwtGenerator` to generate JWT tokens from user profiles and the `JwtAuthenticator` to validate them, that's why the PR must target both.

Now the `JwtAuthenticator`, the appropriate signer is created assuming you have provided the right key as well, but the `JwtGenerator` is not at the same level as you cannot specify a verifier, nor some specific keys.

That's why I think there must be a configuration component which can be used by both.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
garpinc: Yes. If the string is a pem you need to specify which algorithm which will create the right Key object. In addition you can create the key object of the right type yourself

garpinc:
The underlying api has the ablility to select the verifiers based on the header. It does not have a mechanism to do the generation based on config hence the reason why it's one sided. What you are suggestion is that we build a config driven generator based on all the algorithms the underlying library supports. This in my opinion is too brittle as this will get out of sync with the underlying library and eventually if they end user wants a specific algorithm they will bypass or override the generator anyway. I personally don't see the point in creating such a component and why the framework even implements the generator side as it's more like a utility rather than a framework thing. It doesn't change anything about the authentication solution that this framework is.

leleuj:
OK. Choosing the appropriate verifier in `JwtAuthenticator` is a good thing. Though, we cannot escape the configuration for the `JwtGenerator` and the configuration of the different key types.

That said, this PR is a first improvement towards the right direction. So let's keep it that way. Nonetheless, can you add a test with a PEM file before the merge? Thanks

leleuj:
@garpinc : I plan to release pac4j v1.9 next wednesday: can you add a test with a PEM file so I can merge this PR and the one related to master before?

garpinc:
Yes I'll try to get to it before then..

garpinc:
How is the release plan for 1.8.9?

leleuj:
No plan for 1.8.9 (my efforts are now focused on v1.9 adoption), but I can do that in the next few days

leleuj:
I just released pac4j v1.8.9, it should be available in the Maven central repository in a few hours...

garpinc:
But wait.. you didn't merge this pull request..

leleuj:
Too late. I'll release a 1.8.10 if necessary unless you upgrade to 1.9

garpinc:
I think it will require too much testing to upgrade to 1.9.. We're also in process of a release here and I don't want to rock the boat and introduce a lot of regression testing. How soon would I get 1.8.10 with this merged change?

leleuj:
After the merge, I can release a 1.8.10 and it will go in the 1.9.1-SNAPSHOT as well.

What is left to be done is to add a test to confirm the PEM support.

garpinc:
But I already added the trst

garpinc:
But I already added the test. Maybe just to  master though

leleuj:
Indeed, I missed it as you just added it to the master. Can you add it also to the 1.8.x branch?

leleuj:
And then, I'll release the 1.8.10 version.

leleuj:
Looks good to me. Will merge

PR #584, opened 2016-05-18
Use underlying library to decide which verifier to use based on header

garpinc:
#454 On the authenticator side I think the better approach is to use the
underlying api to decide which verifier to use based on the header. On
the generator side you are in control so I don't think you need an
abstraction. Now except for the original MacVerifier I'm not sure how to
create a Key but I think that's up to whoever injects the key into the
authenticator.. I preserved existing functionality allowing secret key
to be converted to SecretKey.

Allow secret to be a PEM with associated algorithm


leleuj:
I just merged: https://github.com/pac4j/pac4j/pull/591 which you must sync with...

garpinc:
Also don't forget to merge this pull request before 1.9 release

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
```diff
@@ -114,9 +126,15 @@ protected String encryptJwt(final SignedJWT signedJWT) throws Exception {
     }
 
     protected SignedJWT signJwt(final JWTClaimsSet claims) throws JOSEException {
```
leleuj: This method is useless: it's not call from any other method, isn't it?

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
```diff
@@ -67,22 +67,34 @@ public JwtGenerator(final String signingSecret, final String encryptionSecret) {
...
-
+    
     /**
      * Generates a JWT from a user profile.
      *
      * @param profile the given user profile
      * @return the created JWT
      */
     public String generate(final U profile) {
+    	return generate(profile, null, this.jwsAlgorithm);
```
leleuj: I see a backward compatibility issue here (compared to the version 1.8): previously, no `signingSecret` means no signing. Now, we don't sign JWT **by default** and this is a problem.

So in this method, you need to check if the `signingSecret` is blank and if so, pass `null`, otherwise, pass a `new MACSigner(this.signingSecret);`

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
garpinc: No.. But I kept it for backwards compatibility since it's protected.. I can remove it if you'd like. Please confirm.

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
garpinc: It says
+ if (signer == null) {
 +        	// Create HMAC signer
 +        	signer = new MACSigner(this.signingSecret);
 +        }

So I don't see how it's a backward compatibility issue

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
leleuj: OK. I see. Please remove it.

pac4j-jwt/src/main/java/org/pac4j/jwt/profile/JwtGenerator.java
leleuj: Indeed, in the `signJwt` method (line 133), the test is correct. Though, the problem is that this method will never be called because of the test line 94: you should not test if the `signer` is `null`, but if the `signingSecret` is not blank.

PR #586, opened 2016-05-24
Make method public so client outside pac4j can instantiate classes of their own

garpinc:
Make method public so client outside pac4j can instantiate classes of their own


leleuj:
What's your use case for building your profile on your own?

garpinc:
My use case is for Kerberos which so far you guys have indicated that you don't want. (which I verified works last night btw). I will still submit a pull request if you want but so far it's outside pac4j framework in it's own package.

leleuj:
OK. Let's do that.

PR #587, opened 2016-05-24
Make method public so client outside pac4j can instantiate classes of their own

garpinc:
Make method public so client outside pac4j can instantiate classes of their own

### DIFF
--- a/pac4j-http/src/test/java/org/pac4j/http/profile/TestCustomProfileHelper.java
+++ b/pac4j-http/src/test/java/org/pac4j/http/profile/TestCustomProfileHelper.java
@@ -1,0 +1,74 @@
+/*
+  Copyright 2012 - 2015 pac4j organization
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


pac4j-http/src/test/java/org/pac4j/http/profile/TestCustomProfileHelper.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.pac4j.core.profile.CommonProfile;
+import org.pac4j.core.profile.ProfileHelper;
+import org.pac4j.core.profile.TestCaseProfileHelper;
+import org.pac4j.core.profile.UserProfile;
+
+/**
+ * This class tests the {@link ProfileHelper} class for the {@link HttpProfile}.
```
leleuj: Feels like a copy/paste which should be fixed.

pac4j-http/src/test/java/org/pac4j/http/profile/TestCustomProfileHelper.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.pac4j.core.profile.CommonProfile;
+import org.pac4j.core.profile.ProfileHelper;
+import org.pac4j.core.profile.TestCaseProfileHelper;
+import org.pac4j.core.profile.UserProfile;
+
+/**
+ * This class tests the {@link ProfileHelper} class for a custom {@link CustomProfileClass}.
+ * 
+ * @author Jerome Leleu
```
leleuj: You should put your name here...

PR #588, opened 2016-05-30
Do not put 'client_secret' in authentication request

papegaaij:
As discussed in #575, the 'client_secret' must not be rendered in the authentication request.


leleuj:
Thanks

Issue #54, opened 2014-05-14
Add logout redirect mecanism


patou:
In the Google Apps UserService, the logout mecanisme create an logout url, which allow to go to a google url for logout from google server and after redirect to the application.

https://developers.google.com/appengine/docs/java/users/#Java_Signing_in_and_out :  

```
The Users API includes a method to generate a URL for signing out of the app. 
The  sign-out URL de-authenticates the user from the app, 
then redirects back to the app's   URL without displaying anything.
```

leleuj:
New implementations based on pac4j v1.8 have an "application logout filter" which performs a local logout (the current user profile is removed from the session).
The idea here would be to perform a server logout after a local logout or not (depending on the fact that the server logout will call or not the application logout). It would depend on the current authenticated profile. So I imagine a new method in the `Client` like `logout(P profile, WebContext context);` and a new "logout filter" which would get the current authenticated profile, loop over all clients and call each logout method (unless we can directly link the profile to the client) to see if there is an url to redirect to.

mavencode01:
@leleuj How can I perform a server logout after a local logout ? I'm experiencing this problem in the vertx wrapper https://github.com/pac4j/vertx-pac4j-demo/issues/37

millross:
Possible to do this via a redirect action in shared logout logic @leleuj ? As you can see @mavencode01 is keen to see this too (and this is at the heart of the issue he reported on the play-pac4j impl as well as the vertx-pac4j impl). I'd like to limit the code impact on implementations where possible, so nicer to pass in all user profiles and let the clients figure out which belong to them and use them to perform server log out (if requested) I think.

PR #589, opened 2016-05-31
Session index added to SAML2Credentials

taggiasco:
Session index is taken from SAML2 response. It is added to SAML2Credentials and then to the SAML2Profile as an attribute (however not sure it is a good choice)

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
@@ -179,7 +179,9 @@ public class SAML2DefaultResponseValidator implements SAML2ResponseValidator {
         List<AuthnStatement> authnStatements = subjectAssertion.getAuthnStatements();
         if(authnStatements != null && authnStatements.size() > 0) {
         	AuthnStatement statement = authnStatements.get(0);
-        	return statement.getSessionIndex();
+        	if(statement != null) {
+        		return statement.getSessionIndex();
+        	}
         }
         return null;
     }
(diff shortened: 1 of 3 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
```diff
@@ -164,7 +166,22 @@ protected final SAML2Credentials buildSAML2Credentials(final SAML2MessageContext
...
+     * Searches the sessionIndex in the assertion
+     * 
+     * @param subjectAssertion assertion from the response
+     * @return the sessionIndex if found in the assertion
+     */
+    private final String getSessionIndex(Assertion subjectAssertion) {
+        List<AuthnStatement> authnStatements = subjectAssertion.getAuthnStatements();
+        if(authnStatements != null && authnStatements.size() > 0) {
+        	AuthnStatement statement = authnStatements.get(0);
+        	return statement.getSessionIndex();
```
leleuj: Can `statement` be null?

leleuj:
What do you want to do with the `sessionIndex`?

taggiasco:
sessionIndex is mandatory to handle Single Logout.
1) when logout is performed on "your" SP, you should redirect to SAML IdP with that sessionIndex.
2) when logout is performed on another SP, IdP should ask "your" SP for logout with that sessionIndex too

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidator.java
taggiasco: The method `validateAuthenticationStatements` does not take into account the fact that a statement could be null (direct call to `statement.getAuthnInstant()`).

However I'm not sure about it. I'll update the PR.

leleuj:
I had in mind that this change targeted the single logout. In fact, the single logout should be handled in a more generic way, as briefly explained in this issue: https://github.com/pac4j/pac4j/issues/54, but this can be a first temporary step.

PR #596, opened 2016-06-14
Kerberos Implementation backport

garpinc:


leleuj:
There won't be any more evolution on 1.8.x since 1.9.0 has been released (only bug fixes).

PR #597, opened 2016-06-14
An interface that provides a way for a direct client to request

garpinc:
An interface that provides a way for a direct client to request additional information from browser. Initially, the kerberos client should implement this


garpinc:
Not currently required

PR #607, opened 2016-07-16
Add support for single profile instances

### LINKED ISSUE #606: ProfileManager should support single profile instances from the session
The ProfileManager instance does this

```
    private LinkedHashMap<String, U> retrieveAll(final boolean readFromSession) {
        LinkedHashMap<String, U> profiles = new LinkedHashMap<>();
        final Object objSession = this.context.getRequestAttribute(Pac4jConstants.USER_PROFILES);
        if (objSession != null && objSession instanceof LinkedHashMap) {
            profiles = (LinkedHashMap<String, U>) objSession;
        }
        if (readFromSession) {
            final Object objRequest = this.context.getSessionAttribute(Pac4jConstants.USER_PROFILES);
            if (objRequest != null && objRequest instanceof LinkedHashMap) {
                profiles.putAll((LinkedHashMap<String, U>) objRequest);
            }
        }
        return profiles;
    }
```

It really should do this instead:

```
    private LinkedHashMap<String, U> retrieveAll(final boolean readFromSession) {
        LinkedHashMap<String, U> profiles = new LinkedHashMap<>();
        final Object objRequest = this.context.getRequestAttribute(Pac4jConstants.USER_PROFILES);
        if (objRequest != null) {
            if (objRequest instanceof LinkedHashMap) {
                profiles = (LinkedHashMap<String, U>) objRequest;
            }
            if (objRequest instanceof UserProfile) {
                profiles.put(...);
            }
        }
        if (readFromSession) {
            final Object objSession = this.context.getSessionAttribute(Pac4jConstants.USER_PROFILES);
            if (objSession instanceof LinkedHashMap) {
                profiles = (LinkedHashMap<String, U>) objSession;
            }
            if (objSession instanceof UserProfile) {
                profiles.put(...);
            }
        }
        return profiles;
    }
```

This should be done to ease configuration of the API, and to make sure 1.9 is backward compatible.

mmoayyed:
Closes #606

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
@@ -54,15 +54,23 @@ public class ProfileManager<U extends CommonProfile> {
      * @return the map of profiles
      */
     private LinkedHashMap<String, U> retrieveAll(final boolean readFromSession) {
-        LinkedHashMap<String, U> profiles = new LinkedHashMap<>();
-        final Object objSession = this.context.getRequestAttribute(Pac4jConstants.USER_PROFILES);
-        if (objSession != null && objSession instanceof LinkedHashMap) {
-            profiles = (LinkedHashMap<String, U>) objSession;
+        final LinkedHashMap<String, U> profiles = new LinkedHashMap<>();
+        final Object request = this.context.getRequestAttribute(Pac4jConstants.USER_PROFILES);
+        if (request != null) {
+            if  (request instanceof LinkedHashMap) {
+                profiles.putAll((LinkedHashMap<String, U>) request);
+            }
+            if (request instanceof CommonProfile) {
+                profiles.put(((CommonProfile) request).getClientName(), (U) request);
+            }
         }
         if (readFromSession) {
-            final Object objRequest = this.context.getSessionAttribute(Pac4jConstants.USER_PROFILES);
-            if (objRequest != null && objRequest instanceof LinkedHashMap) {
-                profiles.putAll((LinkedHashMap<String, U>) objRequest);
+            final Object sessionAttribute = this.context.getSessionAttribute(Pac4jConstants.USER_PROFILES);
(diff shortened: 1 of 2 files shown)


pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
```diff
@@ -54,15 +54,23 @@ public ProfileManager(final WebContext context) {
...
-        if (objSession != null && objSession instanceof LinkedHashMap) {
-            profiles = (LinkedHashMap<String, U>) objSession;
+        final LinkedHashMap<String, U> profiles = new LinkedHashMap<>();
+        final Object request = this.context.getRequestAttribute(Pac4jConstants.USER_PROFILES);
+        if (request != null) {
+            if  (request instanceof LinkedHashMap) {
+                profiles.putAll((LinkedHashMap<String, U>) request);
+            }
+            if (request instanceof CommonProfile) {
+                profiles.put(((CommonProfile) request).getClientName(), (U) request);
```
leleuj: In fact, the `clientName` method did not exist before version 1.9 so this will return `null`. We won't be able to infer the client name so we could use the profile id or some `DEFAULT` value.

leleuj:
+1 but see my comment

The `ProfileManager` is a very sensitive component. A test for the new use case would be welcome.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
mmoayyed: I am confused. This is targeted to 1.9.1 right? Why would the call ever return `null`?

mmoayyed: Ah IC. You mean something like this:

```
String clientName = profile.getClientName();
        if (clientName == null) {
            clientName = "DEFAULT";
        }
```

mmoayyed: PS This really seems like a superfluous check way the way. Client name should never be null.

PR #610, opened 2016-07-16
Added ability to make authnRequestSigned configurable

thinh76:


leleuj:
+1

@mmoayyed : ok for you?

mmoayyed:
Yes this is fine. @thinh76 You really ought to fix the indentation though. See the diff.

thinh76:
Sorry @mmoayyed, I wanted to... but I saw indentation are different in a lot of files, and I thought we should have a format.xml or something so each contributors can import and apply the same styles... when I did the format using IDEA, it caused more differences than without formatting. BTW, do you still wait for me to submit format changes?

leleuj:
Not a big deal, I will merge "as is" to cut the release...

PR #620, opened 2016-08-02
Add DirectCasClient and DirectCasProxyClient

leleuj:
This PR needs some more testing on proxy and logout.


pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
```diff
@@ -1,388 +1,204 @@
...
-        this.ticketValidator = new Cas20ServiceTicketValidator(this.casPrefixUrl);
-        final Cas20ServiceTicketValidator cas20ServiceTicketValidator = (Cas20ServiceTicketValidator) this.ticketValidator;
-        cas20ServiceTicketValidator.setEncoding(this.encoding);
-        if (this.casProxyReceptor != null) {
-            cas20ServiceTicketValidator.setProxyCallbackUrl(this.casProxyReceptor.computeFinalCallbackUrl(context));
-            cas20ServiceTicketValidator.setProxyGrantingTicketStorage(this.casProxyReceptor
-                    .getProxyGrantingTicketStorage());
-        }
-    }
+        configuration.setCallbackUrlResolver(this.getCallbackUrlResolver());
```
leleuj: @mmoayyed @millross the `CasClient` has been broken into small subcomponents and it's really a good idea: makes things clearer and reusable (Client -> ClientV2).

pac4j-cas/src/main/java/org/pac4j/cas/client/direct/DirectCasClient.java
```diff
@@ -0,0 +1,115 @@
...
+ * <p>As no session is meant to be created, this client does not handle CAS logout requests.</p>
+ *
+ * <p>For proxy support, a {@link CasProxyReceptor} must be defined in the configuration (the corresponding "callback filter" must be enabled)
+ * and set to the CAS configuration of this client. In that case, a {@link org.pac4j.cas.profile.CasProxyProfile} will be return
+ * (instead of a {@link org.pac4j.cas.profile.CasProfile}) to be able to request proxy tickets.</p>
+ *
+ * @author Jerome Leleu
+ * @since 1.9.2
+ */
+public class DirectCasClient extends DirectClientV2<TokenCredentials, CasProfile> {
```
leleuj: @mmoayyed This is what you need to have a CAS stateless authentication -> https://github.com/apereo/cas/issues/1571

pac4j-cas/src/main/java/org/pac4j/cas/credentials/extractor/TicketAndLogoutRequestExtractor.java
```diff
@@ -0,0 +1,170 @@
...
+
+            final String ticket = CommonHelper.substringBetween(logoutMessage, CasConfiguration.SESSION_INDEX_TAG + ">", "</");
+            if (CommonUtils.isNotBlank(ticket)) {
+                configuration.getLogoutHandler().destroySessionBack(context, ticket);
+            }
+            final String message = "back logout request: no credential returned";
+            logger.debug(message);
+            throw HttpAction.ok(message, context);
+
+        } else if (isFrontLogoutRequest(context)) {
```
leleuj: This now handles front channel logout as well.

mmoayyed: Are there any docs or samples on how to handle front-channel logout with CAS? How are those requests sent that they are now handled here? I can't recall ever seeing folks using front-channel slo.

leleuj: You have some basic information about front channel logout here: https://apereo.github.io/cas/4.2.x/installation/Logout-Single-Signout.html

You're right: I don't think anyone uses it actually, but it certainly comes from the fragile design. I think it would make more sense to send multiple logout requests (let's say by hidden images on the logout page) with the message encrypted and base64 instead of relying on the multiple round-trips to perform the front channel logout. Maybe one of the last things remaining for CAS ;-)

That's why I developed the upgrade of the Apereo CAS client with the option of ignoring the RelayState and I did the same for pac4j.

PR #629, opened 2016-08-15
Enable customisation of profile manager creation without inheriting security logic class

millross:
This is a PR to show how to enable customization of the profile manager creation without having to extend this class, by offering injection of a function for creating a profile manager for a given WebContext.

If this is accepted I'll submit analogous PRs for other components which need the same behaviour (e.g. callback logic).

The advantage of this approach is that (assuming Java 8) it enables customization of ProfileManager creation without having to extend thse classes, permitting particular implementations of Pac4j on different platforms to customize their profile management without hacking the webcontext behaviour or having to inherit classes just to extend one method.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
@@ -24,6 +24,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.List;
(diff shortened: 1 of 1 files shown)


leleuj:
Excellent. Smart way to address the customisation.

leleuj:
Go for the callback logic...

millross:
Will do, may not be tonight but should be by the weekend. I should have thought of that approach earlier, as I think it also makes it easier to extend the vertx-pac4j implementation if someone wants to diverge from vert.x idioms (which I think will be rare).

PR #630, opened 2016-08-15
Generic OAuth2.0 Clients

ajherrick:
What: An OAuth2.0 client that can be used with any OAuth 2.0 service (theoretically). It allows the user to define all of the URLs to make the relevant requests to. In addition, the user can define a custom AttributesDefinition (that extends the new GenericAttributesDefinition) to extract whatever profile details they expect.

Why: My organization uses pac4j for Google SSO in our apps, and we like how simple it makes the process. We also released our own internal OAuth provider and have migrated our Play applications to use that instead of each app having their own login screen to maintain. I created this because we would like to use pac4j with our internal OAuth provider, but needed more flexibility than a hardcoded Api. The purpose for this client is for use with internal OAuth2.0 services.

I also was looking into making BaseOAuth20Client and BaseOAuth20StateClient non-abstract with this functionality, but am limited on time. Perhaps that could be a future change.

Also, I'm not sure about tests for this. Perhaps I could create a test that configures this to authenticate with Google and run the same Google tests?

### DIFF
--- a/pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/DefaultGenericAttributesDefinition.java
+++ b/pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/DefaultGenericAttributesDefinition.java
@@ -9,19 +9,19 @@ import org.pac4j.core.profile.converter.Converters;
  * @author aherrick
  * @since 1.9.2
  */
-public class GenericAttributesDefinition extends AttributesDefinition {
+public class DefaultGenericAttributesDefinition extends AttributesDefinition {
 
-    public final String EMAIL = "email";
-    public final String FIRST_NAME = "first_name";
-    public final String FAMILY_NAME = "family_name";
-    public final String DISPLAY_NAME = "display_name";
-    public final String GENDER = "gender";
-    public final String LOCALE = "locale";
-    public final String PICTURE_URL = "picture_url";
-    public final String PROFILE_URL = "profile_url";
-    public final String LOCATION = "location";
+    public static final String EMAIL = "email";
+    public static final String FIRST_NAME = "first_name";
+    public static final String FAMILY_NAME = "family_name";
+    public static final String DISPLAY_NAME = "display_name";
+    public static final String GENDER = "gender";
+    public static final String LOCALE = "locale";
+    public static final String PICTURE_URL = "picture_url";
+    public static final String PROFILE_URL = "profile_url";
+    public static final String LOCATION = "location";
 
-    public GenericAttributesDefinition() {
+    public DefaultGenericAttributesDefinition() {
         primary(EMAIL, Converters.STRING);
         primary(FIRST_NAME, Converters.STRING);
         primary(FAMILY_NAME, Converters.STRING);
--- a/pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20Profile.java
+++ b/pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20Profile.java
@@ -1,10 +1,8 @@
 package org.pac4j.oauth.profile.generic;
 
-import org.pac4j.core.context.Pac4jConstants;
-import org.pac4j.core.profile.Gender;
-import org.pac4j.oauth.profile.OAuth20Profile;
+import org.pac4j.core.profile.AttributesDefinition;
 
-import java.util.Locale;
+import org.pac4j.oauth.profile.OAuth20Profile;
 
 /**
  * <p>This is the user profile for the generic OAuth 2.0 client.</p>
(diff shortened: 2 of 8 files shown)


pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20Profile.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * <p>This is the user profile for the generic OAuth 2.0 client.</p>
+ * <p>It is returned by the {@link org.pac4j.oauth.client.GenericOAuth20Client}  </p>
+ *
+ * @author aherrick
+ * @since 1.9.2
+ */
+public class GenericOAuth20Profile extends OAuth20Profile {
+
+    private transient GenericAttributesDefinition attributes = new GenericAttributesDefinition();
```
leleuj: Could be `static final` as well and named with capitals?

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20Profile.java
```diff
@@ -0,0 +1,89 @@
...
+     * Custom definition must extend {@link org.pac4j.oauth.profile.generic.GenericAttributesDefinition}
+     *
+     * @param attributes
+     */
+    public void setAttributesDefinition(GenericAttributesDefinition attributes) {
+        this.attributes = attributes;
+    }
+
+    @Override
+    public String getEmail() {
```
leleuj: All the following getters are the same as in `CommonProfile`: they are not necessary.

leleuj:
I like the idea. Maybe we'll need to do more, like being able to setup how the access token is sent to query the profile url (as a parameter or header), but this is a good first step for now.

See my comments. Plus you must fix the build, there are some Findbugs issues.

I agree with the test proposal:  create a `RunXXX` class in the `run` package which uses the Generic OAuth client and authenticates with Google.

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20Profile.java
ajherrick: I think it's important to remain non-static, so it can be reassigned. In the client, someone can pass in their own attributes definition that extends GenericAttributesDefinition. This allows them to specify that their service returns FAMILY_NAME as something other than "family_name" ("last_name", for example). I can make it final though, if that's what you want

leleuj: As the property is `private`, I'm not sure to understand how you want to reassign it. Besides, every time you will instantiate a `GenericOAuth20Profile`,  you will create an attributes definition, which is definitely not necessary: you have one only definition for all `GenericOAuth20Profile`.

Of course, for a new specific OAuth provider, you would create your specific profile with a specific attributes definition.

ajherrick: I have a setter method a few lines below that allows it to be changed. My thought process is that you can create a custom attributes definition and pass it in when creating the client. Ex:
`private Class MyAttributes extends GenericAttributesDefinition {
static {
FAMILY_NAME = "last_name"
}
public MyAttributes() {
super()
}
}`

This tells the client that when it retrieves the profile, the FAMILY_NAME can be retrieves using the key "last_name".

leleuj: OK. I see now. Let's keep it that way.

ajherrick: So as it turns out, it doesn't work as well as I thought it would. The Maven build tool won't allow me to set the GenericAttributesDefinition variables to non-final, and I know nothing about Maven, so I haven't figured out how to disable that for that file yet.

Using profile.getAttribute() still allows you to get all the values, but it's less user-friendly

leleuj: I think I would make the `attributesDefinition` `final static` and attached to this profile. If you have a new profile, you'll have a new `attributesDefinition` and the problem is how to define what profile is returned by what client (and it's a more general issue than just for OAuth).

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20Profile.java
```diff
@@ -0,0 +1,31 @@
...
+    public AttributesDefinition getAttributesDefinition() {
+        return attributes;
+    }
+
+    /**
+     * Used to set a custom AttributesDefinition.
+     *
+     * @param attributes
+     */
+    public void setAttributesDefinition(AttributesDefinition attributes) {
```
leleuj: I still think it's a wired thing to have this setter, because the attributes definition is a static data of the profile.

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/DefaultGenericAttributesDefinition.java
```diff
@@ -0,0 +1,35 @@
...
+    public static final String LOCATION = "location";
+
+    public DefaultGenericAttributesDefinition() {
+        primary(EMAIL, Converters.STRING);
+        primary(FIRST_NAME, Converters.STRING);
+        primary(FAMILY_NAME, Converters.STRING);
+        primary(DISPLAY_NAME, Converters.STRING);
+        primary(GENDER, Converters.GENDER);
+        primary(LOCALE, Converters.LOCALE);
+        primary(PICTURE_URL, Converters.URL);
```
leleuj: No, for `PICTURE_URL` and `PROFILE_URL`, it's a `Converters.STRING`.

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/DefaultGenericAttributesDefinition.java
```diff
@@ -0,0 +1,35 @@
...
+import org.pac4j.core.profile.AttributesDefinition;
+import org.pac4j.core.profile.converter.Converters;
+
+/**
+ * Default attribute definition of the GenericOAuth20Profile
+ *
+ * @author aherrick
+ * @since 1.9.2
+ */
+public class DefaultGenericAttributesDefinition extends AttributesDefinition {
```
leleuj: This definition (renamed as `CommonAttributesDefinition`) should be put in the core as it's primarily attached to the `CommonProfile`.

leleuj:
Merging as is. I'll rename the `DefaultGenericAttributesDefinition` and move it in the core later on.

PR #632, opened 2016-08-16
Extend profile manager factory injection to DefaultApplicationLogoutLogic and DefaultCallbackLogic

millross:
This is a PR to extend profile manager factory injection, as carried out for DefaultSecurityLogic, to DefaultApplicationLogoutLogic and DefaultCallbackLogic. The approach offers injection of a function for creating a profile manager for a given WebContext.

The advantage of this approach is that (assuming Java 8) it enables customization of ProfileManager creation without having to extend thse classes, permitting particular implementations of Pac4j on different platforms to customize their profile management without hacking the webcontext behaviour or having to inherit classes just to extend one method, both of which are less desirable. The default behaviour is to create new ProfileManager (as was previously occurring) but it is now much easier to customise these behaviours. I would also suggest that it wouldn't be unreasonable to make the getProfileManager private as overriding the getProfileManager offers nothing that injection of a factory does not. However, this PR does not take that step.


leleuj:
Good. Thanks

PR #637, opened 2016-08-26
remove gfw url

zawn:


leleuj:
What's this change for?

zawn:
Here we can not access these sites, if you are willing to change can best

leleuj:
OK. I see. I guess Google is not welcome either ;-)

PR #638, opened 2016-08-26
complete error message

zawn:


pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
```diff
@@ -32,8 +32,9 @@ public TokenCredentials extract(WebContext context) throws HttpAction {
             return null;
         }
 
-        if  (!header.startsWith(this.prefixHeader)) {
-            throw new CredentialsException("Wrong prefix for header: " + this.headerName);
+        if (!header.startsWith(this.prefixHeader)) {
+            throw new CredentialsException("Wrong prefix for header: \"" + this.headerName + ": " + header
```
leleuj: I'm not against improving this error message, though the `this.headerName` property appears twice in the message, which is not useful.

zawn:
Ok,

Issue #612, opened 2016-07-27
Missing getRequestContent in WebContext


victornoel:
Hi,

I'm trying to write a client by extending DirectClientV2 and I would like to be able to read the content of the request (I'm expecting a POST request with the credentials in a JSON document).

The current implementation of WebContext does not seem to provide a way to access this information.

Thanks :)

leleuj:
Keep in mind that the `WebContext` must be implemented in all frameworks where pac4j is used: it's a very sensitive component. And I'm reluctant to add this method for now.

What the implementation do you use? With the `J2EContext` for example, you have access to the `HttpServletResponse` response directly.

Notice that you should create a `JsonCredentialsExtractor` here.

victornoel:
Ok, I understand, I'm currently writing this `JsonCredentialsExtractor` and that's where I was looking for a generic way of accessing the Web request content.

I know in my case that it will be a `J2EContext`, so I can downcast the context to that, but I was wondering if there wasn't some need for a more generic approach to that (since my extractor won't be usable with anything else than a `J2EContext`).

I see it is not trivial because it means to choose a way to represent that content (a stream, a string, etc…) and force all implementors of WebContext to give it in that representation.

leleuj:
Indeed, it's not an easy problem. Let's keep this issue open and marked as "FUTURE" to track the need.

leleuj:
After some thinking on this, I think we should add:

- a `String getRequestContent()` method on the `WebContext` interface, with a default implementation throwing a "not implemented exception" (see: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/context/WebContext.java#L78) to make things backward compatible (at the compilation level) with existing implementations

- the appropriate implementation in the `J2EContext` class.

@victornoel : Do you mind submitting a pull request for that?

mmoayyed:
I'll handle this.

millross:
As a heads up, this approach is unlikely to work very well in at least the vert.x framework. I would prefer a generalised version of the method (which would mean an additional type parameter on the WebContext interface) where the type parameter would determine the type returned from GetRequestContent. That said, I'd far prefer a mechanism for accessing the underlying request (or context) full stop, to enable the coder to work within the parameters of their chosen framework.

victornoel:
In the end, I wonder if it makes really sense to push this method at the interface level instead of simply having it in one of the implementation of the interface (such as `J2EContext`).

[...]

PR #639, opened 2016-08-29
Allow context implementations to return the request content

mmoayyed:
Closes https://github.com/pac4j/pac4j/issues/612

I contemplated writing a test, but didn't see value in testing a mock.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/context/WebContext.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/context/WebContext.java
@@ -1,5 +1,7 @@
 package org.pac4j.core.context;
 
+import org.pac4j.core.exception.TechnicalException;
+
 import java.util.Collection;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


pac4j-core/src/main/java/org/pac4j/core/context/WebContext.java
```diff
@@ -180,4 +182,12 @@ default void invalidationSession() {
...
      */
     String getPath();
+
+    /**
+     * Gets content body of the original request.
+     *
+     * @return the request content
+     * @since 1.9.2
+     */
+    String getRequestContent();
```
victornoel: I think the idea was to provide a default implementation for it…

leleuj: Yes, the idea was to provide a `default` implementation to keep backward compatibility (by returning an empty string).

leleuj: Rather throw a "Not supported by this framework" exception.

Issue #633, opened 2016-08-18
SAML2Client does not map ADFS role claims to user roles.


rhdunn:
Using the `RequireAnyRoleAuthorizer` with a SAML2 client connected to an ADFS server, Role claims are not mapped to user roles, so the authorizer fails.

The ADFS roles are set as "http://schemas.microsoft.com/ws/2008/06/identity/claims/role" claims. These appear as attributes, but do not appear as user roles.

leleuj:
The link is broken, but it makes sense.

Would you mind submitting a PR for that?

rhdunn:
That link is the claim ID (mapped to the SAML attribute name). See https://technet.microsoft.com/en-us/library/ee913589(v=ws.11).aspx for the claim type URIs supported by ADFS. See https://azure.microsoft.com/en-gb/documentation/articles/active-directory-token-and-claims/ for an example of the SAML XML.

I'll look into creating a PR.

mmoayyed:
This can be closed now.

leleuj:
OK. Thanks

PR #640, opened 2016-08-29
Provide an authorizer that works based on attributes

mmoayyed:
See javadocs and tests

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/authorization/authorizer/RequireAnyAttributeAuthorizer.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/authorization/authorizer/RequireAnyAttributeAuthorizer.java
@@ -1,0 +1,51 @@
+package org.pac4j.core.authorization.authorizer;
+
+import org.pac4j.core.context.WebContext;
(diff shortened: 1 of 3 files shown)


mmoayyed:
This may also be used partially to handle https://github.com/pac4j/pac4j/issues/633

pac4j-core/src/main/java/org/pac4j/core/authorization/authorizer/RequireAnyAttributeAuthorizer.java
```diff
@@ -0,0 +1,51 @@
...
+        this.valueToMatch = valueToMatch;
+    }
+
+    @Override
+    protected boolean check(final WebContext context, final U profile, final String element) throws HttpAction {
+        if (!profile.containsAttribute(element)) {
+            return false;
+        }
+
+        if (StringUtils.isEmpty(this.valueToMatch)) {
```
leleuj: Spring framework is only an optional dependency. You should use the `CommonHelper` component here.

leleuj:
+1 with a small change

leleuj:
Good. Thanks

Issue #646, opened 2016-09-01
SAML AuthN requests are not signed, yet metadata says they should be


mmoayyed:
Metadata generated by pac4j looks something like this:

```xml
<md:SPSSODescriptor AuthnRequestsSigned="true" WantAssertionsSigned="true" 
protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol urn:oasis:names:tc:SAML:1.0:protocol urn:oasis:names:tc:SAML:1.1:protocol">
```

`AuthnRequestsSigned="true"`, yet pac4j is not actually signing authn requests. This setting should be taken into account when sending and building authn requests.

mmoayyed:
Detected by CAS running in SAML2 mode:

```
<Metadata for [urn:mace:saml:pac4j.org] says authentication requests are signed, yet this authentication request is not>
```

mmoayyed:
Still trying to make sense out of this. Pac4j definitely signs the authn requests based on that flag, however our encoder removes the signature before actually encoding it. In `Pac4jHTTPRedirectDeflateEncoder`, there is:

```java
 @Override
    protected void doEncode() throws MessageEncodingException {
        final MessageContext messageContext = this.getMessageContext();
        final SAMLObject outboundMessage = (SAMLObject)messageContext.getMessage();
        final String endpointURL = this.getEndpointURL(messageContext).toString();
        this.removeSignature(outboundMessage); //not sure why?
...
```

Behavior seems to be true for both opensaml 3.1 and 3.2.

mmoayyed:
@leleuj This might be something for you to tackle or advise: assuming I can bypass the above issue, it appears that the `RunClient` machinery only handles GET and not POSTs. For SAML, the callback URL is always POSTed to by the IdP, and providing that url manually in the prompt does it no good.

leleuj:
I need to take a closer look at this, we have some setting to indicate if the authentication request must be signed or not and this must be taken into account.

mmoayyed:
Related post http://shibboleth.1660669.n2.nabble.com/HTTPRedirectDeflateEncoder-removing-AuthnRequest-signatures-tt7627970.html

Issue #645, opened 2016-09-01
SAML SP metadata generation should verify the directory path


mmoayyed:
If I set the path to my SP metadata to be: `/Misagh/Somewhere/Invalid/sp-metadata.xml`, pac4j should make sure `/Misagh/Somewhere/Invalid/` exists and if necessary, create that path.

mmoayyed:
Related to #644

Issue #644, opened 2016-09-01
CommonHelper.getResource() always assumes resource exists


mmoayyed:
```java
	public static Resource getResource(final String filePath) {
		return new WritableResource() {

			@Override
			public InputStream getInputStream() throws IOException {
				return getInputStreamFromName(filePath);
			}

			@Override
			public String getFilename() {
				return filePath;
			}

			@Override
			public boolean exists() {
				return true; //problem!
			}
			
			@Override
			public OutputStream getOutputStream() throws IOException {
				return new FileOutputStream(filePath);
			}

		};
	}
```

Issue #643, opened 2016-09-01
Add a RunSAML2Client test


leleuj:
From the lost class: `SAML2ClientIT` and subclasses...

mmoayyed:
Thanks for tracking this.

I am thinking we should take this one step further: Build an embedded CAS server into pac4j for IT tests, and use that for both CAS and SAML tests. (We might optionally do one based on testshib but as you said, the service is a little unreliable). We might have an embedded CAS server in pac4j already, don't know, but it would be good to bump to CAS5 and have it be the external IdP for pac4j tests.

Sound good?

mmoayyed:
Actually, thinking more about this: we don't need to embed. Probably could just run the tests manually and assume the env is up and running.

leleuj:
Yes, I think it might be too much to have an embedded CAS server for that. For a major release, running the manual tests just takes a couple of minutes.

Thinking a little more about this, I think I removed the SAML test because we already check that via the various demos we have.
But we can certainly have one in pac4j (like for Facebook).

leleuj:
Moving this one for 1.9.3, to start working on the 1.9.2 release.

I'll re-test the SAML support with manual tests on the demos (J2E & Play).

PR #647, opened 2016-09-01
SAML Bug Fixes: Metadata Path, Resource Checking, AuthnRequest signature etc

mmoayyed:
This is a WIP, but I'll continue to push so you could see what changes are being added.

This closes hopefully #646 #645 #644 #643


mmoayyed:
I am testing this with CAS5 RC2 running as a SAML IdP, and I realized redirect binding isnt quite supported with signed authn requests. I have added an option in pac4j to allow then, and will continue with CAS to handle signature values as request parameters.

leleuj:
Thanks for taking these ones. Just let me know when you're done for the code review.

mmoayyed:
I was able to figure out why pac4j isn't signing the requests. That bit is done and is now configurable. I also need to fix the CAS module to let it recognize the signature of the request for redirect bindings. However, once I do, I need your help to figure out how to run tests that produce a POST rather than a GET. I suppose you can even do that independant of this PR here and when it's ready, I'll merge and test it here.

leleuj:
Are you referring to a POST (vs GET) towards the SAML IdP? In that case, indeed, it's a problem because the manual test is made for the user to copy/paste an url in the navigator, but we can output the HTML instead of the redirection url by changing things here: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/test/java/org/pac4j/core/run/RunClient.java#L37

Do your best, I'll finish things if necessary.

BTW, I hope to finish the new documentation before you finish this PR, so that you can review, complement and improve the SAML documentation.

mmoayyed:
I have opted to bail on the SAML IT tests for now. I may do that bit in a separate pull request, given this PR is already complicated as it. Here is a quick update:

1. Bumped OpenSAML version
2. Made sure metadata resource existence is properly handled and checked.
3. Made sure authn requests are properly XML signed if needed.
4. Made sure metadata paths are correctly created if needed.

I'll wait for Travis to confirm.

leleuj:
Travis is unhappy: "You have 3 PMD violations."

Apart from that, I finally added the documentation into the pac4j project: can you update this documentation with your changes: https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/saml.md ?

Reviewing your changes...

pac4j-core/src/main/java/org/pac4j/core/io/Resource.java
```diff
@@ -1,21 +1,21 @@
...
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 
 /**
  * Interface for a resource descriptor that abstracts from the actual
  * type of underlying resource, such as a file or class path resource.
- *
+ * <p>
```
leleuj: I'm surprised that the Javadoc task supports unclosed `<p>` tags, but let the build says if it likes it or not.

pac4j-core/src/main/java/org/pac4j/core/io/SpringResourceWrapper.java
```diff
@@ -31,4 +37,13 @@ public String getFilename() {
 		return springResource.getFilename();
 	}
 
+	@Override
+	public File getFile() {
+		try {
+			return springResource.getFile();
+		} catch (final Exception e) {
+			LOGGER.error(e.getMessage(), e);
```
leleuj: It's a not a big deal, but shouldn't we throw a `TechnicalException` instead?

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
```diff
@@ -40,6 +40,8 @@
 
     private boolean forceAuth = false;
 
+    private boolean forceSignRedirectBindingAuthnRequest;
```
leleuj: These new options will need appropriate documentation ;-)

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
```diff
@@ -63,7 +65,8 @@ public SAML2ClientConfiguration() {}
     private List<String> signatureReferenceDigestMethods;
     private String signatureCanonicalizationAlgorithm;
     private boolean wantsAssertionsSigned = true;
-
+    private boolean signErrorResponses;
```
leleuj: What's the meaning of this new property?

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2ServiceProviderMetadataResolver.java
```diff
@@ -126,7 +127,11 @@ public final MetadataResolver resolve() {
                     logger.info("Metadata file already exists at {}.", this.spMetadataResource.getFilename());
                 } else {
                     logger.info("Writing sp metadata to {}", this.spMetadataResource.getFilename());
-
+                    final File parent = spMetadataResource.getFile().getParentFile();
+                    if (!parent.mkdirs()) {
+                        logger.error("Could not construct the directory structure for SP metadata {}",
```
leleuj: Shouldn't we throw a `TechnicalException` here as we were unable to create the directory where the SP metadata will be written?

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2WebSSOMessageSender.java
```diff
@@ -119,7 +123,7 @@ protected final void invokeOutboundMessageHandlers(final SPSSODescriptor spDescr
                 final SAMLOutboundProtocolMessageSigningHandler handler = new
                         SAMLOutboundProtocolMessageSigningHandler();
                 handler.setSignErrorResponses(this.signErrorResponses);
-                handler.invoke(outboundContext);
+                //handler.invoke(outboundContext);
```
leleuj: I'm surprised to see that line commented. I had in mind that this was were the work was done...

pac4j-saml/src/test/java/org/pac4j/saml/sso/impl/SAML2DefaultResponseValidatorTests.java
```diff
@@ -1,100 +0,0 @@
...
-import org.opensaml.saml.saml2.metadata.SPSSODescriptor;
-import org.pac4j.saml.context.SAML2MessageContext;
-import org.pac4j.saml.crypto.SAML2SignatureTrustEngineProvider;
-import org.pac4j.saml.exceptions.SAMLException;
-
-import static org.junit.Assert.*;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
-
-public class SAML2DefaultResponseValidatorTests {
```
leleuj: This test class is no longer necessary?

leleuj:
A few comments here and there, but certainly mostly because of my low skills on SAML...

+1

pac4j-core/src/main/java/org/pac4j/core/io/Resource.java
mmoayyed: Looks like Travis is skipping Javadocs: `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V`

leleuj: Yes, but I'm pretty sure I won't when I'll perform the release ;-)

leleuj:
Unfortunately, tests are broken due to the fact we can't create write the SP metadata...

mmoayyed:
I am playing around with Travis to see why tests pass locally for me but not Travis! This will eventually pass at some point.

leleuj:
Excellent! I'm ready to merge. Anything to add?

I'll reopen #643 and do it myself.

mmoayyed:
Sounds good. Go ahead. I'll take a pass at the docs tomorrow as well (with whatever updates this PR has) and see how things look.

Issue #617, opened 2016-07-28
Introduce comparison method in PasswordEncoder


victornoel:
Hi,

When using hashing/encryption methods for password, sometimes the algorithm to compare the password is specific to the algorithm.

A good example of that is bcrypt. If you look, for example, at the implementation of the pluggable [password encoder for bcrypt from spring-security](https://github.com/spring-projects/spring-security/blob/master/crypto/src/main/java/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.java), you can see it has an `encode` and a `matches` methods.

It means that currently, it is not possible to exploit the `PasswordEncoder` abstraction of pac4j if one wants to use bcryt (for example with jBcrypt).

If you decide to introduce such an extra method to `PasswordEncoder` and exploit it as wanted in pac4j, then we could add a bcrypt implementation of it :)

leleuj:
You raise a good point about hashing / encryption for password. The pac4j `PasswordEncoder` was a first implementation and indeed, the `matches` method is missing.

pac4j is focused on authentication / authorization and I definitely do not want to spend time and energy on encryption / password which are fairly though topics, while there are very good libraries such as Spring Security and Shiro.

So we should:
- add a `matches(String, String)` method to the pac4j `PasswordEncoder`
- add the Spring Security Crypto and Shiro core libraries as optional dependencies in the pac4j core
- create a `SpringSecurityPasswordEncoder` implementing the pac4j `PasswordEncoder` and wrapping a Spring Security [`PasswordEncoder`](https://github.com/spring-projects/spring-security/blob/master/crypto/src/main/java/org/springframework/security/crypto/password/PasswordEncoder.java) 
- create a `ShiroPasswordEncoder` implementing the pac4j `PasswordEncoder` and wrapping the Shiro [`PasswordService`](https://shiro.apache.org/static/1.2.5/apidocs/org/apache/shiro/authc/credential/PasswordService.html)

This way, we could use all the already implemented algorithms available in Spring Security or Shiro.

Would you mind submitting a pull request for that?

Thanks.

victornoel:
Hi,

Yes, I can take care of that, but unfortunately, I'm going to be unavailable for a month, so it won't be before September :)

leleuj:
No problem, lets' do that in a month...

victornoel:
I submitted a PR for this issue (#649) :)

Issue #339, opened 2015-11-06
Account registration


leleuj:
We certainly need a `ProfileWriter` for `LdapProfile`, `MongoProfile`... to create user accounts.

leleuj:
Handle roles / permissions when saving / restoring the user profile.

AzureusNation:
I would really appreciate if this could be squeezed in to 2.0. This would add so much more functionality and may more benefits for using this library.

leleuj:
Yes, I'm considering this option more and more. We'll see after the RC2 release.

leleuj:
@dfish3r :

As you may already know, pac4j uses Ldaptive to handle LDAP authentication (credentials validation): https://github.com/pac4j/pac4j/blob/master/pac4j-ldap/src/main/java/org/pac4j/ldap/credentials/authenticator/LdapAuthenticator.java

I'm currently working on enhancing the LDAP support by allowing the addition, modification and removal of a user in a LDAP.
I've started developing the user creation: https://github.com/leleuj/pac4j/blob/pac4j-339/pac4j-ldap/src/main/java/org/pac4j/ldap/profile/service/LdapProfileService.java#L74

I create a Ldap entry based on given attributes and save it in the LDAP using a `Connection` acquired from a `ConnectionFactory`. I followed the provided example: http://www.ldaptive.org/docs/guide/operations/add.html

Is this the right way to do things? Will it work for all LDAPs?
How can I set the password user? Like a regular attribute?

Thanks.

dfish3r:
> Is this the right way to do things?

Looks ok, you might want to document that you don't support binary attribute values.

> Will it work for all LDAPs?

That's not a simple question. There are lots of ways to configure an LDAP that makes entry creation by an external client difficult. You should spin up a few test fixtures (OpenLDAP, ApacheDS, OpenDJ, etc) to confirm your code works, but know that authorization and schema will be brittle points in the process for your deployers.

> How can I set the password user? Like a regular attribute?

Yes.
Typically something like userPassword: {SHA}yP7QDr........rDhIw=

Also, I'm assuming none of your questions are asking about Active Directory :wink:

leleuj:
Thanks for the feedback.

Sure. I'll take that into account in the documentation.

I've made it work for an embedded LDAP in my tests, so it's going to be mostly a base class for LDAP support. We'll see if adopters require / propose more specific implementations.

PR #649, opened 2016-09-05
Introduces matches() in PasswordEncoder

victornoel:
Here is a PR for #617 :)

I added the `matches` method to `PasswordEncoder`.
I added three password encoders: one for Spring, one for Shiro and one for jBcrypt.

I updated `MongoAuthenticator` and `DbAuthenticator` to take advantage of the `matches` method.

At first I wanted to update also `StormpathAuthenticator` (which extends `AbstractUsernamePasswordAuthenticator` like the other two), but it seems to me that `StormpathAuthenticator` really only need the Encoding part of the `PasswordEncoder`, the matching seems to be handled by the Stormpath library itself. Thus I really wonder if it makes sense to encode the password in the `StormpathAuthenticator`! But this is out of scope of the current PR, I can open an issue if you want.


pac4j-core/src/main/java/org/pac4j/core/credentials/password/ShiroPasswordEncoder.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @author Victor Noël
+ * @since 1.9.2
+ */
+public class ShiroPasswordEncoder implements PasswordEncoder {
+
+    private final PasswordService delegate;
+
+    public ShiroPasswordEncoder(final PasswordService delegate) {
+        this.delegate = delegate;
```
leleuj: Can you add a `CommonHelper.assertNotNull("delegate", delegate);` check?

pac4j-core/src/main/java/org/pac4j/core/credentials/password/SpringSecurityPasswordEncoder.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @author Victor Noël
+ * @since 1.9.2
+ */
+public class SpringSecurityPasswordEncoder implements PasswordEncoder {
+
+    private final org.springframework.security.crypto.password.PasswordEncoder delegate;
+
+    public SpringSecurityPasswordEncoder(final org.springframework.security.crypto.password.PasswordEncoder delegate) {
+        this.delegate = delegate;
```
leleuj: Can you add a `CommonHelper.assertNotNull("delegate", delegate);` check?

leleuj:
Excellent! See my few comments.

Can you update the documentation as well: https://github.com/pac4j/pac4j/blob/master/documentation/docs/authenticators.md#2-passwordencoder? Thanks

victornoel:
@leleuj I wonder if there is not a confusion of functionality of PasswordEncoder with this PR…

According to the documentation, the role of `PasswordEncoder` is to encode password before passing them to an authentication service, while the objective of this PR is to enable authentication services (mainly the DB ones, because most of the the others impose their own algorithm) to have a pluggable password checking algorithm… I guess that explains my problem with `StormpathAuthenticator`…

Maybe we should separate the interface of `PasswordEncoder` from the interface of `PasswordMatcher`?
`StormpathAuthenticator` would use a `PasswordEncoder`, while `MongoAuthenticator` would use a `PasswordMatcher` for example…

Or am I over-complexifying things and the documentation should simply be clearer on that? :)

victornoel:
(or maybe there is simply no reason for PasswordEncoder to be used in StormpathAuthenticator?)

leleuj:
The `encode` and `matches` methods are closely linked. I think both should be in the same interface as you did. And this capability makes sense when we manage the source of data (MongoDB, SQL db), but not for Stormpath.

So I think we must not have any `PasswordEncoder` for the `StormpathAuthenticator` as Stormpath handles passwords on its own.

And "the documentation should simply be clearer on that".

victornoel:
@leleuj all done, I made 2 PR, one for Stormpath (#651) and one for the documentation (#653).

Interestingly enough, now the `PasswordEncoder.encode()` method is not used anywhere except in tests!
It makes sense since encode seems to be needed only when adding users to database such as the one used by `MongoAuthenticator` and `Dbauthenticator`.

leleuj:
@mmoayyed : I'm about to merge this excellent PR and the #653 as well unless you have a concern...

leleuj:
@victornoel : about the `PasswordEncoder.encode` method which is used only in tests, it will be used for account creation one day (see: #339).

mmoayyed:
I am OK with this, but I don't really this moving us towards wrappers. `JBCryptPasswordEncoder` is an exact replica of Spring Security. Pac4j should never have a `JBCryptPasswordEncoder` in my opinion. It should simply wrap SS's PasswordEncoder's interface, and delegate. That's all.

Come to think of it, even providing a wrapper is maybe too much. Remove the middle man.

victornoel:
@mmoayyed I introduced JBCryptPasswordEncoder because it uses a library that only implement bcrypt, while SS has its own bcrypt implementation. But it's true that SS implementation should be good and correct (since they have a huge community, there is more chance for bugs to be caught quickly).
@leleuj should I remove JBCryptPasswordEncoder? SS Crypto is quite small (contrary to shiro for which we need to bring everything) I guess, so maybe no need for this duplication?

victornoel:
@leleuj the commit for the documentation is now in this PR :)

leleuj:
I like this PR the way it is. Proposing `JBCrypt` may be overkill, but it's just a class.

And I'm convinced we need the wrapper for those who want to use Shiro and not Spring Security. And the wrapper is a simple class, not a whole mechanism. And removing it would introduce a dependency on Spring Security whenever you use the `DbAuthenticator` and the `MongoAuthenticator`. And I definitely don't want a required dependency on Spring Security.

leleuj:
@victornoel : thanks for the documentation

mmoayyed:
> but it's just a class.

There are a few of such classes today. Take into account the "Resource" API. That's a duplicate. The password encoder API also a duplicate in some ways. (Some duplication in saml as well) I am concerned that we are asking developers to learn pac4j where it shouldn't matter.  The wrapper idea is fine, but I am also concerned that it would lead into API bloat overtime where you have to maintain API differences across Shiro and SS (and XYZ later), and then have to account for cases where one provides X and the other does not, which means you'd have separate wrappers for each, which basically also means you will be reinventing the wheel where round was a perfectly good enough shape.

I get the use case though and what you're trying to achieve, and I like the idea of being able to plug a password encoder instance of "any" framework into an authenticator and it works. That's great. API-wise though, I think we need to a bit more careful and integrate better rather than working actively "against" a particular framework.

I hate duplication, altogether. and I like doing less work when someone else has already figured out how to perfectly get that done. No offense against this PR or its author at all, but technically speaking, much of Encoders provided here are duplicates of what other frameworks provide more or less.

mmoayyed:
PS I am just brainstorming here out loud, and what this might eventually turn into long-term. For the time being, we should proceed.

victornoel:
@mmoayyed no offense taken, I quite agree with your conclusions actually :)

leleuj:
I also hate duplication, but how could we handle Shiro and Spring Security without a wrapper? Or you want use just to stick with Spring Security?

In all cases, I prefer a wrapper than reinventing the wheel.

And in fact, I don't even want a required dependency on Spring Security or Spring (to replace the `Resource` class) or Shiro. Because the core idea of pac4j is to be framework agnostic because in most cases having Spring makes sense, but in Play or Vertx, it doesn't.

Merging...

mmoayyed:
In most cases, I find sticking with "one thing" generally simpler. We just have to pick from a menu. As a developer, what does Shiro give me in terms of password encoding that SS does not? Or vice versa? If I am using pac4j with Shiro, why would it be a problem to use provided SS's password encoders? You just add one extra dependency (and maybe that's bad) but the alternative is just as bad because of the API bloat, and extra dependencies you added already (like the ones that handle bcrypt and such). A jar is a jar is a jar. Why is adding bcrypt good but adding SS is bad? Why is adding Spring bad, but creating/duplicating Spring classes are good? I don't follow that logic. Any pac4j framework can exclude dependencies that they don't need. The dependencies also don't need to be forced down to the frameworks, as they can be modularized and declared declaratively! I don't see a problem there at all. 

Being framework agnostic is great when you are actually being agnostic. This is not. Here is a theoretical case I submit you; it's a classic decorator-pattern problem:

SS has:

```
class SpringSecurityEncoder
    matches()
    doesSomethingCool()
```

And Shiro may have:


```
class ShiroEncoder
    checks()
    doesFunStuff()
    gotSomethingElseButIAlsoGiveYou()
    matches(hasExtraParameter)
```

And something else has:


```
class SomeOtherSecurityFrameworkThoughtItsCoolToCreateYetAnotherEncoder
    ensure()
    handle()
```

How would you have wrappers to solve this problem? This certainly will lead to API bloat. You could have SpringSecurityEncoderWrapper, ShiroEncoderWrapper, etc? Or maybe you have Pac4jGenericWrapper and somehow handle various implementations?

None is going to scale or be that friendly at all. In fact, either solution will force the developer to pass through proxies and middlemen to get something working just right. It's a whole lot easier to just stick to one thing and provide the functionality. And as a developer, if I am securing my applications with Shiro + Pac4j, I don't suddenly start to hate on Spring Security! :) I am happy to use their password encoder functionality, specially if it's more accepted, closely maintained and used by the globe all over (or vice versa...doesn't matter). 

But let's say we want to support everything and anything. If we really want to be agnostic and make it easier to accommodate all cases then we should not have to ask the developer to learn this wrapper business, or anything else for that matter. Perhaps we can "detect" what is actually intended to be used by the developer and pick that up easily from the classpath. 

A good example of this might be how Spring Boot for instance detects caching technologies and auto-configures one for you. You have nothing? you get a basic cache. You have Guava? it detects and uses that. You have Caffein or Spring Modules Cache? You can use those too. How about Ehcache or Hazelcast? Sure they work too. That's being agnostic. Do you suppose they handle all those conditionals with wrappers or other types of introspection? No. That would be insane to maintain API consistency across frameworks. They go way beyond. As a developer you don't have to even CARE how any of that is wired up. You just know you want to use Guava. That's all you care about. What classes to use? What to plug in? where? what method to call? All of that is boilerplate. 

Of course, I don't have all the answers. This may entirely turn into something dysfunctional but it's something to think about.

victornoel:
Just some notes, not to contradict but to help clarify some points.

I don't think we are in a similar case to the cache example: it's not about configuring a component (in the sense of runtime component) in the system, it is simply about using a module (in the sense of compile time module): choosing one or the other of the password encoding algorithm won't have any impact on the rest of the application, just the authenticator and how it handle passwords.
This feature is clear and well defined: there is encoding and there is matching.
So we can't really compare the way springboot is agnostic to how pac4j wants to be agnostic on password encoding.

It's true then, that we could simply use the most usable crypto library (I think it would be spring security crypto in this case). I'm not really sure why we would need shiro on top of ss currently, especially because their PasswordService is more a runtime component than a compile  time module… Inversely, the bcrypt dependency provides the same implementation than SS, but a newer version, so potentially less buggy…

Also, just to correct something, the bcrypt dependency added is simply optional, like the SS and the Shiro one are: so people need to add them if they want to take advantage of these wrappers, but until then, it won't bloat anything.

leleuj:
Interesting discussion and I'm sensitive to exposed arguments.

For password encoding, this is not something out of the box: people will make a choice, I want SHA512 because this is what I already use / know or my company's recommendation. And I know well Shiro or Spring Security and I'd like to use the framework I know.

For the "Resource" API, as it's deeply in the heart of pac4j, using Spring directly would make the Spring dependency unremovable. And pac4j aims to target multiple frameworks and for Play, DI is handled by Guice, Spring makes no sense.

I don't think there is any general recipe: it's about finding the right compromise, balance.

Are we abusing on wrapper? I don't think so, we have the `PasswordEncoder` and the `Resource`. Are we proposing too any alternatives? maybe we don't really need JBCrypt, but Shiro and Spring Security are popular and valid alternatives.

PR #650, opened 2016-09-05
Improve Mongo support

victornoel:
Hi,

Here is a PR (to be discussed :) to improve the mongodb support.

It introduces support for another library to access mongo named mongo-async-driver from Allanbank. It a driver I use a lot (because it is supported by [Quasar](http://docs.paralleluniverse.co/quasar/) mainly and I needed such support).
I factored the common code in a class named `AbstractMongoAuthenticator`.

The old MongoAuthenticator is still usable exactly as before EXCEPT for the fact that the dependency is now optional (since we have 2 different libraries to access mongo in the dependencies now). Not sure how to handle that better…

Some thoughts I had when I made this PR: in a way the common code is very similar to the code present in pac4j-sql: maybe it would make sense to gather all these database-oriented authenticators in one package, or inversely have the common code in pac4j-core and have a package for each of the database supported. It's just an idea, maybe it's not worth the effort to do all of that :)

Thanks!

### DIFF
--- a/pac4j-mongo/pom.xml
+++ b/pac4j-mongo/pom.xml
@@ -13,7 +13,7 @@
 	<name>pac4j for MongoDB</name>
 
 	<properties>
-		<mongo-driver.version>3.2.1</mongo-driver.version>
+		<mongo-driver.version>3.3.0</mongo-driver.version>
 		<mongo-allanbank.version>2.0.1</mongo-allanbank.version>
 		<flapdoodle.version>1.50.2</flapdoodle.version>
 	</properties>
@@ -25,7 +25,7 @@
 		</dependency>
 		<dependency>
 			<groupId>org.mongodb</groupId>
-			<artifactId>mongo-java-driver</artifactId>
+			<artifactId>mongodb-driver</artifactId>
 			<version>${mongo-driver.version}</version>
 			<optional>true</optional>
 		</dependency>
--- a/pac4j-mongo/pom.xml
+++ b/pac4j-mongo/pom.xml
@@ -14,6 +14,7 @@
 
 	<properties>
 		<mongo-driver.version>3.2.1</mongo-driver.version>
+		<mongo-allanbank.version>2.0.1</mongo-allanbank.version>
 		<flapdoodle.version>1.50.2</flapdoodle.version>
 	</properties>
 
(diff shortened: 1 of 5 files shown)


victornoel:
Also, careful, the dependency for allanbank's mongo driver is not on Maven Central, you have to add a [specific repository](http://www.allanbank.com/mongodb-async-driver/userguide/maven.html): I wasn't sure if I should add it to the `pom.xml`, because it's not a good practice to do so (I personally have a Nexus that mirrors their repository, on top of others, but if not I would have put it in my `settings.xml`…), but we can simply add it there if needed.

mmoayyed:
What is the use case here?

victornoel:
@mmoayyed the use case is being able to use the allanbank mongodb driver instead of the official mongodb driver to authenticate users from a mongo database in pac4j. I thought it was clear! Or is the question not about that?

mmoayyed:
It's not clear to me why one would want to use one vs the other, and it's not clear why pac4j would want to ship with two almost identical? authentication methods for mongo and it's not clear why this should not be the default if it's better than the default and it's not clear why this should go into the core library as opposed to just being an extension. 

You didn't describe a use case. You simply provided a feature.

leleuj:
I agree with @mmoayyed 

Is the allanbank MongoDB driver still an active project: https://github.com/allanbank/mongodb-async-driver? It feels dead (latest commit two years ago).

Having an async driver does not make much sense for a sync library like pac4j and having two MongoDB supports either.

It's really a too specific need to be accepted by the pac4j project.

That said, I see two good points: I'm willing to consider some refactoring which makes extending our current `MongoAuthenticator` easier and the idea of common code between SQL and MongoDB could be a relevant refactoring (for now, I'm not sure we should go in that direction).

victornoel:
@leleuj alright, I will maintain it in my own project :) I agree with you for the non-need to currently refactor the db authenticator if it is not to integrate such a PR now.

FYI, just as background explanations of this PR, here are some comments:

The allanbank driver is not dead, but they seem to do development internally and push their changes for a release… 

Also note that this PR was using the synchronous API of the allanbank driver (which is implemented using the asynchronous one under-the-hood), not the asynchronous one.

The advantage of using this library over the official one can be one of the following:

- Better performances (of both the BSON and the networking aspects)
- Support by Quasar: basically, Quasar transforms asynchronous APIs such as the allanbank one into efficient (as efficient as a asynchronous API can be) synchronous APIs (in that case the sync to async implementation is handled by quasar integrations but using the same interface as before). Such synchronous APIs are called fiber-blocking APIs because instead of using threads that blocks (like the official mongodb driver do), it uses fibers which are light-thread that have very efficient blocking which gives performance as good as asynchronous programming but with a synchronous programming. It composes very well with web-oriented frameworks hence the relation to pac4j :)

@mmoayyed you are right, it was not clear, the use case was simply to be able to do fiber-blocking (as explained above) authentication with pac4j. You simply can't with the official driver and the way it is implemented prevent its integration with quasar (except with a lot of effort).

Anyway, no problem, I did this work internally, so I thought it could have been useful :)

PR #651, opened 2016-09-06
Remove unneeded PasswordEncoder use in StormpathAuthenticator

victornoel:
As discussed in #649, we should remove the use of `PasswordEncoder` in `StormpathAuthenticator`.

@mmoayyed please could you confirm we didn't misunderstood something with @leleuj? Thanks :)


mmoayyed:
This looks fine.

In principle though, I'd say I am very much against pac4j having its own PasswordEncoder. The functionality presented by pac4j is almost an exact replica of Spring Security. I see no reason why that has to be duplicated, specially if it's not made any easier to work with. It can't get any easier than Spring Security. Long-term, my suggestion is to remove all pac4j specific components in this area and use SS as a replacement.

victornoel:
@mmoayyed Exactly what @leleuj said, it is covered by #649 which introduces a SS wrapper. But if you mean we should simply relies on SS and not introduce any interface at all, then please come discuss it in #649 before it is merged.

leleuj:
@mmoayyed : we agree, pac4j does not intent to reinvent the wheel for password encoding (encryption or hashing), but we need a wrapper (`PasswordEncoder`) as we may want to use SS, but not only: **also Shiro or jBCrypt** are good candidates. See: #649 indeed.

PR #653, opened 2016-09-06
Update documentation for #617

victornoel:
As discussed in #649, here is an update of the documentation about PasswordEncoder.


leleuj:
In fact, I had in mind that the documentation changes would be in the #649 PR (code and documentation changes at the same time), but it's not a big deal.

victornoel:
@leleuj  I don't think we can modify two branches (master and gh-pages) in one PR, can we?

leleuj:
In fact, you just need to modify the `master` branch (in the `documentation` directory which is copied to the `gh-pages` branch).

victornoel:
ha ok, so we should not merge this PR, let's close it and modify directly master, I will do that, sorry.

PR #660, opened 2016-09-06
Let the BooleanConverter instances to return false rather than null

### LINKED ISSUE #654: Let the BooleanConverter instances to return false rather than null
Example:

```java
public Boolean convert(final Object attribute) {
        if (attribute != null) {
            if (attribute instanceof Boolean) {
                return (Boolean) attribute;
            } 
            if (attribute instanceof String) {
                return Boolean.parseBoolean((String) attribute);
            }
        }
        return null; //not cool.
    }
```

> A method that returns either Boolean.TRUE, Boolean.FALSE or null is an accident waiting to happen. This method can be invoked as though it returned a value of type boolean, and the compiler will insert automatic unboxing of the Boolean value. If a null value is returned, this will result in a NullPointerException.

leleuj:
Closes #654


mmoayyed:
Sweet :) Merge away.

PR #665, opened 2016-09-07
Auto generate saml keystore if none is defined

### LINKED ISSUE #655: Automate the generation of SAML keystore for SP metadata
The SAML functionality of pac4j requires that one does:

```
keytool -genkeypair -alias pac4j-demo -keypass pac4j-demo-passwd -keystore samlKeystore.jks -storepass pac4j-demo-passwd -keyalg RSA -keysize 2048 -validity 3650
```

Make this optional and automate this step. Let pac4j generate this keystore much like the SP metadata that is generated.

mmoayyed:
Closes #655


mmoayyed:
This seems like a bad PMD match:

```
[INFO] Exception is caught when Exception is not thrown in org.pac4j.saml.client.SAML2ClientConfiguration.createKeystore() [org.pac4j.saml.client.SAML2ClientConfiguration] At SAML2ClientConfiguration.java:[line 469] REC_CATCH_EXCEPTIO
```

Exception is definitely thrown.

leleuj:
It's related to FindBugs and not PMD, isn't it?

The documentation says:

**REC: Exception is caught when Exception is not thrown (REC_CATCH_EXCEPTION)**

This method uses a try-catch block that catches Exception objects, but Exception is not thrown within the try block, and RuntimeException is not explicitly caught. It is a common bug pattern to say try { ... } catch (Exception e) { something } as a shorthand for catching a number of types of exception each of whose catch blocks is identical, but this construct also accidentally catches RuntimeException as well, masking potential bugs.

A better approach is to either explicitly catch the specific exceptions that are thrown, or to explicitly catch RuntimeException exception, rethrow it, and then catch all non-Runtime Exceptions, as shown below:

```java
  try {
    ...
  } catch (RuntimeException e) {
    throw e;
  } catch (Exception e) {
    ... deal with all non-runtime exceptions ...
  }
```

It makes sense, but here you catch the exceptions to throw a `SAMLException` so I'm not sure this is really relevant for you: can you try to explicitly catch the appropriate exceptions?

mmoayyed:
You mean turn:

```
} catch (Exception e) {
    throw something...
}
```


into:

```
        } catch (final KeyStoreException e) {
            throw new SAMLException("Could not create keystore", e);
        } catch (final UnknownHostException e) {
            throw new SAMLException("Could not locate host", e);
        } catch (final NoSuchAlgorithmException e) {
            throw new SAMLException(e.getMessage(), e);
        } catch (final InvalidKeyException e) {
            throw new SAMLException(e.getMessage(), e);
        } catch (final CertificateEncodingException e) {
            throw new SAMLException(e.getMessage(), e);
        } catch (final SignatureException e) {
            throw new SAMLException(e.getMessage(), e);
        } catch (final CertificateException e) {
            throw new SAMLException(e.getMessage(), e);
        } catch (final IOException e) {
            throw new SAMLException(e.getMessage(), e);
        }
```

I am not that is any better really. That's just noise. The rule should be turned off or turn into a warning.

leleuj:
Of course not. There are so many exceptions?

I meant for example:

```java
} catch (final SignatureException | CertificateException | IOException e) {
    throw new SAMLException(e.getMessage(), e);
}
```

Though, this may not be feasible.

Let's make it a warning or ignore it for this class.

mmoayyed:
Sounds good. Dealing with certs and keystores, yes; There are millions of things you have to catch. Goes to say that I always thought checked exceptions in Java are just evil. There is no point to having them. 

I'll see if I can massage it to be a warning.

leleuj:
Thanks for all your contributions

leleuj:
Looks like Findbugs is trying to drive you mad ;-)

mmoayyed:
On error at a time :) Eventually it will pass. Good way to relax and do something while the build goes through.

mmoayyed:
Hooray. We are back in business finally!

leleuj:
Excellent! Merging...

PR #666, opened 2016-09-08
Fix issue: The attribute value of SAML profile is always a list (http…

ruochao:
…s://github.com/pac4j/pac4j/issues/652)

### DIFF
--- a/pac4j-saml/src/test/java/org/pac4j/saml/client/SAML2ProfileTest.java
+++ b/pac4j-saml/src/test/java/org/pac4j/saml/client/SAML2ProfileTest.java
@@ -1,0 +1,51 @@
+package org.pac4j.saml.client;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.pac4j.core.context.Pac4jConstants;
+import org.pac4j.core.profile.Gender;
+import org.pac4j.saml.profile.SAML2Profile;
+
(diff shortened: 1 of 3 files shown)


pac4j-saml/src/test/java/org/pac4j/saml/client/SAML2ProfileTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.pac4j.core.context.Pac4jConstants;
+import org.pac4j.core.profile.Gender;
+import org.pac4j.saml.profile.SAML2Profile;
+
+/**
+ * Generic test for {@link SAML2Profile}
+ */
```
leleuj: Can you just add your name in an `@author` annotation and the `@since 1.9.2`?

pac4j-saml/src/test/java/org/pac4j/saml/client/SAML2ProfileTest.java
```diff
@@ -0,0 +1,51 @@
...
+public class SAML2ProfileTest {
```
leleuj: Can you rename the class with `Tests` at the end? Otherwise, the test won't be played at build: https://github.com/pac4j/pac4j/blob/master/pom.xml#L193

leleuj:
A few comments, but it looks good to me. Thanks

PR #669, opened 2016-09-13
SAML: Fix issue with "no private key found"

mmoayyed:


leleuj:
Thanks. I will re-test the SAML support.

PR #670, opened 2016-09-14
Fix SAML issues in files management before the 1.9.2 release

leleuj:


pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -83,7 +83,7 @@ public static boolean areEquals(final String s1, final String s2) {
     public static boolean areEqualsIgnoreCaseAndTrim(final String s1, final String s2) {
         if (s1 == null && s2 == null) {
             return true;
-        } else if (s1 != null && s2 !=  null) {
+        } else if (s1 != null && s2 != null) {
             return s1.trim().equalsIgnoreCase(s2.trim());
         } else {
```
mmoayyed: No big deal, but a superfluous else statement

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -236,102 +236,156 @@ public static String toString(final Class<?> clazz, final Object... args) {
...
+    /**
+     * Add a slash at the beginning of a path if missing.
+     *
+     * @param path the path
+     * @return the completed path
+     */
+    protected static String startWithSlash(final String path) {
+        if (!path.startsWith("/")) {
+            return "/" + path;
+        } else {
```
mmoayyed: Same here

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -236,102 +236,156 @@ public static String toString(final Class<?> clazz, final Object... args) {
...
+            try {
+                return new FileInputStream(path);
+            } catch (FileNotFoundException e) {
+                throw new TechnicalException(e);
+            }
+        }
+
+        switch (prefix) {
+            case RESOURCE_PREFIX:
+                // The choice here was to keep legacy behavior and remove / prior to
```
mmoayyed: Turn this into javadocs maybe?

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -236,102 +236,156 @@ public static String toString(final Class<?> clazz, final Object... args) {
...
+        return new WritableResource() {
 
             @Override
             public File getFile() {
-                return new File(filePath);
+                final String filename = getFilename();
+                if (filename != null) {
+                    return new File(filename);
+                }
+                return null;
```
mmoayyed: Log this condition.

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -236,102 +236,156 @@ public static String toString(final Class<?> clazz, final Object... args) {
...
+                return filename;
+            }
+
+            @Override
+            public boolean exists() {
+                final File f = getFile();
+                if (f != null) {
+                    return f.exists();
+                }
+                return true;
```
mmoayyed: Shouldn't this be false? This is odd. If we don't have a file, we assume it exists? Are you assuming this a stream?

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
leleuj: In fact, this check is necessary. Let's take an example: s1 = "me", s2 = null;

s1 == null && s2 == null -> false -> else case -> s2.trim() generates a NullPointerException

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
leleuj: You would remove the `else` clause to directly have `return path;`, wouldn't you?

I think it's clearer this way, but it's just a personal feeling.

mmoayyed: Yes, return immediately exists. I'd also guess the compiler will remove that else for you automatically!

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
leleuj: Yes, I will do. Hard to find this comment.

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
leleuj: OK

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
leleuj: The `exists` method is based on the `getFile()` method which is based on `getFilename()` method. `null` filename -> `null File`

In the `getFilename`, if it's a file (no prefix, `resource:` or `classpath:` prefix), we return the real filename otherwise `null`.

So in the `exists` method, if we have a `null File`, it means that it's not a fille, but a URL resource (starting with `http:` or `https:`) so we assume it always exists. I will add a comment to make that clear.

mmoayyed: There is a lot of coupling between these methods. 

How about this:

If file -> check to see if it exists
If resource -> check to see if stream is not null and probably has a length
If anything else -> false

leleuj: "If file -> check to see if it exists": to do that, we would need to take the prefix into account (3 use cases) and we would have almost the same logic in the `getFilename()` method as well. That's why I preferred to make the `getFile()` method depends on the `getFilename()` method.

mmoayyed: Yikes. OK. Lets proceed.

leleuj:
All comments have been taken into account. Merging...

PR #685, opened 2016-09-22
Fix and improve CSRF support

### LINKED ISSUE #673: Fix and improve CSRF support
**Bug**: the `CsrfAuthorizer` seems to consider not having a sessionToken (== null) as authorized if either parameter or header token are missing (== null). We must check the sessionToken.

**Improvement**: the `CsrfTokenGeneratorAuthorizer` sets the CSRF cookie on the current path of the URL while this must be some shared path (by default /). The domain, path, isSecure and isHttpOnly properties should be settable.

leleuj:
Closes #673


othrayte:
This all looks like what we would need; I can't test it right now as my code would need to be migrated to pac4j master and I won't have time to do that until at least Tuesday.

I did notice that after these changes there isn't a way to get the old behavior of using the context's server name as the cookies domain. This wouldn't be an issue for us but might be for someone wants to retain the old behavior.

leleuj:
Yes, you're right, I just restored the old behaviour for backward compatibility: https://github.com/pac4j/pac4j/pull/685/commits/a9ba80712bdc98d450b5b5bc2304b64c99adbd31

I'll merge it soon so that you can test it from master.

leleuj:
Merged

PR #686, opened 2016-09-22
JWT is not checking expiration time

### LINKED ISSUE #672: JWT is not checking expiration time
It has JwtConstants.EXPIRATION_TIME, but seems it doesn't check for it.

leleuj:
Closes #672

### DIFF
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
@@ -28,10 +28,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.text.ParseException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 /**
  * Authenticator for JWT. It creates the user profile and stores it in the credentials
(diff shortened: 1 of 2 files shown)


pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -202,11 +199,19 @@ protected void createJwtProfile(final TokenCredentials credentials, final JWT jw
...
         if (!subject.contains(CommonProfile.SEPARATOR)) {
             subject = JwtProfile.class.getName() + CommonProfile.SEPARATOR + subject;
         }
 
+        final Date expirationTime = claimSet.getExpirationTime();
+        if (expirationTime != null) {
+            final Date now = new Date();
+            if (expirationTime.before(now)) {
+                logger.error("The JWT is expired: no profile is built");
+                return;
```
ruochao: Do you think throw an exception is better? e.g. TokenExpiredException (extends TechnicalException)

leleuj: Good question, I hesitated on this.

If you throw a `TechnicalException`, the login flow will be broken and a 500 error page will be returned. Otherwise, returning `null` or throwing a `CredentialsException` will return a 401 error page.

I tried to organize the error cases in two main categories:

- if it's an expected issue (valid JWT), I want to return a 401 error page: no appropriate encryption / signature configuration, expired token

- if it's an unexpected issue, I want to return a 500 error page: no `sub` claim, not a valid token.

Though, I don't have a strong feeling on this.

ruochao: It's really hard to say which case is better, I think it depends on how user use this token. So for my project, I just catch the TechnicalException in custom filter and return 401. I think if a token has been changed (invalid), then it should not pass the authentication, as well as the expiration.

leleuj: I will merge it as is and we'll revisit that for the 2.0.0 version if necessary.

PR #689, opened 2016-09-26
Cookies also have a version number.

victornoel:
Hi,

Pac4j implementation of cookies was missing the version field.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/context/J2EContext.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/context/J2EContext.java
@@ -202,6 +202,7 @@ public class J2EContext implements WebContext {
         if (cookies != null) {
             for (final javax.servlet.http.Cookie c : cookies) {
                 final Cookie cookie = new Cookie(c.getName(), c.getValue());
+                cookie.setVersion(c.getVersion());
                 cookie.setComment(c.getComment());
                 cookie.setDomain(c.getDomain());
                 cookie.setHttpOnly(c.isHttpOnly());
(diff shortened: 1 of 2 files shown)


leleuj:
+1

This change, of course, does not bring anything new until the specific web contexts implement it.

victornoel:
Ah yes, you are right, I forgot to do so ^^ I will add what is needed!

victornoel:
@leleuj here it is for the pac4j repository.

pac4j-core/src/main/java/org/pac4j/core/context/J2EContext.java
```diff
@@ -202,6 +202,7 @@ public String getFullRequestURL() {
         if (cookies != null) {
             for (final javax.servlet.http.Cookie c : cookies) {
                 final Cookie cookie = new Cookie(c.getName(), c.getValue());
+                cookie.setVersion(c.getVersion());
```
leleuj: Is version 0 the default value? Just to be sure we don't break anything for those who use cookies.

pac4j-core/src/main/java/org/pac4j/core/context/J2EContext.java
```diff
@@ -217,6 +218,7 @@ public String getFullRequestURL() {
     @Override
     public void addResponseCookie(final Cookie cookie) {
         final javax.servlet.http.Cookie c = new javax.servlet.http.Cookie(cookie.getName(), cookie.getValue());
+        c.setVersion(cookie.getVersion());
```
leleuj: The question fits here best: is version 0 the default value? Just to be sure we don't break anything for those who use cookies.

leleuj: Perfect

PR #692, opened 2016-09-27
Missing initialisation of Clients in findClients

victornoel:
`findClient(WebContext)` is missing the initialisation of the clients BEFORE the default client is returned.

I think there is also undefined behaviour if the default client is NOT in the clients collection, but this PR does not cover this problem yet.
If you have an idea how to better handle that case, I can add some commits for it :)


leleuj:
Good catch!

Issue #695, opened 2016-09-29
Improve logging in case of Authentication failed


victornoel:
Hi,

I noticed that when implementing an `Authenticator`, the only way it had to express that the credentials are not valid is by throwing a `BadCredentialsException`.

I also noticed that because of that, this exception is caught somewhere later and an error is logged like so:
```
ERROR [2016-09-29 15:47:46,367] org.ow2.petals.cockpit.server.security.CockpitAuthClient: Failed to retrieve or validate credentials
! org.pac4j.core.exception.BadCredentialsException: Bad credentials for: wrong
! at org.ow2.petals.cockpit.server.resources.MockAuthenticator.validate(UserSessionTest.java:154)
! at org.ow2.petals.cockpit.server.resources.MockAuthenticator.validate(UserSessionTest.java:1)
! at org.pac4j.core.client.IndirectClientV2.retrieveCredentials(IndirectClientV2.java:49)
! at org.pac4j.core.client.IndirectClient.getCredentials(IndirectClient.java:105)
! at org.pac4j.core.engine.DefaultCallbackLogic.perform(DefaultCallbackLogic.java:77)
...
```

I don't think it makes sense to log an error every time an user enter a wrong password.

I guess a solution would be to catch this particular exception and log a debug statement instead?

leleuj:
Yes, in the `IndirectClientV2` and `DirectClientV2` classes, the log can be turned into an `INFO`.

Would you mind handling that?

victornoel:
I think it's alright for this issue now, thanks!

leleuj:
We still have this PR pending for 2.0: https://github.com/pac4j/pac4j/pull/696

victornoel:
@leleuj it is not related I think, #696 is more about improving the use of the Exception and ensuring everything is is logged as it should… but this particular issue is ok with me at least :)
Well, we can still use this issue as a placeholder for discussion, but then the description should be updated.

Issue #622, opened 2016-08-06
Migrate clients to V2


leleuj:
Current clients are based on the `DirectClient` and `IndirectClient` base classes which require to override the `redirect`, `getCredentials` and `getUserprofile` methods. This leads to heavy classes hierarchy.
New clients are based on the `DirectClientV2` and `IndirectClientV2` base classes which use the `ActionRedirectBuilder`, `CredentialsExtractor`, `Authenticator`and `ProfileCreator` subcomponents. This leads to a more modular codebase.

The `IndirectClient` and `DirectClient` will me marked `@Deprecated`

To be migrated: OAuth, SAML, GAE, OpenID

PR #696, opened 2016-09-30
CredentialException handling improvements

victornoel:
EDIT: see comments below, this has been rebased on top of #697 

A fix for #695

It became MUCH bigger than what I expected at first.

The thing is that I found myself needing to differentiate between expect exception (such as bad formatted client request, invalid credentials, missing information, unexisting accounts) and unexpected exception (such as multiple accounts in the db, problem connecting to the remote authenticator provider, etc).

So I made `CredentialsException` extends Exception instead of `TechnicalException`, I made `MultipleAccountsFoundException` extends `TechnicalException`, and then I had to make `OAuthCredentialsException` extends `TechnicalException` too because the OAuth client is not a `IndirectClientV2` and does not use the `Authenticator` abstractions.

Then because `CredentialsException` became a checked exception, I had to change a lot of things in many places, mostly tests.

I took this opportunity to add a few improvements in the `LocalCachingAuthenticator` so that it rethrow the original exception and not the cache's `ExecutionException`.
I also improved (but please check it is correct) the `JwtAuthenticator` to throw a `CredentialsException` instead of a `TechnicalException` when the JWT token can't be parsed, because it is expected that client could send wrong data. The opposite happened with the `StormpathAuthenticator` that was throwing a `BadCredentialsException` when it had a problem connecting to Stormpath.

This changes the API of `Authenticator` because now it can throw 2 checked exception in validate. The same for `CredentialsExtractor` (maybe in its case, an exception for bad format could be introduced instead of using the generic `CredentialsException` by the way...).

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/exception/CredentialsException.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/exception/CredentialsException.java
@@ -1,15 +1,15 @@
 package org.pac4j.core.exception;
 
 /**
- * This class represents an exception occuring during credentials retrieval.
+ * This class represents an expected exception occurring during credentials retrieval.
  * 
  * @author Jerome Leleu
  * @since 1.4.0
  */
-public class CredentialsException extends TechnicalException {
-    
-    private static final long serialVersionUID = 8188990220217650629L;
-    
+public class CredentialsException extends Exception {
+
+    private static final long serialVersionUID = 6013115966613706463L;
+
     public CredentialsException(final String message) {
         super(message);
     }
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
@@ -1,13 +1,14 @@
 package org.pac4j.jwt.credentials.authenticator;
 
-import com.nimbusds.jose.*;
-import com.nimbusds.jwt.EncryptedJWT;
-import com.nimbusds.jwt.JWT;
-import com.nimbusds.jwt.JWTClaimsSet;
-import com.nimbusds.jwt.JWTParser;
-import com.nimbusds.jwt.PlainJWT;
-import com.nimbusds.jwt.SignedJWT;
+import java.text.ParseException;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 import org.pac4j.core.context.WebContext;
+import org.pac4j.core.credentials.TokenCredentials;
 import org.pac4j.core.credentials.authenticator.Authenticator;
 import org.pac4j.core.exception.CredentialsException;
 import org.pac4j.core.exception.HttpAction;
@@ -16,10 +17,9 @@ import org.pac4j.core.profile.CommonProfile;
 import org.pac4j.core.profile.ProfileHelper;
 import org.pac4j.core.profile.creator.AuthenticatorProfileCreator;
 import org.pac4j.core.util.CommonHelper;
-import org.pac4j.core.credentials.TokenCredentials;
 import org.pac4j.jwt.JwtClaims;
-import org.pac4j.jwt.config.encryption.SecretEncryptionConfiguration;
 import org.pac4j.jwt.config.encryption.EncryptionConfiguration;
+import org.pac4j.jwt.config.encryption.SecretEncryptionConfiguration;
 import org.pac4j.jwt.config.signature.SecretSignatureConfiguration;
 import org.pac4j.jwt.config.signature.SignatureConfiguration;
 import org.pac4j.jwt.profile.JwtGenerator;
@@ -27,8 +27,17 @@ import org.pac4j.jwt.profile.JwtProfile;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.text.ParseException;
-import java.util.*;
+import com.nimbusds.jose.EncryptionMethod;
+import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jose.JWEAlgorithm;
+import com.nimbusds.jose.JWEHeader;
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.jwt.EncryptedJWT;
+import com.nimbusds.jwt.JWT;
+import com.nimbusds.jwt.JWTClaimsSet;
+import com.nimbusds.jwt.JWTParser;
+import com.nimbusds.jwt.PlainJWT;
+import com.nimbusds.jwt.SignedJWT;
 
 /**
  * Authenticator for JWT. It creates the user profile and stores it in the credentials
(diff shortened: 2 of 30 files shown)


leleuj:
Wow! It's a too big change for a minor version, it has breaking changes (signature changes).

Currently, the only checked exception is the `HttpAction` used to "break the login process".

All other exceptions are unchecked ones, starting with the `TechnicalException`. And the `CredentialsException` is just a specific subclass to indicate something wrong has happened during credentials validation.

I guess it's clearer to say that an `Authenticator` can throw a `CredentialsException`when credentials validation fails (it's expected), but I'm a bit reluctant (probably wrongly) to do this change.

This needs to be discussed for the 2.0 roadmap. I just added it to the *pac4j roadmap*. 

Indeed, there might be some consistencies between the `JwtAuthenticator` and the `StormpathAuthenticator` and in the `LocalCachingAuthenticator`.

millross:
Will the next release be 2.0.x rather than 1.10.x? Either way I agree, if we're breaking APIs then it needs to be targeted for 2.0.x  rather than 1.9.x

That said, maybe we could start a fork branch for exactly this now to get the breaking changes in there given that the PR already exists? Or are you keen to avoid doing that till we have a roadmap?

leleuj:
The next major version will be the 2.0.0 and the next major one the 3.0.0 because I planned to use the semver versioning and make people easily aware of what are the major versions. It's still to be discussed, though.

For a long time, I have avoided double work on master and maintenance branch by doing many (breaking) changes on master, but it's less and less acceptable as pac4j has become more and more popular. I read this tweet by the creator of Javaslang: https://twitter.com/danieldietrich/status/781576244903153664
So I plan to organize a conf call next week or the following week and then create a 1.9.x branch and move the master in version 2.0.0 (and merge maybe this PR) when we'll all agree about what we want to achieve for 2.0.

millross:
It's also worth spending time discussing the pros and cons of master plus maintenance branch. We can of course do some clever cherry pick stuff to help with the overhead but we may also have to accept some overhead. 

I don't know how many production users there are of pac4j, or if we can maybe offer upgrade guides for compatibility breaks we make when we shouldn't (or can we offer default impls in some cases to retain compat?) I'm currently available to chat via Skype next week till Thursday (feel free to ping me to ask for Skype id but I'd rather avoid posting it on a github chat :)). Personally I think your plan is the right one, but I'm open to persuasion by others who have a different view.

victornoel:
@leleuj I agree :) actually I submitted the PR to see how you would take it, so I will submit another one that simply replace the error by an info, and as much as possible changing some stuffs (without breaking any API) to ensure no important error message is lost in the process :)

victornoel:
Ok, so in the end I submitted #697 and rebased the current PR on top of it.

It's now only making CredentialsException becoming a checked Exception, I almost think this could be thrown out and maybe we could add to the roadmap for the v2 that CredentialsException should become a checked Exception (and maybe also introducing another different exception for badly formatted authenticator request for the CredentialExtractor).

victornoel:
I updated the PR and put all the breaking changes in it. It is till based upon #697 :)

victornoel:
It's now independent from #697.

leleuj:
During our v2 technical meeting, we weakly agree on making credentials validation based on a checked exception. I'm resuming the review of this PR.

First of all, the conflicts must be fixed.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -103,14 +103,14 @@ public CommonProfile validateToken(final String token) {
         final TokenCredentials credentials = new TokenCredentials(token, "(validateToken)Method");
         try {
             validate(credentials, null);
-        } catch (final HttpAction e) {
+        } catch (final HttpAction | CredentialsException e) {
```
leleuj: I disagree: `CredentialsException` are expected exceptions should never turn into unexpected ones (`TechnicalException`). You should return a `null` profile in that case.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -188,7 +188,7 @@ public void validate(final TokenCredentials credentials, final WebContext contex
           	createJwtProfile(credentials, jwt);
 
         } catch (final ParseException e) {
-            throw new TechnicalException("Cannot decrypt / verify JWT", e);
+            throw new CredentialsException("Cannot decrypt / verify JWT", e);
```
leleuj: OK. This changes the meaning: unparseable JWT are now considered just as normal credentials validation failures as others.

victornoel:
Ok, I will update the PR as soon as I have some time :)

leleuj:
An overall +1, but a few comments to address before the merge.

pac4j-core/src/main/java/org/pac4j/core/exception/CredentialsException.java
```diff
@@ -1,12 +1,12 @@
...
 
 /**
- * This class represents an exception occuring during credentials retrieval.
+ * This class represents an expected exception occurring during credentials retrieval.
  * 
  * @author Jerome Leleu
  * @since 1.4.0
  */
-public class CredentialsException extends TechnicalException {
+public class CredentialsException extends Exception {
```
victornoel: ok

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
victornoel: ok

pac4j-oauth/src/main/java/org/pac4j/oauth/exception/OAuthCredentialsException.java
```diff
@@ -2,15 +2,15 @@
...
+import org.pac4j.core.exception.TechnicalException;
 
 /**
  * This class represents an exception occurring during OAuth credentials retrieval.
  * 
  * @author Jerome Leleu
  * @since 1.3.0
  */
-public class OAuthCredentialsException extends CredentialsException {
+public class OAuthCredentialsException extends TechnicalException {
```
leleuj: In fact, the internal `retrieveCredentials` method plays the same role as the `Authenticator`  and will soon be replaced by it (https://github.com/pac4j/pac4j/issues/622). So I think it makes sense.

victornoel: ok

victornoel:
Here it is, rebased and reviews taken into account.

leleuj:
Great! Thanks. Though, it looks like this is a compilation failure.

victornoel:
@leleuj ah yes, that's also why I kept `OAuthCredentialsException` extending `TechnicalException`: if not we have to modify the `OAuthClient`s to handle it since it became checked and then it means modifying `BaseClient`, and so on…

leleuj:
So my advice was a bad one. Revert that. Will handle this change when moving the OAuth clients to the V2 model.

PR #697, opened 2016-10-01
Logging improvements

victornoel:
Here is a less complex fix for #695 than #696 :)

I still made some changes on some of the other exceptions, but everything stayed runtime exceptions.

Everything is explained in the commit messages, it is similar to #696 in terms of behaviour, I made some improvements here and there.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
@@ -38,7 +38,9 @@ public abstract class DirectClientV2<C extends Credentials, U extends CommonProf
             this.authenticator.validate(credentials, context);
             return credentials;
         } catch (CredentialsException e) {
-            logger.error("Failed to retrieve or validate credentials", e);
+            logger.info("Failed to retrieve or validate credentials: {}", e.getMessage());
+            logger.debug("Failed to retrieve or validate credentials", e);
+
             return null;
         }
     }
--- a/pac4j-core/src/main/java/org/pac4j/core/client/IndirectClientV2.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/IndirectClientV2.java
@@ -49,7 +49,9 @@ public abstract class IndirectClientV2<C extends Credentials, U extends CommonPr
             this.authenticator.validate(credentials, context);
             return credentials;
         } catch (CredentialsException e) {
-            logger.error("Failed to retrieve or validate credentials", e);
+            logger.info("Failed to retrieve or validate credentials: {}", e.getMessage());
+            logger.debug("Failed to retrieve or validate credentials", e);
+
             return null;
         }
     }


pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
```diff
@@ -38,7 +39,14 @@ protected C retrieveCredentials(final WebContext context) throws HttpAction {
             this.authenticator.validate(credentials, context);
             return credentials;
         } catch (CredentialsException e) {
-            logger.error("Failed to retrieve or validate credentials", e);
+            if (logger.isDebugEnabled()) {
```
mmoayyed: If debug id enabled, you're logging at info?!

Why is this check even necessary?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClientV2.java
```diff
@@ -49,7 +50,14 @@ protected C retrieveCredentials(final WebContext context) throws HttpAction {
             this.authenticator.validate(credentials, context);
             return credentials;
         } catch (CredentialsException e) {
-            logger.error("Failed to retrieve or validate credentials", e);
+            if (logger.isDebugEnabled()) {
```
mmoayyed: See above

pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/LocalCachingAuthenticator.java
```diff
@@ -56,8 +59,30 @@ public void validate(final T credentials, final WebContext context) throws HttpA
             });
             credentials.setUserProfile(profile);
             logger.debug("Found cached credential. Using cached profile {}...", profile);
-        } catch (Exception e) {
-            throw new CredentialsException(e);
+        } catch (UncheckedExecutionException e) {
+            if (e.getCause() instanceof RuntimeException) {
+                throw (RuntimeException) e.getCause();
+            } else if (e.getCause() != null) {
```
mmoayyed: Note that all such `else` clauses are redundant. You're typing too much! :)

pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
victornoel: I always want to log an info, but if in debug, I want the full stacktrace of the exception and not only its message.

pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/LocalCachingAuthenticator.java
victornoel: The cache implementation can throw one of these 3 exceptions to wrap the original ones (the cause) OR for one of its own error (no cause).
The point of this change is to extract the original cause (from delegate.validate) and throw it instead of the wrapper.
The only optimisation I can see is to extract the getCause null check and TechnicalException throwing into a method.
But I don't see what is redundant. If you have a better way of expressing that, please share :)

victornoel:
Thanks for your feedbacks @mmoayyed :)

I answered inline. As for the breaking change, why would you say so? I didn't change the semantics of the exceptions, only made them clearer and fixed some places where that semantics wasn't respected and would lead to unexpected behaviour.

victornoel:
(the only thing that could be considered a break is the change to OAuthCredentialsException, but I think it was an error to have made it extend CredentialsException (for the reasons explained in the commit).
I don't think anybody would rely on that anyway, it seems used only internally in pac4j, but I can revert it :)

victornoel:
And maybe the same for MultipleAccountsFoundException, but I wouldn't like it to be logged as info but as error, so we would need one for catch in the clients for this special case until it becomes a real technical exception.

victornoel:
Actually, if you prefer, we can throw away the while first commit and add catch/instanceof hacks in the logging part to have the expected logging behaviour. It's just that I feel like it's not worth it :)

pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
mmoayyed: Cool, but you're always logging at `loger.info`? Shouldnt one of these calls be `logger.debug`?

pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/LocalCachingAuthenticator.java
mmoayyed: Right, but your explanation doesn't answer my question, that the else clause is redundant:

```
if (x)
	throw a;

if (y)
	throw b;

throw c;

```

mmoayyed:
It's a breaking change, (I suspect),because you're changing the signature of the exception thrown right? If I was catching X before, do I need to now adjust that block to catch Y? If so, that is a breaking change.

pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
victornoel: Nop, it's on purpose, because I don't want it to be a debug statement!

Another option would have been to have the info statement and then add a debug statement with the exception, but I don't like to have two log statement for the same information, so I usually change what is exactly logged (here adding the whole stacktrace instead of the message) depending on the configured level.

pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/LocalCachingAuthenticator.java
victornoel: ha... isn't that simply cosmetics? I mean... it's a problem of style, not of correctness...

victornoel:
@mmoayyed concerning MultipleAccountsFoundException and OAuthCredentialsException, yes, you can't catch them anymore by catching CredentialsException, but who would want to do that? In the first case, it's not a problem of invalid credential, it's a problem of a buggy database data. And in the second case, who would expect to get a CreentialsException from something that is not part of the Authenticator and CredentialsExtractor framework (which are the one that are documented to throw such an exception).

Still, as I said, we can revert these choices, but care should be taken to log MultipleAccountsFoundException as an error and not as an info.

As for the throwing changes I made in JwtAuthenticator and StormpathAuthenticator, they were bugs in my opinion.

mmoayyed:
The question isn't about who would *want* that today. It's about who *did* that yesterday. Behavior X was provided by version A. Behavior X was changed by version A. Regardless of how "bogus" behavior X is now, it is something people may be doing/using. Changing that is the very definition of breaking BC; and is considered a regression unless you target it for a relevant release. It's not what one can expect. It's about what one does expect, unless you can guarantee that the user-facing code does not change or "break", or else we target this for a 2.0 type of release to indicate the change in behavior.

pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/LocalCachingAuthenticator.java
mmoayyed: A problem of style is a problem of correctness.

victornoel: The only reason I would argue against or pro the else in that place is the following: without the else, maintainability is threatened. It is easier to introduce bugs by modifying this code if there is no explicit structure about what happens and when. It is the same thing as when removing the braces around the argument of a `if`: there is much more chance to add a statement that the developer think will be executed as part of the `if` while actually it won't.
Now, I really think there is no reason to spend more time discussing that, so I will simply follow your recommendations and be done with it :)

mmoayyed: To clarify, it's a matter of consistency and keeping the codebase lean. It would pose a problem longterm if we ended up repeating things, repeaing things repeaing things, that makes it harder to read the code read the code read the code read the code :)

mmoayyed:
Btw I am certainly not saying you should revert this. Not at all. I am saying this isn't something I'd expect from a patch release.

victornoel:
I know, I will simply update this PR so that I get a fix for #695 that can enter 1.9.x.

pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/LocalCachingAuthenticator.java
victornoel: FYI @mmoayyed https://www.imperialviolet.org/2014/02/22/applebug.html

That's the kind of reason why I use `else` and braces.

victornoel:
@mmoayyed what is your opinion about Jwt and Stormpath Authenticators bugfixes? Are they also breaking changes?

victornoel:
Here is the new PR on the light of the discussion, there should be no breaking changes at all now.

mmoayyed:
I suspect they are, yes. Anything that affects user-facing code, specially if it's something expected/available to users to extend/modify falls under that category.

pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
mmoayyed: Gotcha. I am ambivalent either way, but to make things simpler, I'd simply stick to a single debug statement and call it a day.

pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/LocalCachingAuthenticator.java
mmoayyed: I take your point completely. There is inherent danger in "forgetting" to put in the proper return statements. No doubt. Else clauses help. Unit tests do too! My comment isn't so much about which is better; it's more about sticking to one single consistent style across the board. In general, it's lots easier to say "This is our style, deal with it" rather than launching into some debate about the metrics of compiler efficiency, etc :) I simply advocate for a style that helps me do/write less. :D

@leleuj that might be something we could put into our styling rules here; Pick one and apply across the board.

victornoel: my last commit should satisfy you :)

victornoel:
ok, my last commit does not include them, so it's all good now.

As a side discussion (just for the sake of it :), the two fixes in the Authenticators are only being used by the pac4j framework, they are not SPI that are extended by users or API used by users. Moreover these changes will only produce the same result as before, except that they will correctly be logged w.r.t. the changes in the clients that are present in this PR.
Well, anyway, now they are in #696 with all the other breaking change modifications.

pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
victornoel: I will do that

victornoel: Well... do we agree @mmoayyed that it is what you had in mind:
```java
            logger.debug("Failed to retrieve or validate credentials", e);
            logger.info("Failed to retrieve or validate credentials: {}", e.getMessage());
```
Or was it:
```java
            if (logger.isDebugEnabled()) {
                logger.debug("Failed to retrieve or validate credentials", e);
            } else {
                logger.info("Failed to retrieve or validate credentials: {}", e.getMessage());
            }
```

In the first case it feels really strange to get two logs saying almost the same thing and in the second case, it's enough to have DEBUG logs redirected to another place than INFO logs (in the logs configuration) for the information to be easily missed...

Please tell me what you prefer :)

leleuj:
@mmoayyed @victornoel are we good on this one for the merge?

victornoel:
I am (except maybe for the debug statement… see with @mmoayyed if the current state is ok or if he wants to choose between the two options I gave him).

Also, I would be interested in your point of view @leleuj  regarding the two thrown Exception bugfixs for Stormpath and Jwt Authenticators: do we keep them in #696 or would you accept them for 1.9.x? (in that case I will make another PR so that we can simply merge this one).

leleuj:
I'm a bit biased because lately, I have received several inquiries about why things are broken here and there (when upgrading from 1.9.1 to 1.9.2).

About the changes regarding JWT and Stormpath, I'm not sure these are obvious bugs, more a lack of consistency. I would keep it for 2.0.0 within your bigger PR.

pac4j-core/src/main/java/org/pac4j/core/client/DirectClientV2.java
```diff
@@ -38,7 +39,18 @@ protected C retrieveCredentials(final WebContext context) throws HttpAction {
             this.authenticator.validate(credentials, context);
             return credentials;
         } catch (CredentialsException e) {
-            logger.error("Failed to retrieve or validate credentials", e);
+            if (e instanceof MultipleAccountsFoundException) {
+                // MultipleAccountsFoundException is a subclass of CredentialsException even though
```
mmoayyed: Turn this into javadocs plz.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClientV2.java
```diff
@@ -49,7 +50,18 @@ protected C retrieveCredentials(final WebContext context) throws HttpAction {
             this.authenticator.validate(credentials, context);
             return credentials;
         } catch (CredentialsException e) {
-            logger.error("Failed to retrieve or validate credentials", e);
+            if (e instanceof MultipleAccountsFoundException) {
+                // MultipleAccountsFoundException is a subclass of CredentialsException even though
```
mmoayyed: Same here

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClientV2.java
```diff
@@ -49,7 +50,18 @@ protected C retrieveCredentials(final WebContext context) throws HttpAction {
...
+                // it is not caused by an authentication error but by a a bug, so we consider it an error
+                throw e;
+            }
+            
+            if (logger.isDebugEnabled()) {
```
mmoayyed: This entire block appears to be duplicated here. Is there a way we could remove that duplication?

mmoayyed:
It's clear that things wrt exception handling needs to be reworked. While this might be OK for the short-term, the whole block of if-else-then-throw-but-whatif-then-do-this-except-when-it's-this that is added here is just too assuming. I'd rework that entirely to remove the conditionals altogether, perhaps in 2.0.

mmoayyed:
And there should be unit tests in this PR to make sure the correct thing is thrown and that no else clauses are missed accidentally :)

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClientV2.java
victornoel: I don't think it is the scope of this PR, DirectClientV2 and IndirectClientV2 both contains a lot of duplication already, I don't want to tackle that here, but if you do another PR, I can rebase this one on top of it.

victornoel:
Here you have it, I simply changed the error into an info.

leleuj:
Given this simple change and the long discussion, I guess we all agree on the merge.

PR #705, opened 2016-10-13
Added SecurityLogic value to the Config object with DefaultSecurityLo…

richard77:
…gic as default.

This allow us to retrieve the actual value from the Config object for instance when using the SecureAction annotation within the Pac4j-play project. Currently there is now easy way to switch securityLogic implementations.


richard77:
Removed Config object changes, going for '''@Inject''' in play-pac4j instead

richard77:
Reopening as per request.
```@inject``` Solution did not do the trick. Using Config object approach instead.

leleuj:
I don't see any changes in this PR. I guess there is an issue.

gpgekko:
@leleuj It's because the second commit in the PR reverts the changes before that

leleuj:
Fixed by: https://github.com/pac4j/pac4j/commit/044f01023e085b93c27f2c44b3be5c84f4baef94

PR #722, opened 2016-10-20
Improve exception handling in caching authenticator

victornoel:
Throw original exception or error instead of the cache implementation's exception.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/LocalCachingAuthenticator.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/LocalCachingAuthenticator.java
@@ -48,18 +48,17 @@ public class LocalCachingAuthenticator<T extends Credentials> extends Initializa
     public void validate(final T credentials, final WebContext context) throws HttpAction, CredentialsException {
         init(context);
 
-        try {
-            final CommonProfile profile = this.cache.get(credentials, () -> {
-                logger.debug("Delegating authentication to {}...", delegate);
-                delegate.validate(credentials, context);
-                return credentials.getUserProfile();
-            });
+        CommonProfile profile = this.cache.getIfPresent(credentials);
+        if (profile == null) {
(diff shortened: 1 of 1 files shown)


pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/LocalCachingAuthenticator.java
```diff
@@ -56,10 +59,34 @@ public void validate(final T credentials, final WebContext context) throws HttpA
...
+        } catch (ExecutionException e) {
+            // this covers all the checked exception of validate
+            if (e.getCause() instanceof HttpAction) {
+                throw (HttpAction) e.getCause();
+            } else {
+                throw rethrowWrapper(e);
+            }
+        } catch (UncheckedExecutionException e) {
+            if (e.getCause() instanceof RuntimeException) {
+                throw (RuntimeException) e.getCause();
```
victornoel: (Note that normally you can simply add commits to my branch, if you want to, according to the way github works)

leleuj: I'll take a closer look to see if I have something simpler to propose...

victornoel: One way of simplifying a lot this code is to use `cache.getIfPresent()` and simply populate the cache by hand in case the value is not there: like this there is no need at all to manipulated the various `ExecutionException`, `UncheckedExecutionException` and `ExecutionError`.
We loose potential atomicity of the operation, but I don't think it is a problem functionally, and anyway I don't think the cache are implementing atomicity in practice either.
I can update the PR if you want :)

leleuj: Yes please, it seems a good solution...

victornoel:
@leleuj much more elegant now :)

leleuj:
Yes, a lot better. :+1:

PR #723, opened 2016-10-20
Logout OK should be a 204

### LINKED ISSUE #701: Logout OK should be a 204
Hi,

The implementation of ApplicationLogoutLogic returns a 200 when there is no redirect configured.

Since there is no content in the answer, the OK should be a 204 and not a 200.

leleuj:
Closes #701

### DIFF
--- a/pac4j-cas/src/test/java/org/pac4j/cas/client/CasClientTests.java
+++ b/pac4j-cas/src/test/java/org/pac4j/cas/client/CasClientTests.java
@@ -127,7 +127,7 @@ public final class CasClientTests implements TestsConstants {
         final MockWebContext context = MockWebContext.create().addRequestParameter(CasConfiguration.LOGOUT_REQUEST_PARAMETER, LOGOUT_MESSAGE)
             .setRequestMethod("POST");
         TestsHelper.expectException(() -> casClient.getCredentials(context), HttpAction.class, "back logout request: no credential returned");
-        assertEquals(204, context.getResponseStatus());
+        assertEquals(200, context.getResponseStatus());
     }
 
     private String deflateAndBase64(final String data) {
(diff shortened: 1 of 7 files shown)


pac4j-cas/src/test/java/org/pac4j/cas/client/CasClientTests.java
```diff
@@ -127,7 +127,7 @@ public void testBackLogout() {
         final MockWebContext context = MockWebContext.create().addRequestParameter(CasConfiguration.LOGOUT_REQUEST_PARAMETER, LOGOUT_MESSAGE)
             .setRequestMethod("POST");
         TestsHelper.expectException(() -> casClient.getCredentials(context), HttpAction.class, "back logout request: no credential returned");
-        assertEquals(200, context.getResponseStatus());
+        assertEquals(204, context.getResponseStatus());
```
leleuj: @mmoayyed I'm wondering if returning a 204 (instead of 200) on CAS logout and proxy requests will cause issues with the CAS server.

leleuj: I think it will regarding: https://github.com/apereo/cas/blob/master/core/cas-server-core-util/src/main/java/org/apereo/cas/util/http/SimpleHttpClientFactoryBean.java#L74, the `HTTP_NO_CONTENT` code is not in the list of the acceptable ones.

mmoayyed: HTTP_NO_CONTENT is probably a safe one to add to CAS but let me ask: why 204? The change makes sure the logout endpoint produces a 204? for CAS?

leleuj: 204 is for Ok without content. The proposed change was for the `ApplicationLogoutLogic`, but the code is used for CAS logout / proxy as well. I guess it's safer for now to keep 200 for CAS while updating the CAS server itself.

PR #735, opened 2016-10-31
Check all JWT signing configurations before giving up

estsauver:
There are cases where a service publishes several keys that are all valid signers of JWT Tokens, for example google publishes several associated with their android sign in tool. (See https://developers.google.com/identity/sign-in/android/start-integrating for an in depth description of the flow, https://www.googleapis.com/oauth2/v3/certs is a list of the keys that they provide.)

When adding multiple signature configurations to a jwtAuthenticator, only the first signature configuration is tried.

I think this change should cause the JWT Authenticator to correctly attempt each successive configuration. This would be my first patch into Pac4J, and I would appreciate any feedback on the patch.

### DIFF
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
@@ -163,7 +163,9 @@ public class JwtAuthenticator implements Authenticator<TokenCredentials> {
                             try {
                                 verified = config.verify(signedJWT);
                                 found = true;
-                                break;
+                                if (verified) {
+                                  break;
+                                }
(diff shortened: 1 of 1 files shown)


leleuj:
It looks good to me. Though, a test could confirm the "fix": two signature configurations with the same algorithms but different keys.

I don't think we have the same issue for encryption, but a test could help as well.

Finally, you certainly want to backport that on the 1.9.x branch (another pull request) as version 1.9.5 will come sooner than version 2.0.0.

estsauver:
Can do both, but it will probably be until next week. We're launching this week, so things may be just a bit hectic.

leleuj:
No problem. Take your time.

leleuj:
I'm in line with the change. Merging. Feel free to come back on this with more feedbacks.

PR #739, opened 2016-11-07
Add a mechanism to handle attributes serialization into strings

leleuj:
@mmoayyed :

There is now an `InternalAttributeHandler`, global to all profiles which handle attributes before saving them in the map or after restoring them from the map.

In CAS, we will do: `ProfileHelper.getInternalAttributeHandler().setStringify(true);` to force attributes to be turned into strings (otherwise, nothing is required).

In that case:
A string stays a string.
A boolean becomes a "{#bool} + boolean value" String.
An integer becomes a "{#int} + integer value" String.
A long becomes a "{#long} + long value" String.
A date becomes a "{#date} + date value" String.
An URI becomes a "{#uri} + uri value" String.
Otherwise, it becomes a "{#sb64} + serialized and base64 value" String.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/InternalAttributeHandler.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/InternalAttributeHandler.java
@@ -25,13 +25,13 @@ public class InternalAttributeHandler implements Serializable {
 
     protected transient final Logger logger = LoggerFactory.getLogger(getClass());
 
-    public transient final static String PREFIX = "{#";
-    public transient final static String PREFIX_BOOLEAN = PREFIX + "bool}";
-    public transient final static String PREFIX_INT = PREFIX + "int}";
-    public transient final static String PREFIX_LONG = PREFIX + "long}";
-    public transient final static String PREFIX_DATE = PREFIX + "date}";
-    public transient final static String PREFIX_URI = PREFIX + "uri}";
-    public transient final static String PREFIX_SB64 = PREFIX + "sb64}";
+    public transient static final String PREFIX = "{#";
+    public transient static final String PREFIX_BOOLEAN = PREFIX + "bool}";
+    public transient static final String PREFIX_INT = PREFIX + "int}";
+    public transient static final String PREFIX_LONG = PREFIX + "long}";
+    public transient static final String PREFIX_DATE = PREFIX + "date}";
+    public transient static final String PREFIX_URI = PREFIX + "uri}";
+    public transient static final String PREFIX_SB64 = PREFIX + "sb64}";
 
     private JavaSerializationHelper serializationHelper = new JavaSerializationHelper();
 
@@ -48,17 +48,17 @@ public class InternalAttributeHandler implements Serializable {
             return value;
         } else {
             if (value instanceof Boolean) {
-                return PREFIX_BOOLEAN + value;
+                return PREFIX_BOOLEAN.concat(value.toString());
             } else if (value instanceof Integer) {
-                return PREFIX_INT + value;
+                return PREFIX_INT.concat(value.toString());
             } else if (value instanceof Long) {
-                return PREFIX_LONG + value;
+                return PREFIX_LONG.concat(value.toString());
             } else if (value instanceof Date) {
-                return PREFIX_DATE + newSdf().format((Date) value);
+                return PREFIX_DATE.concat(newSdf().format((Date) value));
             } else if (value instanceof URI) {
-                return PREFIX_URI + value.toString();
+                return PREFIX_URI.concat(value.toString());
             } else {
-                return PREFIX_SB64 + serializationHelper.serializeToBase64((Serializable) value);
+                return PREFIX_SB64.concat(serializationHelper.serializeToBase64((Serializable) value));
             }
         }
     }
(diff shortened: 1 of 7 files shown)


mmoayyed:
Great. I'll review by EOD! Thanks.

leleuj:
Thanks. The idea is to stay compatible with CAS with minimal efforts on both sides.

leleuj:
@mmoayyed : planning to merge tomorrow to move on.

mmoayyed:
Sorry, I realized I had not submitted comments; overall this is fine really so don't let it hold up work.

pac4j-core/src/main/java/org/pac4j/core/profile/InternalAttributeHandler.java
```diff
@@ -0,0 +1,128 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+/**
+ * Internally handles attributes (set / get).
+ *
+ * @author Jerome Leleu
+ * @since 2.0.0
+ */
+public class InternalAttributeHandler implements Serializable {
```
millross: I may be late to this party but if the idea is that this should not be treated as an external API, it might be useful for the documentation to specify that this is not part of the public API and can be removed or changed without notice, so should not generally be used by code external to pac4j (assuming we can't tweak the visibility to something other than public).

leleuj: This component was added to allow some transformation on the attributes before saving them or retrieving them. Currently, it's only for CAS, but this could be enhanced in some other ways. Tough, as it seems unlikely, it's mostly an internal component. Hence the name.

pac4j-core/src/main/java/org/pac4j/core/profile/InternalAttributeHandler.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private static final long serialVersionUID = 3391226300882369266L;
+
+    protected transient final Logger logger = LoggerFactory.getLogger(getClass());
+
+    public transient final static String PREFIX = "{#";
```
leleuj: OK. Will change.

pac4j-core/src/main/java/org/pac4j/core/profile/InternalAttributeHandler.java
```diff
@@ -0,0 +1,128 @@
...
+            } else if (value instanceof Integer) {
+                return PREFIX_INT + value;
+            } else if (value instanceof Long) {
+                return PREFIX_LONG + value;
+            } else if (value instanceof Date) {
+                return PREFIX_DATE + newSdf().format((Date) value);
+            } else if (value instanceof URI) {
+                return PREFIX_URI + value.toString();
+            } else {
+                return PREFIX_SB64 + serializationHelper.serializeToBase64((Serializable) value);
```
leleuj: It cannot be. Profiles are `Serializable` and so are the attributes which are part of it.

mmoayyed: OK. This probably is fine, but I would have preferred perhaps overloaded method so you could be a lot more explicit about the type rather than a generic object type.

pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
```diff
@@ -154,12 +156,18 @@ public String getTypedId() {
...
      */
     public Map<String, Object> getAttributes() {
-        return Collections.unmodifiableMap(this.attributes);
+        final Map<String, Object> newAttributes = new HashMap<>();
+        for (Map.Entry<String, Object> entries : this.attributes.entrySet()) {
+            final String key = entries.getKey();
+            final Object value = getAttribute(key);
+            newAttributes.put(key, value);
+        }
+        return Collections.unmodifiableMap(newAttributes);
```
leleuj: Yes, as we create a new list, adding an attribute won't affect the original attributes map. But I prefer to prohibit direct addition to making obvious that the `addAttribute` method should be used.

pac4j-core/src/main/java/org/pac4j/core/profile/InternalAttributeHandler.java
leleuj: Can you elaborate a little more? I'm not sure to follow.

pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
mmoayyed: Of course, while I can see the usage of method be problematic in general, I can't quite foresee a practical issue so this may just be moot for now, but I'd prefer to play safe specially as this PR handles CAS-related changed, and CAS certainly prefers things to be serializable friendly.

mmoayyed:
This is great. Please do merge when you get around to it.

pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
leleuj: I understand your concern and indeed, as I already create a new Map, this is not absolutely necessary.
An `unmodifiableMap` has always been returned by the `getAttributes` method and it has always worked with CAS. So I'd better keep it that way for backward compatibility, but this is something we can keep in mind and fix in a 2.0.1 at worst if necessary.

leleuj:
@mmoayyed I'm thinking of one more change. The `InternalAttributeHandler` has been attached to the `UserProfile` but it's a `static` object, common to all profiles, and this requires to serialize it along the profile (which feels a bit cumbersome). I'd better keep it in the `ProfileHelper` instead: sounds good?

PR #744, opened 2016-11-09
Refactor CAS internals support: remove Kryo

leleuj:


mmoayyed:
Has this been tested with CAS?

leleuj:
Not directly as version 2.0.0 is far from being ready.

That said, it has been tested indirectly. Currently, strings were properly serialized by Kryo, other objects needed to be declared (with some specific serializers). As all attributes will be turned into strings, there won't be any problem.

PR #746, opened 2016-11-12
Changes to allow IDC's PAC4J Extensions build against PAC4J

jkacer:
SAML2ClientConfiguration has been made non-final to allow for sub-classes. A new init() method has been introduced. It does nothing in SAML2ClientConfiguration but can be overridden in sub-classes.

SAML2Client now calls the new init() method on its configuration.

A piece of code in Clients has been extracted into a separate method. It allows for reuse in sub-classes of Clients.


leleuj:
Looks good to me. The same changes are expected for the master...

Issue #747, opened 2016-11-15
JWT UserInfo response ignored


gpgekko:
The code at [OidcProfileCreator.java, line 155](https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/creator/OidcProfileCreator.java#L155) only processes information from the `getUserInfo` accessor, but the `parse` method called above explicitly sets the `claimsSet` property to `null` if the content is a JWT. I think there should be an `else` statement there that also checks the `getUserInfoJWT` accessor.

leleuj:
Indeed, it might be some missing use case. For which OIDC provider did you encounter the issue?

Would you mind submitting a pull request for that?

gpgekko:
I'll try to get a PR up today.

I encountered this while playing around with the settings of a private Keycloak instance. The application kept breaking when I enabled JWT. After a bit of digging I noticed that claims were missing from the profile.

PR #748, opened 2016-11-16
Check if UserInfo was in a JWT

gpgekko:
Fix for issue #747 

Previously, the UserInfo response was only checking the `getUserInfo` method. However, if the response was a JWT, the content would only be set on the `getUserInfoJWT` method, which was not checked.


leleuj:
Thanks, but it does not compile (see the Travis build report).

gpgekko:
I suppose that's what happens when I test things locally and then change things after that >_<"

leleuj:
No problem. Merging...

PR #753, opened 2016-11-22
Scribe3

julioa:
-Upgrade scribe to version 3.2.0
-Make possible to use additional parameters in the authorization request.
-Avoid NPE where scope is null
-Fix error in OAuth2Client to allow use of scope when defined.
-Modify jsonhelper to allow search attributes in a path including arrays.
-Specify the verb for the requests in oauth client.


pac4j-oauth/src/main/java/org/pac4j/oauth/client/GenericOAuth20Client.java
```diff
@@ -106,12 +110,22 @@ protected GenericOAuth20Profile extractUserProfile(String body) {
...
             profile.setAttributesDefinition(attributesDefinition);
         }
-        final JsonNode json = JsonHelper.getFirstNode(body);
+        final JsonNode json = JsonHelper.getFirstNode(body, getProfileNodePath());
         if (json != null) {
             profile.setId(JsonHelper.getElement(json, "id"));
             for (final String attribute : profile.getAttributesDefinition().getPrimaryAttributes()) {
                 profile.addAttribute(attribute, JsonHelper.getElement(json, attribute));
             }
+            for (final String attribute : profile.getAttributesDefinition().getSecondaryAttributes()) {
```
leleuj: This raises an interesting about the way we find attributes, besides the Scribe upgrade. I think you lean towards path matching without going that far completely.

leleuj:
Thanks for this upgrade. You now have conflicts since I merged my latest changes.

You raise interesting side proposals: path matching for attribute retrieval and a few others.

Stay tuned. I will keep you posted about my progress on this.

julioa:
I've fixed the conflicts in the code and commit changes to my branch. Let me know if you need anything else for the code.

I agree with you in the point about attribute path and this is my opinion: Since there isn't a common standard on how to retrieve profile information we are going to find multiple implementations in the near future. There is always the change to implement a custom client but if the generic oauth client is flexible enough it would be useful in many cases without any further coding and simplify the implementation of new clients.

That's why I added the idea of a path for the root node and secondary attributes. I wanted to keep backward compatibility with current implementation so I didn't change the declaration of the attribute definition from List<String> to Map<String,String> to store the label and path of the attribute in each entry.

leleuj:
Thanks. The Travis is still broken on some OAuth 2.0 tests: https://travis-ci.org/pac4j/pac4j/builds/178301223

I'm thinking more and more that we should add the path with the attribute definition (and primary vs secondary attributes would not be necessary).

leleuj:
I don't want to waste this great work! I will merge "as is" and rebuild a new generic `OAuth10Client` and `OAuth20Client` to use for all clients.

Many thanks.

julioa:
Thanks. One more comment about the primary/secondary attributes discussion:

IMO, the main point on joining both kind of profile attributes is how compatible do you want to keep this new version to current users. To maintain compatibility, secondary attributes may be the main source of information and primary attributes may be kept as proxies to its mapped secondary attr.

Otherwise, a new way to access profile information would be a good solution. In both cases it would be nice to expose some methods in the client interface to set the attribute definitions instead of give direct access to the profile definition.

I'll keep looking your changes to see if I can help anyway in the clienteV2 implementation.

leleuj:
OK. I'll ping you when I submit the pull request for you to review it.

leleuj:
Here is the change: https://github.com/pac4j/pac4j/pull/757

You now have an `OAuth10Client` and an `OAuth20Client` (using internal components) on which all OAuth clients are built.

PR #759, opened 2016-12-01
Central logout for indirect clients

taggiasco:
Hello,

Here is my PR for my logout implementation. 

What's still missing I think : 

- the control of the logout response => should it be a different callback controller ?
- the generation of the SingleLogoutService in the SAML metadata
- the logout that would be initiated by the IdP (any idea on how to do this, not sure the IdP would get access to the user session...)
- NotOnOrAfter not set in the LogoutRequest
- the way the SessionIndex is got does not seem very well done. Any idea if the SessionIndex should be stored somewhere else ? (in the context ?)

I used the IndirectClient. But I read somewhere (in the roadmap I think) that every clients should be migrated to the IndirectClient V2. 

All comments are welcome !


leleuj:
Thanks for sharing this. I will review it.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
```diff
@@ -77,6 +90,32 @@ public final RedirectAction getRedirectAction(final WebContext context) throws H
...
+     * {@link #redirect(WebContext)} should be generally called instead.</p>
+     * <p>If an authentication has already been tried for this client and has failed (<code>null</code> credentials) or if the request is an AJAX one,
+     * an authorized response (401 HTTP status code) is returned instead of a redirection.</p>
+     *
+     * @param context context
+     * @return the redirection action
+     */
+    public final RedirectAction getLogoutRedirectAction(final WebContext context) {
+        // it's an AJAX request -> unauthorized (instead of a redirection)
+        if (ajaxRequestResolver.isAjax(context)) {
```
leleuj: These checks are not necessary for logout.

leleuj:
Yes, all clients will be migrated to the `IndirectClientV2` model: it means they use internal components (`RedirectActionBuilder`, `CredentialsExtractor`, `Authenticator` and `ProfileCreator`)  instead of relying on methods overriding. The `SAML2Client` still needs to be migrated. That said, it doesn't change much here.

I see your solution, here is mine (in progress, but the main parts are done): https://github.com/pac4j/pac4j/pull/760

To handle multi profiles (and thus multiple logout), I took a back-channel approach (direct calls from the application to the identity provider), but I realize this won't work as I don't have the identity providers cookies.
I also had in mind to keep the current logout behavior and redirect to an application page (post logout).

Your solution is better, but does not handle multi identity provider logout and I guess the final page is an IdP one, isn't it?
I guess multi profile is rather an edge use case so we can forget about it. We can logout from the first identity provider we found if ever this has been configured for that.

About the behavior, I guess it's hard to stay in the application as we'll need to call the identity provider logout.

@jkacer : what do you think of this (especially the SAML part)?

I'm going to think about it a little more, but currently, I'm in favor of merging your PR as is and fix the issues I see and reapply the good parts from my PR in a new one.

jkacer:
Hello!
It's quite a lot of new code, I did not fully review it but went over it several times. Overall, it looks fine to me.

Can I have just one question? How does the SP-initiated logout (this SP initiates it) and IdP-initiated logout (either the IdP or another SP via the IdP initiates it) differ here? Is logoutRedirect() called in both cases? I'm sorry, I'm little lost here. In one case, we should generate a logout request and get a logout response later, in the other case we should first get a logout request and generate a response later. The code probably does it because I see there is a request generator and response validator, I just cannot see how it is all linked together. Thank you!

taggiasco:
@leleuj 

I saw your solution. Not everything is clear for me. I have to have a look at it again.

For instance the final page is a static page, either on the IdP side, or on the SP side. Ideally I think there should be a callback again and we should check the logout response to see if it's a success or not, and then display a different page maybe (what I directly do). I tried to reuse the current callback controller, but it tries to get the credentials which it can't find since it's a logout response. Thus, I didn't know if it was better to create a different callback logic dedicated to the logout process or change the current one (that's how I came on the mailing list and saw your message ;-)).

I don't think it's a problem to get out of the application and go the IdP. However if we come back again to the application, maybe we can then deal with other profiles to logout. Again and again until all profiles have been log out. 


@jkacer 

No, here just the SP-initiated logout is managed. For the IdP-initiated logout I'm not really sure on how we can do it. The URL should be called by the IdP itself. Thus it won't have access to the user session (no cookies, etc), or ? And to argue about what I said before, it's maybe better to have a different callback logic that can deal with a logout response (SP-initiated logout) and a logout request (IdP-initiated logout). Otherwise the current callback logic may look horrible. 

I created the logout response validator in order to be symmetric with the login process. However it's not used yet if I remember well. The link that is missing is the callback logic.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
```diff
@@ -88,6 +127,16 @@ public final RedirectAction getRedirectAction(final WebContext context) throws H
     protected abstract RedirectAction retrieveRedirectAction(final WebContext context) throws HttpAction;
 
     /**
+     * Retrieve the redirect action for the logout.
+     * 
+     * @param context the web context
+     * @return the redirection action
+     */
+    protected RedirectAction retrieveLogoutRedirectAction(final WebContext context) {
```
taggiasco: Well that should be refactored then. But not all indirect clients should perform a logout redirect. How would you do this ?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
taggiasco: Done.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
taggiasco: Ok I see what you mean

leleuj:
OK. I see your point with the callback controller. Indeed, we need some receiver for the IdP-initiated logout.
And we want to reuse the callback controller as we don't want to ask people to define another filter.

For CAS, we use the callback filter and thanks to the `logoutRequest` parameter, we know if it's a callback for the login process or for the logout process. This is done by the appropriate credentials extractor: https://github.com/pac4j/pac4j/blob/master/pac4j-cas/src/main/java/org/pac4j/cas/credentials/extractor/TicketAndLogoutRequestExtractor.java#L56

So the left question with SAML is: can you distinguish the logout request from the login one if both are sent to the callback endpoint?

"However if we come back again to the application, maybe we can then deal with other profiles to logout.": that's an excellent idea, although it may become really complicated (that is, just removing the current profile). But I'll think about it anyway.

My plan remains unchanged:
- Next week, merge this PR in unless there is still some work in progress
- Then, refactor things a bit
- Upgrade j2e-pac4j with pac4j v2.0.0-SNAPSHOT and let Jaroslav finish the SAML support

jkacer:
Thank you for the explanation, @taggiasco !

leleuj:
I merged your PR and re-started changes on a new PR. I'll ping you when ready.

leleuj:
@taggiasco @jkacer I closed my PR and opened a new one, please take a look: https://github.com/pac4j/pac4j/pull/764

PR #761, opened 2016-12-01
Add support for collections and iterables to authZ generators

### LINKED ISSUE #758: FromAttributesAuthorizationGenerator should support other collections
There exists:

```java
    public FromAttributesAuthorizationGenerator(String[] roleAttributes, String[] permissionAttributes) {
        this.roleAttributes = this.copyArray(roleAttributes);
        this.permissionAttributes = this.copyArray(permissionAttributes);
    }
```

Also add support for lists and perhaps sets, as convenience ctors.

Other generators likely need to be massaged as well.

mmoayyed:
Closes #758

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/authorization/generator/FromAttributesAuthorizationGenerator.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/authorization/generator/FromAttributesAuthorizationGenerator.java
@@ -1,37 +1,42 @@
 package org.pac4j.core.authorization.generator;
 
+import org.pac4j.core.profile.CommonProfile;
+
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.StringTokenizer;
-
-import org.pac4j.core.profile.CommonProfile;
 
 /**
  * <p>Generate the authorization information by inspecting attributes.</p>
  * <p>The attributes containing the roles separated by the {@link #splitChar} property (can be set through {@link #setSplitChar(String)}) are
  * defined in the constructor. It's the same for the attributes containing the permissions.</p>
- * 
+ *
  * @author Jerome Leleu
  * @since 1.5.0
  */
 public class FromAttributesAuthorizationGenerator<U extends CommonProfile> implements AuthorizationGenerator<U> {
-    
-    private final String[] roleAttributes;
-    
-    private final String[] permissionAttributes;
-    
+
+    private final Iterable<String> roleAttributes;
+
+    private final Iterable<String> permissionAttributes;
+
     private String splitChar = ",";
-    
-    public FromAttributesAuthorizationGenerator(final String[] roleAttributes, final String[] permissionAttributes) {
-        this.roleAttributes = copyArray(roleAttributes);
-        this.permissionAttributes = copyArray(permissionAttributes);
+
+    public FromAttributesAuthorizationGenerator(final Iterable<String> roleAttributes, final Iterable<String> permissionAttributes) {
+        this.roleAttributes = roleAttributes;
+        this.permissionAttributes = permissionAttributes;
     }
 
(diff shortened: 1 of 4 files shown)


leleuj:
+1

pac4j-core/src/main/java/org/pac4j/core/authorization/generator/FromAttributesAuthorizationGenerator.java
```diff
@@ -1,72 +1,80 @@
...
  * @since 1.5.0
  */
 public class FromAttributesAuthorizationGenerator<U extends CommonProfile> implements AuthorizationGenerator<U> {
-    
-    private final String[] roleAttributes;
-    
-    private final String[] permissionAttributes;
-    
+
+    private final Iterable<String> roleAttributes;
```
victornoel: Maybe we could have `Collection` everywhere, as in the other class? I'm not so clear about the difference between using `Iterable` or `Collection` in this context... I feel like `Collection` is the main way to represent a collection of something, `Iterable` seems too low level in a way. Not sure :)

pac4j-core/src/main/java/org/pac4j/core/authorization/generator/FromAttributesAuthorizationGenerator.java
```diff
@@ -1,72 +1,80 @@
...
+        this.roleAttributes = roleAttributes;
+        this.permissionAttributes = permissionAttributes;
     }
 
-    private String[] copyArray(final String[] original) {
-        if (original != null) {
-            return Arrays.copyOf(original, original.length);
+    public FromAttributesAuthorizationGenerator(final String[] roleAttributes, final String[] permissionAttributes) {
+        if (roleAttributes != null) {
+            this.roleAttributes = Arrays.asList(roleAttributes);
```
victornoel: Careful, the behaviour is a little changed: `Arrays.copyOf` was making a copy, while `Arrays.asList` use the array itself as the backend of an ArrayList; changes to the array will be reflected in the `ArrayList`.

mmoayyed: What sort of changes to the array do you have in mind? Could you demonstrate that with an example perhaps?

victornoel: I'm thinking of [this kind](https://www.securecoding.cert.org/confluence/display/java/OBJ06-J.+Defensively+copy+mutable+inputs+and+mutable+internal+components) or [that kind](http://cwe.mitre.org/data/definitions/374) of best practice.

Basically in object oriented programming, the object encapsulates data and if this data is directly modifiable by an external object, then you have a problem. So when you can, you should copy mutable data in prevention, in particular when you don't have responsibility of the code that will call the concerned method. It's a very basic best practice and it was already enforced in the previous code you modify here, so I thought it would be best to continue with that.

mmoayyed: I can't think of a practical scenario in this case where that would make sense. If you insist, I can make a change, sure, but I don't see how this applies really.

victornoel: do as you wish, sorry for the bother :)

mmoayyed: Oh no worries at all. Thanks. I understand the principal and I don't mind it at all; mostly wanted to see if there was anything concrete that I should pay closer attention to. Updates coming up.

PR #763, opened 2016-12-06
Saml2ClientConfiguration supports wantAssertionsSigned but it seems ignored

andre-doherty:
Saml2ClientConfiguration supports wantAssertionsSigned however this looks like it's ignored (no effect).
The fix integrates the use of the wantAssertionsSigned property into the configuration (with direct effect on the sp metadata)

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2ServiceProviderMetadataResolver.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2ServiceProviderMetadataResolver.java
@@ -44,13 +44,14 @@ public class SAML2ServiceProviderMetadataResolver implements SAML2MetadataResolv
     private final String callbackUrl;
     private final boolean forceSpMetadataGeneration;
     private boolean authnRequestSigned;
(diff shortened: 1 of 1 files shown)


leleuj:
The `wantAssertionsSigned` property is used for the `SAML2ResponseValidator`, but indeed, I think it should be used for the generated metadata.

@mmoayyed : are you in line with this change?

@andre-doherty : can you submit the same change for the master? Thanks

andre-doherty:
Well that helped fixing a very concrete issue. 
The property was set in the shiro.ini (saml2Config.wantsAssertionsSigned = false) and the idp was sending a samlResponse without the assertions signed. On the SP side consuming the token was raising an exception (no assertion found).
That's when i started investigating into the code and seeing that that property had no effect on the consumer and on the sp_metadata as well.
As a matter of fact there was/is also a gitch on the idp (CA IDP) which didn't send the assertions signed as it didn't take the metadata properly into account (the flag wantsAssertionsSigned was indeed set to true in the first place).
So now it works fine. 

@leleuj ok i will submit that on master.

PR #764, opened 2016-12-07
Handle central logout

### LINKED ISSUE #54: Add logout redirect mecanism

leleuj:
Closes #54 
Replaces #760


pac4j-core/src/main/java/org/pac4j/core/client/Client.java
```diff
@@ -67,4 +60,14 @@
...
+
+    /**
+     * <p>Return the logout action (indirect clients).</p>
+     *
+     * @param context the current web context
+     * @param currentProfile the currentProfile
+     * @param targetUrl the target url after logout
+     * @return the redirection
+     */
+    RedirectAction getLogoutAction(WebContext context, U currentProfile, String targetUrl);
```
leleuj: I finally changed the name and the signature of this method to get a `RedirectAction` and delayed its processing if ever we want to implement custom behaviors (with multiple logouts for example).
We need the current profile as well (for OpenID Connect for example) or the computed redirection URL (useful for CAS).

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultLogoutLogic.java
```diff
@@ -0,0 +1,152 @@
...
+        logger.debug("redirectUrl: {}", redirectUrl);
+        HttpAction action;
+        if (redirectUrl != null) {
+            action = HttpAction.redirect("redirect", context, redirectUrl);
+        } else {
+            action = HttpAction.ok("ok", context);
+        }
+
+        // local logout if requested or multiple profiles
+        if (localLogout || profiles.size() > 1) {
```
leleuj: I changed the logic here as well.

The `localLogout` and the  `centralLogout` may be disabled/enabled.

taggiasco:
Hello,
I don't really see how the DefaultLogoutLogic will be used. By default, a local logout is always done (default to true), and a central logout is never done (default to false).
For SAML2, how would you define that a central logout is needed ?
Thanks

leleuj:
I updated the j2e-pac4j library: https://github.com/pac4j/j2e-pac4j

And the demo: https://github.com/pac4j/j2e-pac4j-demo and especially: https://github.com/pac4j/j2e-pac4j-demo/blob/master/src/main/webapp/WEB-INF/web.xml#L371

I will reply to your question on the mailing list.

leleuj:
@jkacer this PR has been merged. The logout support has been upgraded so you should be able to finish the dedicated support for SAML. I recommend you to use the j2e-pac4j and j2e-pac4j-demo projects for your tests as they have been upgraded to pac4j v2.

jkacer:
Hi @leleuj and thanks for the hints! We will probably get to it in January 2017. Is it OK for you or is it too late? Do you have any time constraints for version 2.0? Thanks!

leleuj:
No, it's not too late. The release date is May 2017 and the first RC in April. So you have some time left to work on that.

PR #770, opened 2016-12-08
Generic oauth20 client

julioa:
Hi,

Finally I had time to work with the new clients and the only thing I missed was a generic oauth client that can be used out-of-the-box so I've created the one included in this pr.

Additionally I've added a new attribute definition map in the profileDefinition that allows the separation between the attribute name and the path in the json response that may replace the primary/secondary attributes.


leleuj:
The build is broken due to Javadoc errors.

leleuj:
The `OAuth20Client` is already intended to be the generic OAuth 2.0 client: a new `GenericOAuth20Client` is not necessary, though some components may be missing or incomplete.

pac4j-oauth/src/main/java/org/pac4j/oauth/client/GenericOAuth20Client.java
```diff
@@ -0,0 +1,149 @@
...
+
+    public void setProfileAttrs(Map<String, String> profileAttrsMap) {
+        this.profileAttrs = profileAttrsMap;
+    }
+
+    public void setCustomParams(Map<String, String> customParamsMap) {
+        this.customParams = customParamsMap;
+    }
+
+    private static class GenericOAuth20Api extends DefaultApi20 {
```
leleuj: No, this class already exists: https://github.com/pac4j/pac4j/blob/master/pac4j-oauth/src/main/java/org/pac4j/scribe/builder/api/GenericApi20.java

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20Profile.java
```diff
@@ -0,0 +1,13 @@
+package org.pac4j.oauth.profile.generic;
+
+import org.pac4j.oauth.profile.OAuth20Profile;
+
+/**
+ * This class is the generic OAuth2 profile definition.
+ * 
+ * @author julioa
+ */
+public class GenericOAuth20Profile extends OAuth20Profile {
```
leleuj: I'm in favor of using the `OAuth20Profile` directly (by unmaking it `abstract`).

pac4j-oauth/src/main/java/org/pac4j/oauth/client/GenericOAuth20Client.java
```diff
@@ -0,0 +1,149 @@
...
+import org.pac4j.oauth.profile.generic.GenericOAuth20ProfileDefinition;
+
+/**
+ * <p>This class is a generic OAuth2 client to authenticate users in a standard OAuth2 server.</p>
+ * <p>All configuration parameters can be specified setting the corresponding attribute.</p>
+ * <p>It returns a {@link org.pac4j.oauth.profile.generic.GenericOAuth20Profile}.</p>
+ *
+ * @author Julio Arrebola
+ */
+public class GenericOAuth20Client extends OAuth20Client<GenericOAuth20Profile> {
```
leleuj: From my point of view, this is not needed. You should be able to do the same with the `OAuth20Client` and the appropriate configuration.

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20ProfileDefinition.java
```diff
@@ -0,0 +1,117 @@
...
+    String profileUrl = null;
+    String profileMethod = null;
+    String firstNodePath = null;
+    
+    public void setProfileMethod(String value) {
+        this.profileMethod = value;
+    }
+
+    @Override
+    public Verb getProfileVerb() {
```
leleuj: This `profileMethod` should be instead a `profileVerb` of type  `Verb`.

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20ProfileDefinition.java
```diff
@@ -0,0 +1,117 @@
...
+        final GenericOAuth20Profile profile = new GenericOAuth20Profile();
+        final JsonNode json = JsonHelper.getFirstNode(body, getFirstNodePath());
+        if (json != null) {
+            for (final String attribute : getPrimaryAttributes()) {
+                convertAndAdd(profile, attribute, JsonHelper.getElement(json, attribute));
+            }
+            for (final String attribute : getSecondaryAttributes()) {
+                convertAndAdd(profile, attribute, JsonHelper.getElement(json, attribute));
+            }
+            for (final Map.Entry<String, String> entry : getProfileAttributes().entrySet()) {
```
leleuj: I don't understand what this `profileAttributes` are intended for.

pac4j-oauth/src/main/java/org/pac4j/oauth/client/GenericOAuth20Client.java
julioa: Yes, since I saw other clients were using the api classes from scribe instead of creating one in this library I thought you were trying to remove the package pac4j.scribe.

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20Profile.java
julioa: That's perfect. I created this class just because that was abstract.

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20ProfileDefinition.java
julioa: I found two issues with primary/secondary attributes:
-In this new implementation both kind of attributes collect its values in the same attributes map, so I wasn't sure both lists would be used in the future.
-Most important, since both are lists, the name of the attribute and the jsosn attribute containing its values have to be the same.

That's why I added a map of profile attributes where its key is the name of the attribute in the profile and its value is the reference to the json element containing its value.

pac4j-oauth/src/main/java/org/pac4j/oauth/client/GenericOAuth20Client.java
julioa: Yes, is the OAuth20Client+profile+definition may be used as a generic client, all changes in this pr should be moved to this class.

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20ProfileDefinition.java
julioa: I guess you mean the name of the function should be "getProfileMethod" instead of "getProfileVerb". I think so too. And I just realized I forgot to make this method configurable.

julioa:
Hi,

As a summary from your comments, I think the best I can do is remove the "Generic*" classes I've created and move the valid changes to the basic OAuth20Client(+profile+definition) making them concrete.

Please, add your comments to my responses to be sure what changes do you want to have.

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20ProfileDefinition.java
leleuj: I meant that you should have a `profileVerb` property with getter and setter.

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20ProfileDefinition.java
leleuj: OK. I see your point. It makes sense.

leleuj:
I do think the `OAuth20Client` should not be changed, but the `OAuth20Profile` should become concrete and we should have a `GenericOAuth20ProfileDefintion`.

leleuj:
Merged. I slightly refactored your changes: https://github.com/pac4j/pac4j/commit/c8f7024fe5684a814b94eb5f070dd4f98e42a142

PR #774, opened 2016-12-14
Fixed wrong implementation of GenericOauth2Client

julioa:
I've found that previous implementation of GenericOAuth2Client I sent was adding additional attributes to the class instead of using the configuration object.

I've also made customParameters more flexible for different client implementations.

### DIFF
--- a/pac4j-oauth/src/main/java/org/pac4j/oauth/config/OAuth20Configuration.java
+++ b/pac4j-oauth/src/main/java/org/pac4j/oauth/config/OAuth20Configuration.java
@@ -22,7 +22,7 @@ public class OAuth20Configuration extends OAuthConfiguration<OAuth20Client, OAut
     private static final String STATE_SESSION_PARAMETER = "#oauth20StateParameter";
 
     /* Map containing user defined parameters */
-    private Map<String, String> customParams = new HashMap<>();
+    private Map<String, Object> customParams = new HashMap<>();
 
     private boolean withState;
 
@@ -32,14 +32,22 @@ public class OAuth20Configuration extends OAuthConfiguration<OAuth20Client, OAut
         return getClient().getName() + STATE_SESSION_PARAMETER;
     }
 
-    public Map<String, String> getCustomParams() {
+    public Map<String, Object> getCustomParams() {
         return customParams;
     }
 
-    public void setCustomParams(final Map<String, String> customParams) {
+    public Object getCustomParam(String name) {
+        return (customParams!=null)?customParams.get(name):null;
+    }
+
+    public void setCustomParams(final Map<String, Object> customParams) {
         this.customParams = customParams;
     }
 
+    public void addCustomParam(String name, Object value) {
+        this.customParams.put(name, value);
+    }    
+    
     public boolean isWithState() {
         return withState;
     }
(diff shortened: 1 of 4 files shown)


pac4j-oauth/src/main/java/org/pac4j/oauth/config/OAuth20Configuration.java
```diff
@@ -32,14 +32,22 @@ public String getStateSessionAttributeName() {
...
         return customParams;
     }
 
-    public void setCustomParams(final Map<String, String> customParams) {
+    public Object getCustomParam(String name) {
+        return (customParams!=null)?customParams.get(name):null;
+    }
+
+    public void setCustomParams(final Map<String, Object> customParams) {
         this.customParams = customParams;
```
leleuj: I would add a `CommonHelper.assertNotNull` for the `customParams`.

leleuj:
I like the idea. A few comments. Fix the build also.

leleuj:
Can you sync with master? I think I have fixed the build error.

julioa:
I still have a doubt for these changes: I'm not sure if mandatory config parameters for the clients should be declared in its constructor or taken from the configuration object. What policy are you using in others clients?

leleuj:
A specific parameter for a client would be attached to the client (via its constructor), though custom parameters should be attached to the configuration.

pac4j-oauth/src/main/java/org/pac4j/oauth/redirect/OAuth20RedirectActionBuilder.java
```diff
@@ -53,7 +53,7 @@ public RedirectAction redirect(final WebContext context) throws HttpAction {
 
                 service = this.configuration.getService();
             }
-            final String authorizationUrl = service.getAuthorizationUrl(this.configuration.getCustomParams());
+            final String authorizationUrl = service.getAuthorizationUrl();
```
leleuj: One last question: why custom parameters are not passed anymore to the `getAuthorizationUrl()` method?

julioa: The parameter in getAuthorizationUrl are used to add additional parameters when building the authorization url by the scribe api implementation, as in org.pac4j.scribe.builder.api.GenericApi20, so customParameters should not be used here.

leleuj: OK. Let's merge it.

leleuj:
Reviewing this once more, I don't think we took the right path. Custom parameters should be used for additional parameters on the authorization URL (like for OpenID Connect) and well-known parameters should be defined via setters/getters in the clients (to make the contract obvious).

I'm reverting this change. Would you mind revisiting it while keeping the setters/getters for the expected parameters?

leleuj:
@julioa reverting with: https://github.com/pac4j/pac4j/pull/781

leleuj:
It would be great to have a manual test as well (https://github.com/pac4j/pac4j/tree/master/pac4j-oauth/src/test/java/org/pac4j/oauth/run) and improve the documentation (https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/oauth.md).

PR #775, opened 2016-12-16
Remove a possibly unnecessary/unneeded spring dependency

andre-doherty:
This pull request aims at removing spring dependencies on the saml pac4j implementation.
Spring-core seems only to be in use for an Assert that can be implemented using the utils class CommonHelper
Of course test dependencies which relies deeply on spring are unchanged.

This is also a sanity change to me as shiro in the first place targets java EE application as a clear alternative to Spring Security.


leleuj:
Good catch! Can you submit the same PR for the master?

andre-doherty:
Done...

leleuj:
Many thanks.

PR #778, opened 2016-12-19
added PathMatcher

wardr:
Proposed replacement for ExludedPathMatcher.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/matching/ExcludedPathMatcher.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/matching/ExcludedPathMatcher.java
@@ -23,11 +23,6 @@ public final class ExcludedPathMatcher extends PathMatcher {
         setExcludePath(excludePath);
     }
 
-    @Override
-    public boolean matches(final WebContext context) {
-        return super.matches(context);
-    }
-
     public String getExcludePath() {
         return super.getExcludedPatterns().iterator().next().pattern();
     }
--- a/pac4j-core/src/main/java/org/pac4j/core/matching/ExcludedPathMatcher.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/matching/ExcludedPathMatcher.java
@@ -2,11 +2,8 @@ package org.pac4j.core.matching;
 
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.exception.TechnicalException;
-import org.pac4j.core.util.CommonHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.util.regex.Pattern;
 
 /**
  * To match requests by excluding path.
--- a/pac4j-core/src/main/java/org/pac4j/core/matching/ExcludedPathMatcher.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/matching/ExcludedPathMatcher.java
@@ -11,19 +11,16 @@ import java.util.regex.Pattern;
 /**
  * To match requests by excluding path.
  *
+ * @deprecated use {@link org.pac4j.core.matching.PathMatcher} instead.
  * @author Jerome Leleu
  * @since 1.8.1
  */
-public final class ExcludedPathMatcher implements Matcher {
+@Deprecated
+public final class ExcludedPathMatcher extends PathMatcher {
 
     private final static Logger logger = LoggerFactory.getLogger(ExcludedPathMatcher.class);
 
-    private String excludePath;
-
-    private Pattern pattern;
-
-    public ExcludedPathMatcher() {
-    }
+    public ExcludedPathMatcher() {}
 
     public ExcludedPathMatcher(final String excludePath) {
         setExcludePath(excludePath);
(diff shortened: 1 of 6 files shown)


leleuj:
Thanks. The build is broken due to Javadoc issues.

leleuj:
This is a core component, we need a test class to verify the behaviour. Also, the `ExcludedPathMatcher` should be marked as deprecated and certainly be nothing more than an inherited class from this one.
Finally, the documentation must be updated, in the release notes (https://github.com/pac4j/pac4j/blob/master/documentation/docs/release-notes.md) as well as in the matchers page (https://github.com/pac4j/pac4j/blob/master/documentation/docs/matchers.md).

pac4j-core/src/main/java/org/pac4j/core/matching/PathMatcher.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public void addExcludedPath(final String path) {
+        excludedPaths.add(path);
+    }
+
+    /**
+     * Convenience method for excluding all paths starting with a prefix e.g. "/foo" would exclude "/foo", "/foo/bar", etc.
+     */
+    public void addExcludedBranch(final String root) {
+        addExcludedRegex("^" + root + "(/.*)?$");
```
leleuj: Same check here.

pac4j-core/src/main/java/org/pac4j/core/matching/PathMatcher.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Any path matching this regex will be excluded.
+     */
+    public void addExcludedRegex(final String regex) {
+        excludedPatterns.add(Pattern.compile(regex));
```
leleuj: And there.

wardr:
Not included with this pull, but something to keep in mind in future as a usability improvement:

It might be useful for matchers to have a name field. So, in configuration, you could do:

Now:

```
PathMatcher matcher = new PathMatcher()
    .excludePath("/")
    .excludePath("/login")
    .excludePath(FORM_CLIENT_CALLBACK);
config.addMatcher("authPathMatcher", matcher);
```

Future:

```
PathMatcher matcher = new PathMatcher("authPathMatcher")
    .excludePath("/")
    .excludePath("/login")
    .excludePath(FORM_CLIENT_CALLBACK);
config.addMatcher(matcher);

```

victornoel:
@wardr I'm not sure about the fluent API style: or at least, I think it should also work in a Java Bean style, because for example if one wants to define a `PathMatcher` as part of his dropwizard configuration, it should work out of the box with something like this:
```yaml
  matchers:
    authPathMatcher:
      class: o.p.c.m.PathMatcher
      excludePaths:
        - /
        - /login
```

As for the name, to me it seems cleaner to have the name as part of the `Config` and not as part of the `Matcher` object, this would prevent adding multiple matchers with the same name!
Actually, I think it should be the same with the `Client`s @leleuj

wardr:
Good point. Would adding a method: `setExcludePaths(String... paths)` make it work for the Dropwizard configuration case?

victornoel:
Let me do some tests tomorrow morning and tell you, but if I take too long, just add it and then I will submit another PR if it's problematic :)

victornoel:
@wardr actually Jackson will use the getters to add elements to the `Set`s, so you just need to make them `public` (which is not a bad idea anyway).

pac4j-core/src/main/java/org/pac4j/core/matching/PathMatcher.java
```diff
@@ -0,0 +1,102 @@
...
+
+    protected Set<Pattern> getExcludedPatterns() {
+        return excludedPatterns;
+    }
+
+    private static void validatePath(String path) {
+        CommonHelper.assertNotBlank("path", path);
+        if (!path.startsWith("/")) {
+            final String msg = "Excluded path must begin with a /";
+            logger.error(msg);
```
victornoel: Don't both log and throw: simply put the error message as the exception message.

victornoel: (I realised it was like that before, but there have been a discussion on the matter lately here: https://groups.google.com/d/msg/pac4j-users/CR9_KVkkxGk/IcFb3SqbEAAJ :) )

pac4j-core/src/main/java/org/pac4j/core/matching/ExcludedPathMatcher.java
```diff
@@ -1,58 +1,38 @@
...
-            if (!excludePath.startsWith("^") || !excludePath.endsWith("$")) {
-                final String msg = "Your regular expression: '" + excludePath + "' must start with a ^ and ends with a $ to define a full path matching";
-                logger.error(msg);
-                throw new TechnicalException(msg);
-            }
-            pattern = Pattern.compile(excludePath);
+        if (!super.getExcludedPatterns().isEmpty())
+        {
+            String msg = "ExcludedPathMatcher does not support excluding multiple paths. Use PathMatcher instead.";
+            logger.error(msg);
```
victornoel: Same as in `PathMatcher`

leleuj:
+1

@victornoel Please merge in when it's ok for you.

wardr:
@victornoel all changes incorporated, thanks.

Note, I don't recommend adding the regex's directly (e.g. pathMatcher.getExcludedPatterns().add("...")), because doing so would by-pass the warnings and validation checks in the excludeRegex method.  Not really a big deal though.

leleuj:
@wadr Just to be sure people don't add regexp directly, why not return an immutable set in the getter?

victornoel:
@leleuj the problem is that I would have liked if people could exploit the javabean nature of the matcher so that they could define them declaratively, for example via the dropwizard configuration file.
For this we need either a mutable getter of the collection or a setter for the collection.

In that case @wardr simply add a setter for the whole collections `setExcludedPatterns` and `setExcludedRegexps` and iterate over them to validate them if needed and also print the message.

wardr:
Ok, should be good to go now.

pac4j-core/src/main/java/org/pac4j/core/matching/PathMatcher.java
```diff
@@ -89,6 +90,14 @@ boolean matches(final String path) {
         return excludedPatterns;
     }
 
+    public void setExcludedPaths(Collection<String> paths) {
+        paths.forEach(path -> excludePath(path));
```
victornoel: Wouldn't it be better if this method remove previous values? Since it is called set and not add... sorry :)

pac4j-core/src/main/java/org/pac4j/core/matching/PathMatcher.java
```diff
@@ -89,6 +90,14 @@ boolean matches(final String path) {
...
+    }
+
+    public void setExcludedPatterns(Collection<String> regularExpressions) {
+        regularExpressions.forEach(regex -> excludeRegex(regex));
```
victornoel: same thing :)

wardr:
Fair point; done.

PR #780, opened 2016-12-26
Keycloak

julioa:
Here is the missed test case for the keycloak oidc client

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/client/KeycloakOidcClient.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/client/KeycloakOidcClient.java
@@ -20,8 +20,6 @@ import org.pac4j.oidc.profile.keycloak.KeycloakOidcProfile;
  */
 public class KeycloakOidcClient extends OidcClient<KeycloakOidcProfile> {
 
-    private static final long serialVersionUID = -665205280702028258L;
-
     public KeycloakOidcClient() {
     }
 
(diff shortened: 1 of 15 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/client/KeycloakOidcClient.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public KeycloakOidcClient(final OidcConfiguration configuration) {
+        super(configuration);
+    }
+
+    @Override
+    protected void internalInit(final WebContext context) {
+        CommonHelper.assertNotNull("configuration", getConfiguration());
+        String baseUri = getConfiguration().getCustomParam("authUrl");
```
leleuj: In the same spirit as for the `GenericOAuth20Client`, can you use specific properties: `authUrl` and `realm` for these parameters?

Thanks.

leleuj:
It would be good to improve the documentation (https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/openid-connect.md).

pac4j-oidc/src/main/java/org/pac4j/oidc/client/KeycloakOidcClient.java
julioa: I've made a change moving those properties to the configuration class instead of the client to see how it looks. If you think this is better than use the client to set those configuration parameters I can do the same changes to the GenericOauth20Client; otherwise I'll move then to the client.

leleuj:
You raise a good point. It makes sense here to have a specific configuration for Keycloak to avoid setting things at two different places (at the configuration and client levels). For OAuth, for consistency with other clients, I'd prefer to keep setters at the client level.

Just fix the build and we'll be good for the merge.

leleuj:
Many thanks for this new contribution

Issue #783, opened 2017-01-03
SAML: Unable to find resource '/templates/saml2-post-binding.vm'


jknack:
Saml client isn't able to load velocity templates. I tracked it down think the problem is in [VelocityEngineFactory.java](https://github.com/pac4j/pac4j/blob/468f2d006bcdb7a210db1e6660fda10a8d374920/pac4j-saml/src/main/java/org/pac4j/saml/util/VelocityEngineFactory.java#L25):

```java
 final Properties props = new Properties(net.shibboleth.utilities.java.support.velocity.VelocityEngine.getDefaultProperties());

            props.setProperty(RuntimeConstants.INPUT_ENCODING, "UTF-8");
            props.setProperty(RuntimeConstants.OUTPUT_ENCODING, "UTF-8");
            props.setProperty(RuntimeConstants.RESOURCE_LOADER, "classpath");
            props.setProperty(RuntimeConstants.RUNTIME_LOG_LOGSYSTEM_CLASS, SLF4JLogChute.class.getName());
```

Not sure how `defaults` properties work (first time I see them) but they definitely don't work.
I change:

```java
 final Properties props = new Properties(net.shibboleth.utilities.java.support.velocity.VelocityEngine.getDefaultProperties());
```

with:

```java
 final Properties props = new Properties();
props.putAll(net.shibboleth.utilities.java.support.velocity.VelocityEngine.getDefaultProperties());
...
```

and templates load properly.

leleuj:
Good catch. Would you mind submitting two PRs for that (one for 1.9.x branch and the other one for master)? Thanks

jknack:
Sure, will do.
Thanks.

PR #786, opened 2017-01-04
SAML: Unable to find resource '/templates/saml2-post-binding.vm' fix …

jknack:
…#783


leleuj:
Thanks for your contribution.

Issue #716, opened 2016-10-20
Produce Fat/Shaded JARs


leleuj:
https://docs.google.com/document/d/1k-Cw8QiRHVA91nI-rww4Bh1ealU-qa0avBmUbDCofqQ/edit#heading=h.iprr5t12q3ut

PR #792, opened 2017-01-07
Produce fat/shaded JARs for modules

mmoayyed:
Closes https://github.com/pac4j/pac4j/issues/716

Produces a pac4j-moduleName-version-all.jar that contains all the needed dependencies for a given pac4j module. Also adds the maven-wrapper plugin to remove the need for one to download/install maven manually.

Note for @leleuj: must ensure the new JAR is also published at release time.


leleuj:
I'm a bit surprised by this PR. I thought it was just about adding the appropriate `wrapper-maven-plugin`.

@millross you were interested in this PR as well

leleuj:
@mmoayyed are all these Maven stuffs really necessary? I may admit that we could have them in the root dir, but it seems overkill to have them in each submodule. Don't you think?

millross:
The shade plugin should do what's necessary I think. But maybe @mmoayyed found that he needed to introduce the maven wrapper to make the whole setup work cleanly to produce the uber-jar (most of the changes seem to be around the maven wrapper introduction unless I'm missing something).

leleuj:
I'd like to close a few issues before the RC1.

@mmoayyed I'm not really convinced by the embedded Maven files, but if we want to go down that road, shouldn't we update the Travis configuration as well?

leleuj:
Proposing this one (just with the shade plugin) to start: https://github.com/pac4j/pac4j/pull/811

PR #793, opened 2017-01-09
initialization should happen before accessing ajaxRequestResolver

### LINKED ISSUE #787: IndirectClient are not initialised soon enough in getRedirectAction
Hi,

I think that in the implementation of `IndirectClient.getRedirectAction()` the call to `init(Context)` should be at the beginning.

For example, it relies on `ajaxRequestResolver` in the beginning, and at that moment, we are not sure it has been correctly initialised (since `internalInit()` does verify that it is different than `null`).

victornoel:
Fixes #787


leleuj:
+1

PR #794, opened 2017-01-09
Keep abstract methods close to their use

### LINKED ISSUE #788: Proposal: move BaseClient.retrieveCredentials to subclasses.
Hi,

While browsing the clients code, I noticed that `BaseClient` was defining `retrieveCredentials` as abstract, but wasn't making any use of it.

Then each one of the two types of clients did make some use of it.

And finally, some of their subclass would implement it.

I think it would make more sense to have `BaseClient.retrieveCredentials` defined as abstract in each of the subclasses. Even though it would seem like code duplication, it seems more important to me to keep the declaration and usage of an abstract method in the same class for maintenance and code comprehension.

It is not really critical, but I thought I would raise the discussion here :)

victornoel:
Closes #788


leleuj:
+1

PR #797, opened 2017-01-11
Bugfix for State parameter check in OidcExtractor

entropicdrifter:
I noticed that the OidcExtractor has been failing the state parameter check even when the state value strings match. 
This is because the redirect action builder saves the state as a String, but the comparison check here (on line 79) relies on the com.nimbusds.oauth2.sdk.id.State.equals() method, which checks first if the other object is an instanceof State (before checking to see if the value strings match). Since the retrieved object is a raw String, it fails the check.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/extractor/OidcExtractor.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/extractor/OidcExtractor.java
@@ -76,7 +76,7 @@ public class OidcExtractor extends InitializableWebObject implements Credentials
         if (state == null) {
             throw new TechnicalException("Missing state parameter");
         }
-        if (!state.equals(new State(context.getSessionAttribute(OidcConfiguration.STATE_SESSION_ATTRIBUTE)))) {
+        if (!state.equals(new State((String)context.getSessionAttribute(OidcConfiguration.STATE_SESSION_ATTRIBUTE)))) {
             throw new TechnicalException("State parameter is different from the one sent in authentication request. "
                     + "Session expired or possible threat of cross-site request forgery");
         }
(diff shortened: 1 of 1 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/extractor/OidcExtractor.java
```diff
@@ -76,7 +76,7 @@ public OidcCredentials extract(final WebContext context) throws HttpAction {
         if (state == null) {
             throw new TechnicalException("Missing state parameter");
         }
-        if (!state.equals(context.getSessionAttribute(OidcConfiguration.STATE_SESSION_ATTRIBUTE))) {
+        if (!state.equals(new State((String)context.getSessionAttribute(OidcConfiguration.STATE_SESSION_ATTRIBUTE)))) {
```
leleuj: What will happen if the value in the session is `null`? Does it work?

As we previously check that the `state` is not `null`, I'd prefer to have: `if (CommonHelper.areNotEquals(state.getValue(), context.getSessionAttribute(OidcConfiguration.STATE_SESSION_ATTRIBUTE)) {`

victornoel:
I was also wondering if we are 100% sure that it is always a String that is put in that attribute?

leleuj:
Well done! @victornoel is right: we have a `State` in session.

See: https://github.com/pac4j/pac4j/blob/master/pac4j-oidc/src/main/java/org/pac4j/oidc/redirect/OidcRedirectActionBuilder.java#L92

```java
        State state = new State();
        params.put(OidcConfiguration.STATE, state.getValue());
        context.setSessionAttribute(OidcConfiguration.STATE_SESSION_ATTRIBUTE, state);
```

@entropicdrifter are you sure of your issue?

entropicdrifter:
You guys are right! The issue is with the PlayCookieStore. Since it serializes the State as JSON, it comes back out of the Store as a String. I'll need to add a workaround to the PlayCookieStore. Thanks for your help, this has sped up my progress on this quite a bit.

Issue #782, opened 2017-01-03
TechnicalException:  Do not use the resource: or classpath: prefix for non-existing files


mmoayyed:
Happy new year!

```
Do not use the resource: or classpath: prefix for non-existing files. Use a direct path (relative or absolute, no prefix)
```

Why would I want to use ANY sort of path when something is non-existing?

leleuj:
Happy new year! This is exactly what the error message intends to say. The use case is for a non-existing keystore which will be automatically created. These prefixes must be clearly revisited in version 2.0.

PR #821, opened 2017-02-11
Refactor SAML and its files support

### LINKED ISSUE #721: Refactor SAML and its files support
Use Spring to handle files support, maybe remove `resource:` or / and `classpath:` prefixes support, clean `CommonHelper` from files support

https://docs.google.com/document/d/1k-Cw8QiRHVA91nI-rww4Bh1ealU-qa0avBmUbDCofqQ/edit#

leleuj:
Closes #721 and #782 

@mmoayyed this PR is for you ;-) I used the Spring `Resource` instead of our custom implementations to handle file support in the SAML support. I'm pretty happy with the result. Just let me know if everything is ok for you.
Notice that the SAML properties in CAS will be changed: `saml.identityProviderMetadataFilepath`, `saml.identityProviderMetadataClasspath`, `saml.identityProviderMetadataUrl`... to accomodate that.


leleuj:
Related PR: https://github.com/pac4j/j2e-pac4j-demo/pull/26, https://github.com/pac4j/spring-webmvc-pac4j-demo/pull/2 and https://github.com/pac4j/spring-webmvc-pac4j-boot-demo/pull/5

mmoayyed:
Excellent. Initial comments:

While this is OK, I dont like the separation of properties at all because it adds learning curve. A resource is either defined as filesystem resource, a URL resource or a classpath resource [or in certain cases a servlet resource]. These are all defined into the URI directly with file, classpath, etc. [Which one may argue is still confusing, but those are already taught to developers, and those are the patterns which are consistent in Spring applications and previous versions of pac4j] 

So rather than creating individual methods to load each type [and rather than parsing things manually], we should begin to use `ResourceLoader` instead since you're already using `Resource` anyway. That loader does exactly what you're trying to do manually.

Which means:

I just need a single property called: `samlIdentityProviderMetadataLocation` or something which I can then use classpath or https or c:\whatever in the value. The loader will load and parse the relevant resource type for you, which then reduces your dependency on Spring with specific types. Pac4j does and should not have to care what sort of resource is injected as much as possible. All it cares about should be the resource.

What I dont remember exactly at this moment is, whether `ResourceLoader` can be used outside a spring application context. It very likely can.

leleuj:
OK. I see your point. In that case, I think it's better to be backward compatible with our previous mechanism: keep the `setIdentityProviderMetadataPath` setter for example and not use `ResourceLoader` and:

- `resource:` or `classpath:` prefixes -> `ClassPathResource`
- `http:` or `https:` prefixes -> `UrlResource`
- `file:` prefix or no prefix -> `FileSystemResource`

mmoayyed:
Yes. That's good.

Also I want  to suggest that maybe we open up a `setIdentityProviderMetadataPath (Resource)` method too. That should make it very convenient for those who want to use the resource loader facility.

leleuj:
This is now fully backward compatible and you can define a path (like in v1.9) or directly a `Resource`.

pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
```diff
@@ -431,10 +417,9 @@ public boolean isAuthnRequestSigned() {
 	 *            Web context to transport additional information to the configuration.
 	 */
     protected void init(final String clientName, final WebContext context) {
```
leleuj: @jkacer: I wanted to draw your attention to this change (since 2.0.0-RC1). The `SAML2ClientConfiguration` component is now an `InitializableObject` with its internal `init()` method, which I choose to call from your additional `init(String, WebContext)` one. It may imply a change in your extension.

jkacer: Thank you for letting me know in advance! I think I will get to the update of the extension after PAC 2.0 is released, not before. Thanks!

leofromgroza:
After changes that was done in scope of current issue, the possibility of using config server as resource of storing Service Provider Metadata was broken. I'm receiving such exception:

>  org.pac4j.core.exception.TechnicalException: http://config-server/.../saml_sp_metadata.xml must be a writable resource
        at org.pac4j.saml.client.SAML2ClientConfiguration.setServiceProviderMetadataPath(SAML2ClientConfiguration.java:278) ~[pac4j-saml-2.1.0.jar!/:na]

As I understand it is not possible to make serviceProviderMetadataResource not WritableResource and remove instanceof check from [here](https://github.com/pac4j/pac4j/blob/pac4j-2.1.0/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java#L277). But we not always need to write into SP metadata, only if forceSpMetadataGeneration == true. 
@leleuj What do you think about that?

leleuj:
The SAML SP metadata is generated from the configuration. I'm not sure why you would already have one...

leofromgroza:
We have scalable application and want every instance to use the same SP metadata. So we've generated it once and put to configuration server. But we cannot.

leleuj:
If you have already generated your metadata, then you should set `forceSpMetadataGeneration` to `true`.

leofromgroza:
If I do so, Pac4J will generate new SP metadata and start to use it. That will broke connection with IdP server, because we should provide new metadata for them. 
You know that in some cases it is much more comfortable to have pre-configured metadata ([Spring also knows about this](https://docs.spring.io/autorepo/docs/spring-security-saml/1.0.2.RELEASE/reference/html/configuration-metadata.html#configuration-metadata-sp-import)), so you left such option (forceSpMetadataGeneration == false). 
But the only thing that we currently cannot do: use SP that is stored on Config Server.

leleuj:
I meant "set forceSpMetadataGeneration to **false**". And indeed, it's one-way system currently: from the configuration in the Java source code, we generate the metadata. But we don't read metadata to generate a Java configuration. Any contribution will be appreciated.

PR #825, opened 2017-02-14
CallbackUrlResolver -> UrlResolver

### LINKED ISSUE #820: CasClient redirect URL ignores callback url resolver? (1.9)
At [CasClient:87](https://github.com/pac4j/pac4j/blob/1.9.x/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java#L87), the url resolver is used to construct the login url, but at [CasClient:89](https://github.com/pac4j/pac4j/blob/1.9.x/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java#L89) the redirect url is built just by calling computeFinalCallbackUrl on the IndirectClient ancestor class. The IndirectClient class doesn't use the CasConfig, and therefore has no way to use the url resolver when building the redirect url - it just uses DefaultUrlCallbackResolver.

This feels like a bug, but perhaps there's something I'm missing. Is there a reason that the callback url resolver should be used for building the login url and not the redirect url? It seems like it should be the other way round, given that the login url is not a callback. While I didn't read the 2.0 code very carefully, that seems to be the way it works there (resolver used for redirect, not for login).

leleuj:
Closes #820


pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
```diff
@@ -154,12 +154,32 @@ public void setCallbackUrl(final String callbackUrl) {
...
     public String getCallbackUrl() { return this.callbackUrl; }
 
-    public CallbackUrlResolver getCallbackUrlResolver() {
-        return callbackUrlResolver;
+    /**
+     * Use {@link #getUrlResolver()} instead.
+     *
+     * @return the URL resolver for the callback URL
+     */
+    @Deprecated
```
victornoel: @leleuj any reason to keep these deprecated methods in version 2.0.0?

leleuj: Based on what I see on the mailing list, the `CallbackUrlResolver` concept is used by many people. These deprecated methods are meant to ease the transition to the new `UrlResolver` concept...

PR #838, opened 2017-02-22
Use HttpConstants.HTTP_METHOD for the HTTP method

### LINKED ISSUE #837: Use HttpConstants.HTTP_METHOD for the HTTP method

leleuj:
Closes #837

### DIFF
--- a/pac4j-cas/src/test/java/org/pac4j/cas/client/CasClientTests.java
+++ b/pac4j-cas/src/test/java/org/pac4j/cas/client/CasClientTests.java
@@ -15,6 +15,8 @@ import java.util.zip.Deflater;
 
 import static org.junit.Assert.*;
 
+import static org.pac4j.core.context.HttpConstants.*;
+
 /**
  * This class tests the {@link CasClient} class.
  * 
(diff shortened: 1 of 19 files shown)


victornoel:
@leleuj why impose a fixed number of accepted methods?
You could imagine that people would want to accept any method, even non-standard ones... for example there is PATCH sometimes used in REST API!

It makes sense to have an enum like `HTTP_METHOD` for referencing existing methods if you want, but not to force the `WebContext` to return one of them...

leleuj:
I have never seen the case of people using non-standard HTTP methods. It seems very unlikely, doesn't it?
But you're right: I forgot the PATCH method.

victornoel:
Well, in my mind it's better to leave some door opens if it doesn't cost anything than to close some for no advantage: by accepting a `String` you force yourself to implement a solution to this situation, while by accepting an enum, you let integration implementors (such as me :) with the responsibility of doing something with a non-supported http method... and for sure there is nothing to do in that situation, things will just crash with a `RuntimeException` while it could have been easily handled by pac4j I think...

leleuj:
I see your point, but I'm not really convinced. You can still access the raw request and check the custom HTTP method.
@mmoayyed @millross what do you think?

victornoel:
I'm not worried about integration code (that has access to the raw request as you say) trying to access this method, I'm worried about all the pac4j and user code that doesn't have access to the raw request and that will fail when `WebContext.getMethod()` is called because it returns null or throws an exception when the method is not one of the expected enum values.

And I bet most of the code calling `WebContext.getMethod()` is likely to simply pass it to another code expecting a String because everybody considers that an HTTP method can be an arbitrary String.
Look at all specs like servlet or jaxrs, at HTTP servers and clients like Apache or Jetty HttpClient: none has ever introduced an enum to close the space of possible methods because it is not an error to use a non-standard method! At worst callers of `WebContext.getMethod()` will compare the result with some expected method to trigger some method-specific code.

leleuj:
OK. I'll follow your recommendation (and revert) unless someone else has a strong opinion on this...

victornoel:
Thanks, I really feel like it's better like this :)

mmoayyed:
I am OK with this PR as is. I'd also be OK with this was reversed :)

leleuj:
Reverting...

leleuj:
Using the `HTTP_METHOD` as much as possible: https://github.com/pac4j/pac4j/pull/850

PR #847, opened 2017-03-06
Remove final on client method in order to proxify it

mouminoux:
To manage multi-tenant authentication I removed final on client method to be able to proxify client classes with CGLib.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
@@ -46,7 +46,7 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
     protected abstract C retrieveCredentials(final WebContext context) throws HttpAction;
 
     @Override
-    public final U getUserProfile(final C credentials, final WebContext context) throws HttpAction {
+    public U getUserProfile(final C credentials, final WebContext context) throws HttpAction {
         init(context);
         logger.debug("credentials : {}", credentials);
         if (credentials == null) {
--- a/pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
@@ -40,7 +40,7 @@ public abstract class IndirectClient<C extends Credentials, U extends CommonProf
     }
 
     @Override
-    public final HttpAction redirect(final WebContext context) throws HttpAction {
+    public HttpAction redirect(final WebContext context) throws HttpAction {
         final RedirectAction action = getRedirectAction(context);
         if (action.getType() == RedirectType.REDIRECT) {
             return HttpAction.redirect("redirection via 302", context, action.getLocation());
@@ -59,7 +59,7 @@ public abstract class IndirectClient<C extends Credentials, U extends CommonProf
      * @return the redirection action
      * @throws HttpAction requires an additional HTTP action
      */
-    public final RedirectAction getRedirectAction(final WebContext context) throws HttpAction {
+    public RedirectAction getRedirectAction(final WebContext context) throws HttpAction {
         init(context);
 
         // it's an AJAX request -> unauthorized (instead of a redirection)
@@ -101,7 +101,7 @@ public abstract class IndirectClient<C extends Credentials, U extends CommonProf
      * @throws HttpAction whether an additional HTTP action is required
      */
     @Override
-    public final C getCredentials(final WebContext context) throws HttpAction {
+    public C getCredentials(final WebContext context) throws HttpAction {
         init(context);
         final C credentials = retrieveCredentials(context);
         // no credentials -> save this authentication has already been tried and failed


pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
```diff
@@ -46,7 +46,7 @@
     protected abstract C retrieveCredentials(final WebContext context) throws HttpAction;
 
     @Override
-    public final U getUserProfile(final C credentials, final WebContext context) throws HttpAction {
+    public U getUserProfile(final C credentials, final WebContext context) throws HttpAction {
```
leleuj: Overriding the `retrieveUserProfile` method was not sufficient in your case?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
```diff
@@ -40,7 +40,7 @@ protected void internalInit(final WebContext context) {
     }
 
     @Override
-    public final HttpAction redirect(final WebContext context) throws HttpAction {
+    public HttpAction redirect(final WebContext context) throws HttpAction {
```
leleuj: Overriding the `getRedirectAction` method would not be sufficient in your case?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
```diff
@@ -101,7 +101,7 @@ public final RedirectAction getRedirectAction(final WebContext context) throws H
      * @throws HttpAction whether an additional HTTP action is required
      */
     @Override
-    public final C getCredentials(final WebContext context) throws HttpAction {
+    public C getCredentials(final WebContext context) throws HttpAction {
```
leleuj: Overriding the `retrieveCredentials` method was not sufficient in your case?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
mouminoux: In fact I need different callback URL by tenant (tenant1.mydomain.com, tenant2.mydomain.com)
This URL is constructed in the OAuth20Service and checked when the user come back from google login page.
If I proxify Google2Client by tenant I am sure I have one OAuth20Service by tenant too with the right callback URL.

leleuj: Did you try using a more generic `CallbackUrlResolver` than the default one?

mouminoux: The CallbackUrlResolver is called one time during the construction and the value computed is stored in OAuthConfig which is a properties of the OAuthService, I don't know how I could make it work with a custom CallBackUrlResolver.

leleuj: OK. You're right. It should be much easier in v2.0 by the way. Isn't it enough to override the  `getRedirectAction` method in that case?

mouminoux: I think I could but I need to reconstruct the location URL.
It's really easier to me to proxify two pac4j bean with my scope tenant and I construct the Clients like this: 
`    @Bean
    @TenantScope
    public Clients clients(Google2Client google2Client) {
        String callbackUrl = ServletUriComponentsBuilder.fromCurrentContextPath().path("/callback?client_name=Google2Client").build().toString();
        return new Clients(callbackUrl, google2Client);
    }`

leleuj: I'm not sure to follow you. Reading the source code again, I even think you could just override the `retrieveRedirectAction` method, the `redirect` and `getRedirectAction`  methods just wrap it with code you should not get rid of.

mouminoux: The callback URL is checked by the `GoogleOAuthServiceImpl` which is a singleton I can't make it work by overriding the `Google2Client`.
In my case I have two callback URL: 
http://tenant1.com/callback?client_name=Google2Client
http://tenant2.com/callback?client_name=Google2Client

The authentication works with the first tenant but with the second I have this error:

> Caused by: com.github.scribejava.core.exceptions.OAuthException: Response body is incorrect. Can't extract a '"access_token"\s*:\s*"(\S*?)"' from this: '{
>  "error": "redirect_uri_mismatch",
>  "error_description": "Bad Request"
> }

leleuj: In your case, I think you can override the `getAccessToken` method and create the appropriate Scribe service based on the tenant callback URL to retrieve the token: https://github.com/pac4j/pac4j/blob/1.9.x/pac4j-oauth/src/main/java/org/pac4j/oauth/client/BaseOAuth20Client.java#L55

mouminoux: Ok,

I found a solution, I overrided:

OAuthConfig
```
@Override
public String getCallback() {
        return ServletUriComponentsBuilder.fromCurrentContextPath().path("/callback?client_name=Google2Client").build().toString();
}
```
Google2Client
```
@Override
public String getName() {
    return Google2Client.class.getSimpleName();
}

@Override
public String getCallbackUrl() {
    return ServletUriComponentsBuilder.fromCurrentContextPath().path("/callback?client_name=Google2Client").build().toString();
}

@Override
public String computeFinalCallbackUrl(WebContext context) {
    return callbackUrlResolver.compute(getCallbackUrl(), context);
}

@Override
protected OAuthConfig buildOAuthConfig(WebContext context) {
    OAuthConfig oAuthConfig = super.buildOAuthConfig(context);
    String state = oAuthConfig.getState();
    return new MultitenantOAuthConfig(this.getKey(), this.getSecret(), computeFinalCallbackUrl(context),
            SignatureType.Header, getOAuthScope(), null, this.getConnectTimeout(), this.getReadTimeout(), hasOAuthGrantType() ? "authorization_code" : null, state, this.getResponseType());
}
```

leleuj: Good. Is this PR still necessary? Please close it if you don't need this change anymore.

mouminoux: No it's ok, Thx

PR #849, opened 2017-03-14
Revert "Use HttpConstants.HTTP_METHOD for the HTTP method"

leleuj:
Reverts pac4j/pac4j#838

### DIFF
--- a/pac4j-cas/src/test/java/org/pac4j/cas/client/CasClientTests.java
+++ b/pac4j-cas/src/test/java/org/pac4j/cas/client/CasClientTests.java
@@ -15,8 +15,6 @@ import java.util.zip.Deflater;
 
 import static org.junit.Assert.*;
 
-import static org.pac4j.core.context.HttpConstants.*;
-
 /**
  * This class tests the {@link CasClient} class.
  * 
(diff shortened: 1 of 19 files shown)


pac4j-cas/src/test/java/org/pac4j/cas/client/CasClientTests.java
```diff
@@ -127,23 +125,23 @@ public void testBackLogout() {
         casClient.setCallbackUrl(CALLBACK_URL);
         casClient.init(null);
         final MockWebContext context = MockWebContext.create().addRequestParameter(CasConfiguration.LOGOUT_REQUEST_PARAMETER, LOGOUT_MESSAGE)
-            .setRequestMethod(HTTP_METHOD.POST);
+            .setRequestMethod("POST");
```
victornoel: I think you could keep the enum (or maybe some constants) for all the cases where in pac4j code you need to refer to an HTTP method instead of using a plain string.

leleuj: Yes, I could use the enum here and there to avoid using simple strings. I will take a look at that while adding the `PATCH` method.

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/ParameterExtractor.java
```diff
@@ -37,10 +36,10 @@ public ParameterExtractor(final String parameterName, final boolean supportGetRe
 
     @Override
     public TokenCredentials extract(WebContext context) throws HttpAction, CredentialsException {
-        final HttpConstants.HTTP_METHOD method = context.getRequestMethod();
-        if (HttpConstants.HTTP_METHOD.GET == method && !supportGetRequest) {
+        final String method = context.getRequestMethod();
+        if ("GET".equals(method) && !supportGetRequest) {
```
victornoel: Here also :)

pac4j-cas/src/test/java/org/pac4j/cas/client/CasClientTests.java
victornoel: Ok! ignore my other comment then :)

leleuj: Here we go: https://github.com/pac4j/pac4j/pull/850

PR #850, opened 2017-03-14
use HttpConstants.HTTP_METHOD wherever possible

leleuj:


pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/ParameterExtractor.java
```diff
@@ -37,9 +39,9 @@ public ParameterExtractor(final String parameterName, final boolean supportGetRe
     @Override
     public TokenCredentials extract(WebContext context) throws HttpAction, CredentialsException {
         final String method = context.getRequestMethod();
-        if ("GET".equals(method) && !supportGetRequest) {
+        if (HTTP_METHOD.GET.name().equals(method) && !supportGetRequest) {
```
victornoel: Shouldn't we compare by ignoring case?

leleuj: Yes, it would be better. I will change that.

leleuj: done

PR #857, opened 2017-03-19
Update OidcConfiguration.java

ror6ax:


mmoayyed:
Why can't you just use 'ClientAuthenticationMethod.parse(auth);' directly with the existing setter? Is this just about adding a convenience method that saves  a few keystrokes or does it provide some other value?

leleuj:
@ror6ax thanks for the PR

@mmoayyed in Shiro (with a shiro.ini file), this is necessary to be able to set the `ClientAuthenticationMethod`

PR #858, opened 2017-03-21
Use getter/setter in OidcConfiguration

julioa:
Hi,
I have an small fix for OidcConfiguration that makes proper use of properties using its getter/setter methods instead of access directly to the variables .


leleuj:
OK. Is it a bug you found somewhere or you want to override some of the getters?

julioa:
Hi,
Kind of both, I found there was a problem in the keycloack implementation due to the way it was accessing some properties and I decided to fix all of them in order to allow all properties to be overrode.

leleuj:
OK. Thanks.

Issue #871, opened 2017-03-25
UserProfile::setLinkedId should allow null value


millross:
For decoding of Json-stored user profile data in vertx sessions. I can build a custom json-serializer for UserProfile but really shouldn't have to maintain one. 

I can't see any clear reason why prevent this. At the very least if you want to prevent unsetting of a linked id, you should wrap the null check on the provided linked id in code to validate we have already been linked. 

I'll submit a PR but until this is done, vertx-pac4j will remain broken (and this is a blocker to fixing it).

millross:
I'm working on this and will submit a PR soon

leleuj:
My idea was that it doesn't make any sense to link an account to another account having the `null` identifier. That said, I did not provide any way to remove the link so there is an issue in my reasoning  here.
You can safely remove the check.

leleuj:
Reverted the master to 2.0.0-RC3-SNAPSHOT...

PR #872, opened 2017-03-26
Issue #871 UserProfile::setLinkedId should allow null value

millross:
It should be permissible (for Json deserialization) to set the linked id to null if it's already null, since this is a no-op.

Assuming the null check is in place to prevent un-settiing of a linked id, then it should only be enforced if the existing linked id as non-null.

Also added tests which validate this behaviour.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
@@ -312,10 +312,6 @@ public abstract class UserProfile implements Serializable, Externalizable {
     }
 
     public void setLinkedId(final String linkedId) {
-        if (this.linkedId != null) {
-            // Don't allow unsetting of already set linked id
-            CommonHelper.assertNotNull("linkedId", linkedId);
-        }
         this.linkedId = linkedId;
     }
 }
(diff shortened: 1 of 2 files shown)


pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
```diff
@@ -316,7 +312,10 @@ public String getLinkedId() {
     }
 
     public void setLinkedId(final String linkedId) {
-        CommonHelper.assertNotNull("linkedId", linkedId);
+        if (this.linkedId != null) {
```
leleuj: If we want to remove a link to an account, we should allow to set `null`. We should just remove the `CommonHelper.assertNotNull` test.

millross:
I’m ok with that, I just assumed there was a reason for the assertNotNull test in the first place. I’ll push with that change tomorrow (I’m not going to have any time for OSS coding this evening).

leleuj:
No problem. Things were not clear in my mind too, until you made me think about it.

millross:
Removed the null check altogether in the latest commit, as it now appears redundant.

leleuj:
Thanks

millross:
Thanks, I'll get the latest vertx-pac4j working against snapshot tomorrow evening, and commit the relevant changes.

PR #879, opened 2017-03-29
Enforce use of commons-collections 3.2.2 in pac4j-saml

### LINKED ISSUE #878: pac4j-saml uses vulnerable commons-collections:3.2.1 (throught Velocity 1.7)
pac4j-saml depends on Velocity 1.7.
Velocity 1.7 depends on commons-collections 3.2.1.
commons-collection 3.2.1 is known to have be vulnerable (CVE-2015-6420) and should not be used
While this vulnerability doesn't affect directly pac4j-saml, the transitive dependency add the vulnerable code in the classpath
As there is no 1.7.1 version of Velocity ([PR](https://github.com/apache/velocity-engine/pull/4) is open and it would seem the plan is to wait Velocity 2 before adapting pac4j-saml), a solution might be to explicitely exclude commons-collections 3.2.1 and add commons-collections 3.2.2 manually

cporte:
Fix #878

### DIFF
--- a/pac4j-saml/pom.xml
+++ b/pac4j-saml/pom.xml
@@ -16,6 +16,9 @@
         <opensaml.version>3.2.0</opensaml.version>
         <joda-time.version>2.9.2</joda-time.version>
         <velocity.version>1.7</velocity.version>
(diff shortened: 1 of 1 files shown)


leleuj:
Excellent! Thanks

Issue #892, opened 2017-04-12
Define constants for "isAuthenticated", "isRemembered", etc.


rwalkerands:
There are a lot of string literals that are intended for use in user code. E.g., `DefaultAuthorizationChecker` has `"isAuthenticated"`, `"isRemembered"`, etc.

At least for these values that are to appear in user code, please define constants for them, so that I can say, e.g., something like:

```
    @Pac4JSecurity(authorizers = AuthorizationConstants.IS_AUTHENTICATED)
```

leleuj:
Yes, it would be better. We already have the `HttpConstants` and `Pac4jConstants` classes, but it could be a new one.
Feel free to submit a pull request for that.

millross:
Good call

rwalkerands:
I will do a pull request.

What is your preference?

- Add them to HttpConstants?
- Add them toPac4jConstants?
- Make a new class?

leleuj:
It does not fit well in `HttpConstants` and  `Pac4jconstants`. I would create a new `DefaultAuthorizers` class for this.

PR #894, opened 2017-04-13
Define constants for authorizers (#892)

rwalkerands:
Create a DefaultAuthorizers interface with constants for the
authorizers recognised by DefaultAuthorizationChecker.
Use them in DefaultAuthorizationCheckerTests.

One subtle difference: it was "securityHeaders" in the test, versus "securityheaders" in the main code.


leleuj:
Excellent! About documentation: can we add a sentence to advertise about this class in: https://github.com/pac4j/pac4j/blob/master/documentation/docs/authorizers.md?

To locally browse the documentation on `http://localhost:4000`, type: `cd documentation; bundle exec jekyll serve` (assuming you have installed jekyll: https://jekyllrb.com/docs/installation/)

rwalkerands:
Sentence added. (But link will not work until after the merge.)

PR #904, opened 2017-04-27
add sql index in doc

### LINKED ISSUE #903: index on the sql backend
The [doc for sql](http://www.pac4j.org/docs/authenticators/sql.html) currently only states to create a table. In my experience, creating indexes on the fields that will be queried significantly improves performance. Looking at the code, I think asking users to also run

```sql
ALTER TABLE `users` ADD UNIQUE KEY `username` (`username`), ADD UNIQUE KEY `linkedid` (`linkedid`);
```

would be helpful (removing the `UNIQUE` keyword if entries are not supposed to be unique). I can't really test though.

eroux:
This fixes #903. This is not tested though, we don't use sql.

### DIFF
--- a/docs/authenticators/sql.md
+++ b/docs/authenticators/sql.md
@@ -48,6 +48,11 @@ CREATE TABLE users
   linkedid varchar(255),
   serializedprofile varchar(10000)
 );
+
+ALTER TABLE users
+	ADD PRIMARY KEY (id), 
+	ADD KEY username (username), 
+	ADD KEY linkedid (linkedid);
 ```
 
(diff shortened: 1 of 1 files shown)


leleuj:
This won't work in fact, the PR should target the master and this file: https://github.com/pac4j/pac4j/blob/master/documentation/docs/authenticators/sql.md

In the build process, they are copied to the gh-pages branch: https://github.com/pac4j/pac4j/blob/master/.travis.yml#L21

eroux:
Ok, maybe a CONTRIBUTING.md would help?

leleuj:
Yes, I keep that on mind.

leleuj:
Here it is: https://github.com/pac4j/pac4j/blob/master/CONTRIBUTING.md

Issue #886, opened 2017-04-04
OIDC : provider metadata ignored and suggesting improving configuration 


andre-doherty:
hello,
It looks like some elements available in the SP metadata are ignored such as : 
- scopes_supported
- id_token_signing_alg_values_supported 

=> The client integration would be simpler if there was less configuration in the shiro.ini file. 

Also it might be useful if the metadata could be loaded from a resource instead of being fetched from a distant url : the discoveryURI setting is somehow compulsory however the sp doesnt always serve such a file).

leleuj:
Yes, there is always room for improvement. Maybe we could re-use the resource definition mechanism we already have for SAML for example.

gpgekko:
The newly added `logoutUrl` could also be optimised to use the available metadata under `getEndSessionEndpointURI()`.

leleuj:
Sure. To be added in the future pull request...

leleuj:
We started the development of version 3.0 only a few weeks ago, so it's perfect for this contribution...

leleuj:
@andre-doherty : any update on this?

andre-doherty:
Hello Jérôme,

Sorry i haven't been able to find the time to cover that issue yet.
I should hopefully get back into that topic in the next coming months and
then i will be than happy to contribute to the project.

Regards

Andre

leleuj:
No problem. Thanks for the follow-up.

As I'm sorting issues, I close this one. Feel free to re-open it with your contribution.

PR #912, opened 2017-05-18
OIDC use provided data (#886)

gpgekko:
Uses the end session endpoint url from the provider meta data as fallback, if available.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
@@ -264,6 +264,9 @@ public class OidcConfiguration extends InitializableWebObject {
     }
 
     public String getLogoutUrl() {
+        if(logoutUrl == null && getProviderMetadata().getEndSessionEndpointURI() != null) {
+            return getProviderMetadata().getEndSessionEndpointURI().toString();
+        }
         return logoutUrl;
     }
 


pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
```diff
@@ -264,6 +264,9 @@ public void setResponseMode(final String responseMode) {
     }
 
     public String getLogoutUrl() {
+        if(logoutUrl == null && getProviderMetadata().getEndSessionEndpointURI() != null) {
+            return getProviderMetadata().getEndSessionEndpointURI().toString();
```
gpgekko: I didn't want to edit the rest of the code without an OK from you @leleuj, but right now I'm doing a toString on an java.net.URI here, only to parse it as a the same class at a later time when it is actually used. Seems inefficient.

gpgekko:
I'm not sure if this is the best way to do this, but it's basically what I did in an extended class locally and I figured it could serve as a starting point.

pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
leleuj: In 2.1, we must remain backward compatible with 2.0, which is the case. Let's not go further and merge it as is.

Issue #915, opened 2017-05-23
OIDC authorization code flow does not work for SPAs


tookko:
When using OpenID Connect authorization code flow, a login request is made from the frontend to the pac4j-oidc callback at the backend, and it should redirect the frontend to the identity provider's login page. A simple HTTP redirect is fine for traditional web applications.

However, in case of modern Single Page Applications calls to the backend are typically made with XMLHttpRequests. In a standards-conforming browser any HTTP redirections with XMLHttpRequests are automatically followed by the browser without giving control back to the frontend JavaScript. This is a problem when frontend login functionality is done with an AJAX call, or simply when a session times out and the user should be re-authenticated by redirecting them to the login page after an AJAX call.

To work around this, by default the IndirectClient implementation in pac4j-oidc will return a 401 instead of redirection if the AJAX request has e.g. X-Requested-With: XMLHttpRequest header set to identify it as an AJAX request. This seems to be a typical way of handling with this problem, except that the URL of the login page should be passed to the frontend as well. Otherwise the frontend will have no way of knowing the login page URL without out-of-band information.

Suggestion to fix is to include the same URL that a normal redirect would use in a suitable header field instead of just returning a simple HTML page with unhelpful text "HTTP Error 401" that a typical AJAX client would ignore or mishandle in any case.

Tested with pac4j 2.0.0 and buji-pac4j 2.1.0-RC1.

tookko:
An appropriate header field could be "Referer"?

tookko:
An appropriate header field could be Location?

leleuj:
Notice that buji-pac4j v2.1 has not been upgraded yet to pac4j v2.0, so if it works without any issue, it's luck.

So you finally expect the front to redirect the user to the login URL found in the Location header of the 401 error, don't you?

This will work most of the time, but we cannot address some SAML configuration as it might be a form POST instead of a URL redirection.

tookko:
Yes, the frontend would need to do the redirect in this case which is not ideal, but I couldn't come up with a better way to tackle that and it's what seems to be the commonly suggested solution.

It's absolutely true that this will not cover everything 100% e.g. with SAML. But as you said at least it would work most of the time and be useful for some popular protocols like OpenID Connect, which is better than it not working at all. And the change should be pretty safe from not breaking existing clients.

WDYT, do you agree on the suggested solution, or any better ideas how we could tackle AJAX calls? If there's a green light that this would be useful, we could work on a PR.

[...]

PR #917, opened 2017-05-24
Redirect location in case of ajax calls

andreaTP:
Fix for #915 
It contains 2 commit, one is a small fix to a pom the other is the optional redirect url to be incorporated into 401 answer


leleuj:
Merged. Thanks

Issue #548, opened 2016-04-14
Pac4j Kerberos support


garpinc:
I am faced with having to implement kerberos support in a pac4j/spring security enabled app. I have a decision point where either I use [spring-security-kerberos](http://projects.spring.io/spring-security-kerberos/) directly or If I need to start a pac4j project to support it. Any insight would be appreciated.

leleuj:
I don't think spring-security-kerberos is a bad solution, Spring projects are generally good ones, though this one is not mainstream and the last commit is from February.
I have already received several requests about Kerberos support in pac4j. Adding Kerberos support to pac4j is a guarantee to be able to use Kerberos almost in every Java framework.

So I think it's mostly up to you: either you directly use the Spring project or you can spend time and contribute a new `pac4j-kerberos` module. This new module would be more than welcome.
I guess we would use direct clients like the basic auth or digest ones and have a specific `KerberosAuthenticator`. With further reflection, I think we can even leverage the CAS Kerberos support on the pac4j one? @mmoayyed

garpinc:
Here's what I've come up with so far.. I'll make a module later if I can get it to work and if we can agree on direction. Specifically I'm told i need to support Spnego Negotiate.
[kerberos.zip](https://github.com/pac4j/pac4j/files/220869/kerberos.zip)

leleuj:
I just took a quick look, but it seems fairly easy. But you use the Spring Security Kerberos module, don't you (I see: `org.springframework.security.kerberos.authentication.KerberosTicketValidator`)
If we want to create a `pac4j-kerberos` module, it must not depend on Spring Security, but it's certainly more work. Though, what you created is perfectly useful in the Spring Security pac4j implementation.

garpinc:
If there is some ticket validator elsewhere (i.e: in CAS) I may be able to leverage that but perhaps it may need to be pulled out of that module into a common module.

leleuj:
The service ticket validator in CAS is very specific, I don't think we could re-use it. Generally, for a new protocol support, I try to find the best plumbery library and build the pac4j module on top of that.

mmoayyed:
I don't see much point in doing this. Kerb support is almost universally handled via JAAS, and there are plenty modules that already deal with Kerb and Jaas. Why would we want to repeat that experience when it's already done many many times?

leleuj:
@mmoayyed  : so you do think that having a `pac4j-kerberos` module does not make any sense at all, don't you?

[...]

PR #595, opened 2016-06-14
Kerberos implementation

garpinc:
I know you guys said you won't accept a kerberos implementation for whatever reasons.. But I wanted to make sure that the work I've done is somewhere. We are using a 1.8 pac4j version of this with our clients. The rationale in my mind for this is as follows. We create software that's distributed to multiple clients and each of them have there own security requirements. Our software allowed them to configure in the kind of security they want and we're using pac4j as that library. Even though kerberos may be supported via other mechanisms using pac4j as the common integration layer provides value for us. I'm not sure how to implement unit tests yet and until I know you're going to include it there's no point making the effort.

PR #919, opened 2017-05-25
Add improved pac4j-kerberos

vidma:
- [x] Rebase https://github.com/pac4j/pac4j/pull/595/ over master, fixing changes to the APIs
  * DirectClientV2, io.Resource directly from springframework etc
  * update `since 2.1.0`
- [x] Cleanup/simplify code, indentation, etc
  * Use regular HttpAction instead of a custom DeferredHttpAction
  * Fix bug when absolute keytab path was lost (using only filename)
- [x] Add integration/unit test with Kerby
  * mocked web-context
  * real Kerberos keytabs and real SPNEGO tickets
- [x] Add/split into `IndirectKerberosClient` and `DirectKerberosClient`, [discussion here](https://github.com/pac4j/pac4j/pull/919#discussion_r118471748)
- [x] Update/add tests for Direct/Indirect
- [ ] ~~Direct client with FORCE_AUTH_IF_NO_CREDENTIALS option (postponed, can be implemented easily without tweeking core code)~~.
- [x] Add docs
  * basically as done in `setupKerberosClient`, could also describe some common pitfalls:
    - what should be the service pricipal, i.e. `HTTP/full-qualified-domain-name@DOMAIN`
    - Java Cryptography Extension (JCE)
- [x] test again with `play-pac4` example with a real Kerberos setup. works fine.

fixes https://github.com/pac4j/pac4j/issues/548
@leleuj   P.S. my actual changes are only in the last two commits.


pac4j-kerberos/src/test/java/org/pac4j/kerberos/client/direct/KerberosClientKerbyTest.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * This is a test with real kerberos ticket validation:
+ * - spins up a mini/lightweight Kerberos server (using Apache Kerby)
+ * - generates the keytabs for service and client
+ * - generates a real service ticket, and passes it to mock context
+ * - checks that the ticket is correctly validated, and yields a correct client ID
+ *
+ * @author Vidmantas Zemleris, at Kensu.io
+ */
+public class KerberosClientKerbyTest implements TestsConstants {
```
vidma: it's somewhere between integration and unit test.
likely I should make it run together with the integration tests...

pac4j-kerberos/src/main/java/org/pac4j/kerberos/profile/KerberosProfile.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Represents a user profile based on a Kerberos authentication.
+ *
+ * @author Garry Boyce
+ * @since 2.1.0
+ */
+public class KerberosProfile extends CommonProfile {
+
+    private static final long serialVersionUID = -1388563485891552197L;
+    public GSSContext gssContext = null;
```
vidma: added optional `gssContext` for more advanced usage, if any.
if `SunJaasKerberosTicketValidator.setHoldOnToGSSContext(true)` was called, when this would allow subsequent encrypt/decrypt operations.  see https://github.com/pac4j/pac4j/pull/919/files#diff-c6ef70675c7cb3b2df3dde625bac09ecR124

pac4j-kerberos/src/main/java/org/pac4j/kerberos/client/direct/KerberosClient.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    protected void clientInit(final WebContext context) {
+        setCredentialsExtractor(new KerberosExtractor(getName()));
+        // FIXME: is this needed
+        assertAuthenticatorTypes(KerberosAuthenticator.class);
+    }
+
+    // FIXME: this was copied from earlier version
+    protected void assertAuthenticatorTypes(final Class<? extends Authenticator>... classes) {
```
vidma: this fn was part of pac4j `DirectClient*`, but was removed. do you remember why? @leleuj

leleuj: This was a check on the `Authenticator` type, but it was definitely too complicated and a bit too stringent.

During the development of the 2.0.0 version, I finally decided to remove it. You can delete it also.

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/authenticator/KerberosAuthenticator.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param ticketValidator The ticket validator used to validate the Kerberos ticket.
+     * @since 2.1.0
+     */
+    public KerberosAuthenticator(KerberosTicketValidator ticketValidator) {
+        this.ticketValidator = ticketValidator;
+    }
+
+    @Override
+    public void validate(KerberosCredentials credentials, WebContext context) throws CredentialsException {
+        logger.trace("Try to validate Kerberos Token:" + credentials.getKerberosTicketAsString());
```
leleuj: The call to the `init()` method is missing here.

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/authenticator/KerberosAuthenticator.java
```diff
@@ -0,0 +1,79 @@
...
+            subject = KerberosProfile.class.getSimpleName() + UserProfile.SEPARATOR + subject;
+        }
+        logger.debug("Succesfully validated, updated subject: " + subject);
+    }
+
+    @Override
+    protected void internalInit(WebContext context) {
+        CommonHelper.assertNotNull("ticketValidator", this.ticketValidator);
+        if (this.ticketValidator instanceof InitializableObject) {
+            ((InitializableObject) this.ticketValidator).init();
```
leleuj: This is the way I used to initialize objects, but I tend to do things differently right now: initialization must be handled by the component which is initializable, not by another one using it; otherwise, initialization may be forgotten.

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/authenticator/SunJaasKerberosTicketValidator.java
```diff
@@ -0,0 +1,251 @@
...
+        try {
+            return Subject.doAs(this.serviceSubject, new KerberosValidateAction(token));
+        } catch (PrivilegedActionException e) {
+            throw new BadCredentialsException("Kerberos validation not successful", e);
+        }
+    }
+
+    @Override
+    protected void internalInit() {
+        try {
```
leleuj: We should call the `init()` method first, instead of letting the `KerberosAuthenticator` do that for us.

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/authenticator/SunJaasKerberosTicketValidator.java
```diff
@@ -0,0 +1,251 @@
...
+    private String servicePrincipal;
+    private Resource keyTabLocation;
+    private Subject serviceSubject;
+    private boolean holdOnToGSSContext;
+    private boolean debug = false;
+    protected final Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Override
+    public KerberosTicketValidation validateTicket(byte[] token) throws BadCredentialsException {
+        try {
```
leleuj: We should call the `init()` method first, instead of letting the KerberosAuthenticator do that for us.

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/authenticator/KerberosAuthenticator.java
```diff
@@ -0,0 +1,79 @@
...
+
+        String subject = ticketValidation.username();
+        logger.debug("Succesfully validated " + subject);
+
+        KerberosProfile profile = new KerberosProfile();
+        profile.setId(subject);
+        profile.gssContext = ticketValidation.getGssContext();
+        credentials.setUserProfile(profile);
+
+        // FIXME: which one do we prefer to use a profile ID? `KerberosProfile#garry vs garry` ?
```
leleuj: Can you elaborate a little bit on this?

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/extractor/KerberosExtractor.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public KerberosExtractor(final String clientName) {
+        this.clientName = clientName;
+    }
+
+    @Override
+    public KerberosCredentials extract(WebContext context) throws CredentialsException, HttpAction {
+        final String header = context.getRequestHeader(HttpConstants.AUTHORIZATION_HEADER);
+        if (header == null) {
+            throw HttpAction.unauthorizedNegotiate("Kerberos Header not found", context);
```
vidma: so, @leleuj , looking at other extractors/clients,  I guess this exception throwing is not the responsibility of `KerberosExtractor` and should be moved to `Kerberos(In)?DirectClient`.

there is one semantic nuance with SPNEGO/Kerberos - the browser (e.g. Firefox) do NOT send the credentials (`Authorization: Negotiate someTicket`) if server did not provide a `WWW-Authenticate: Negotiate` header. 

Thus a trully DirectClient don't quite make sense here. As it is now, the 401 indicates to repeat the request with credentials, but does do a redirect, so it's in between direct and indirect, I guess?

---
in my case, I'd like to combine session/cookie storage with Kerberos, so would not need renegotiate with every request. so in brief what I'd like to have sounds like indirect client.

would this happen automatically if it implemented an indirect client?

pac4j-kerberos/src/main/java/org/pac4j/kerberos/client/direct/KerberosClient.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public KerberosClient(final Authenticator authenticator, final ProfileCreator<KerberosCredentials, KerberosProfile> profileCreator) {
+        setAuthenticator(authenticator);
+        setProfileCreator(profileCreator);
+    }
+
+    @Override
+    protected void clientInit(final WebContext context) {
+        setCredentialsExtractor(new KerberosExtractor(getName()));
```
leleuj: We should have `defaultCredentialsExtractor(new KerberosExtractor(getName()));` to setup the `KerberosCredentialsExtractor` if none has already been set.

pac4j-kerberos/src/main/java/org/pac4j/kerberos/profile/KerberosProfile.java
leleuj: OK

pac4j-kerberos/src/test/java/org/pac4j/kerberos/client/direct/DummyAuthenticator.java
```diff
@@ -0,0 +1,14 @@
+package org.pac4j.kerberos.client.direct;
+
+import org.pac4j.core.context.WebContext;
+import org.pac4j.core.credentials.Credentials;
+import org.pac4j.core.credentials.authenticator.Authenticator;
+import org.pac4j.core.exception.HttpAction;
+
+public class DummyAuthenticator implements Authenticator<Credentials> {
```
leleuj: I generally don't create this kind of class but use lambdas for brevity in tests. That said, it's not a big deal.

pac4j-kerberos/src/test/java/org/pac4j/kerberos/client/direct/KerberosClientKerbyTests.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Before
+    public void before() throws KrbException, IOException {
+        setupKerbyServer();
+    }
+
+    private void setupKerbyServer() throws KrbException, IOException {
+        kerbyServer = new SimpleKdcServer();
+        kerbyServer.setKdcHost("localhost");
+        kerbyServer.setKdcRealm("MYREALM.LT"); // FIXME
```
leleuj: `FIXME`?

pac4j-kerberos/src/test/java/org/pac4j/kerberos/client/direct/KerberosClientTests.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Test
+    public void testMissingProfileCreator() {
+        final KerberosClient kerberosClient = new KerberosClient(kerberosAuthenticator);
+        kerberosClient.setProfileCreator(null);
+        TestsHelper.initShouldFail(kerberosClient, "profileCreator cannot be null");
+    }
+
+    @Test
+    public void testBadAuthenticatorType() {
```
leleuj: Finally, this can be removed as well as the `DummyAuthenticator`.

pac4j-core/src/main/java/org/pac4j/core/exception/HttpAction.java
```diff
@@ -112,6 +112,19 @@ public static HttpAction unauthorizedDigest(final String message, final WebConte
...
 
     /**
+     * Build a response requesting to provide credentials via Kerberos/SPNEGO Negotiate mechanism.
+     *
+     * @param message message
+     * @param context context
+     * @return 401 Unauthorized with "WWW-Authenticate: Negotiate"
+     */
+    public static HttpAction unauthorizedNegotiate(final String message, final WebContext context) {
+        context.setResponseHeader(HttpConstants.AUTHENTICATE_HEADER, "Negotiate");
```
leleuj: Maybe `Negotiate` would deserve a constant.

pac4j-kerberos/src/test/java/org/pac4j/kerberos/client/direct/KerberosClientKerbyTests.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * This is a test with real kerberos ticket validation:
+ * - spins up a mini/lightweight Kerberos server (using Apache Kerby)
+ * - generates the keytabs for service and client
+ * - generates a real service ticket, and passes it to mock context
+ * - checks that the ticket is correctly validated, and yields a correct client ID
+ *
+ * @author Vidmantas Zemleris, at Kensu.io
+ */
```
leleuj: `@since 2.1.0`

pac4j-kerberos/src/test/java/org/pac4j/kerberos/client/direct/KerberosClientTests.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.servlet.http.HttpServletResponse;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.*;
+
+
+public class KerberosClientTests implements TestsConstants {
```
leleuj: Javadoc...

leleuj:
Overall, this is excellent. I made a few comments here and there, some things need to be fixed.

The tests are excellent as well.

The documentation is missing: the `README.md` must reference the Kerberos support and the website must be updated (new authenticator) via the files in the `documentation` directory. Please read: https://github.com/pac4j/pac4j/tree/master/documentation.

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/authenticator/KerberosAuthenticator.java
vidma: is there any use of having `KerberosProfile#` prefix in profile ID (was in original code).
so I guess just delete the `L56-59` below.
earlier `subject/profileID = "garry"` after second assignment `subject/profileID=KerberosProfile#garry`.

pac4j-kerberos/src/main/java/org/pac4j/kerberos/client/direct/KerberosClient.java
vidma: so to allow the default to be overridden...

vidma:
so it seems to me I should split `KerberosClient` into two:
- **`IndirectKerberosClient`** - behaves like now, just extends `IndirectClient`. when used with `forceLogin`q, it seems to give this session persistence what I wanted.
- `DirectKerberosClient` - behaves like now, mostly.
  * not yet sure what it should do if no `Authenticate: Negotiate SomeTicket` header present:
     - `throw a HttpAction` in `retrieveCredentials` (like in `IndirectBasicAuthClient`) 
     - or simply`Extractor.extract() { return null }` like in `DirectBasicAuthClient`

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/extractor/KerberosExtractor.java
leleuj: Yes, you're completely right. The direct client is meant to take the credentials provided with the request and not break the flow if there are no credentials. So you should have no `throw HttpAction.unauthorizedNegotiate` in the direct client.

To keep the authenticated user profile in session, the indirect client seems to be the right choice, but the indirect client also redirects to the identity server for login (at the beginning of the flow). In the case of basic auth, the login URL is the callback URL (where credentials are expected) and in that case, if credentials are missing, a 401 error is returned.

My guess is that you should have something really really close to the basic auth support: `DirectBasicAuthClient`, `BasicAuthExtractor` and `IndirectBasicAuthClient`.

The `KerberosExtractor` should be shared among both clients, should not throw any `HttpAction` while the `KerberosIndirectClient` should handle that.

My only question is whether a direct client makes sense for Kerberos, but you know better...

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/authenticator/KerberosAuthenticator.java
leleuj: No, when setting an identifier, you don't need to use the `#` separator, but when you request the typed identifier, you get `the class name + "#" + the identifier`.

pac4j-kerberos/src/main/java/org/pac4j/kerberos/client/direct/KerberosClient.java
leleuj: This won't work: `setCredentialsExtractor` sets the credentials extractor while `defaultCredentialsExtractor `sets the credentials extractor only if it is null (if is has not been set).

Example:

```java
KerberosClient client = new KerberosClient();
client.setAuthenticator(mySpecificAuthenticator);
```

`clientInit` is only called after the setter, so the authenticator must not be set to the default one, except if it is null.

leleuj:
To sum up, you should have only one `KerberosExtractor` shared among the `DirectKerberosClient` and the `IndirectKerberosClient`. In the latter case, the indirect client should throw the 401 Negotiate response.

See the basic auth support, it seems to be the same...

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/extractor/KerberosExtractor.java
vidma: yeah makes sense.
at least this time we gonna use an `IndirectClient`, so that's what I'll be focusing on.

> My only question is whether a direct client makes sense for Kerberos, but you know better...


good question for which I don't know a full answer. but if one is using Kerberos as a webservice/headless auth, the current `401` flow in direct client might make sense more than redirect with indirect client.
So to keep more flexibility for others, I'd just add an option in `DirectClient` -
 `FORCE_AUTH_IF_NO_CREDENTIALS` if true, when credentials are missing, a 401 should be returned.

pac4j-core/src/main/java/org/pac4j/core/exception/HttpAction.java
vidma: dunno, all other fns use string values too...

pac4j-kerberos/src/main/java/org/pac4j/kerberos/credentials/extractor/KerberosExtractor.java
leleuj: Yes, good option

pac4j-core/src/main/java/org/pac4j/core/exception/HttpAction.java
leleuj: OK. Keep things that way

leleuj:
The build is broken due to PMD checks: `PMD Failure: org/pac4j/kerberos/credentials/authenticator/KerberosAuthenticator.java:8 Rule:UnusedImports Priority:4 Avoid unused imports such as 'org.pac4j.core.util.InitializableObject'.`

vidma:
yeah, work still in progress.

so I think Direct/Indirect split should be +/- as it is. just I may add an optional `401` to force auth (as otherwise sometimes no ticket might be sent at all...), to DirectClient too. what do you think?

leleuj:
"just I may add an optional 401 to force auth (as otherwise sometimes no ticket might be sent at all...), to DirectClient too. what do you think?": yes, but auth would not be forced by default

vidma:
So currently this PR behaves very similarly to the BasicAuth.
As Direct and Indirect clients behave similarly in terms of Kerberos testing, I decided to test both in KerberosClientsKerbyTests.

still would like to retest with `play-pac4j` (getting some sort of jarhell; will check this on ~Monday), otherwise this is mostly finished, finally.

--

P.S. for `DirectClient` I decided **not to add** the `optional HTTP401 with WWW-Authenticate: Negotiate` yet. 
  * if someone has a clear use-case and needs, they can use it even without modifying `pac4j-kerberos` code, simply by defining their own `DirectClient`.

leleuj:
You don't need to squash the commits, I'll use the appropriate Github option for the merge.

Yes, regarding the DirectClient, it's better not to make it blocking (although feasible).

Is this ready for merge or is there anything left?

vidma:
yeah, this OK ready to merge now (changed 3 lines of docs recently).
tested OK with `play-pac4j` and real kerberos on Chrome, Firefox, and Safari.

vidma:
@leleuj btw,  there's a related thing to disscuss for `play-pac4j` - we need `pac4j-kerberos` with play 2.3 or 2.4 (as play 2.5 don't support scala 2.10 anymore, and we still need scala 2.10).
  so looks like we need to backport `pac4j-kerberos` for an earlier version of `play-pac4j / pac4j`,  right (as master is for play 2.5)?

vidma:
FYI: I backported pac4j-kerberos to `pac4j 1.7.x` (as required by the damn old play 2.3 we are still locked to) https://github.com/vidma/pac4j/pull/1

leleuj:
Excellent work! Thanks. Merging.

Issue #916, opened 2017-05-23
Integration with simpleSAMLphp


len-ro:
Hello, I would like to contribute my initial experience in integrating this library with simpleSAMLphp via the j2e-pac4j-demo.

This is a more complex library. There is also a demo application for j2e.

To clone, compile and run this demo the sequence is straight forward:

git clone https://github.com/pac4j/j2e-pac4j-demo
cd j2e-pac4j-demo
mvn package
mvn jetty:run
The test with https://www.testshib.org works.

To configure if for simpleSAMLphp. Modify DemoConfigFactory.java

```java
final SAML2ClientConfiguration cfg = new SAML2ClientConfiguration("resource:samlKeystore.jks",
 "pac4j-demo-passwd",
 "pac4j-demo-passwd",
 "resource:idp-metadata.xml");
 cfg.setMaximumAuthenticationLifetime(3600);
 cfg.setServiceProviderEntityId("test.pac4j");
 cfg.setServiceProviderMetadataPath(new File("sp-metadata.xml").getAbsolutePath());
 final SAML2Client saml2Client = new SAML2Client(cfg);
```
The idp-metadata.xml file is the file from: http://idp-domain/simplesamlphp/saml2/idp/metadata.php?output=xhtml wrapped in an additional EntitiesDescriptors element:

```xml
<md:EntitiesDescriptor xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata" xmlns:ds="http://www.w3.org/2000/09/xmldsig#"> 
 <md:EntityDescriptor entityID="http://idp-domain/simplesamlphp/saml2/idp/metadata.php">
...
```

However at this point the application gives a "fatal error":

```org.pac4j.saml.exceptions.SAMLException: Identity provider has no single sign on service available for the selected profileorg.opensaml.saml.saml2.metadata.impl.IDPSSODescriptorImpl@2d6719d3```

The error seems to be here https://github.com/pac4j/pac4j/blob/master/pac4j-saml/src/main/java/org/pac4j/saml/context/SAML2MessageContext.java#L104 so I am left with no clue to the problem. The only solution is to change a bit the code to see which is the binding required.

Just cloning the main repository and trying to compile it with maven does not work. The error is:

```
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[...]

PR #921, opened 2017-05-27
#916

len-ro:
As suggested in the response to #916 I created a pull request for the small changes in SAML2MessageContext which allowed me identify the missing binding by replacing the error:
```
org.pac4j.saml.exceptions.SAMLException: Identity provider has no single sign on service available for the selected profileorg.opensaml.saml.saml2.metadata.impl.IDPSSODescriptorImpl@2d6719d3
```
with a much more clear:

```
org.pac4j.saml.exceptions.SAMLException: Identity provider has no single sign on service available for the selected profileurn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST
```

And the doc update for SimpleSAMLphp.

Thank you.


leleuj:
Thanks

PR #922, opened 2017-05-29
fix public api

andreaTP:
Unfortunately during this cherry pick I managed to break the public API, this is a fix for the error.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/exception/HttpAction.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/exception/HttpAction.java
@@ -79,6 +79,19 @@ public class HttpAction extends Exception {
      * @param message message
      * @param context context
      * @param realmName realm name
+     * @return a basic auth popup credentials
+     */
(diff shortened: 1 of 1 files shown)


leleuj:
I guess we'll need a 1.9.9 version just after merging that, won't we?

andreaTP:
Yes, this time not in a hurry for me, since I already have a proper release.
Probably some "release notes are needed", sorry for the inconvenience.

leleuj:
Yes, maybe, I'll even skip the 1.9.8 in the release notes. Sometimes, one release is a failure...

leleuj:
We have the same issue for version 2.1.0. Since v2.0.0, we user semver so v2.1.0 should be backwards compatible with v2.0.0. Can you push the same fix for the master? Thanks

leleuj:
Thinking about that a little more, I don't think the method should be marked as `Deprecated`. I'll change that.

leleuj:
I'm releasing the 1.9.9 version...

PR #924, opened 2017-05-30
Add ip extractor for multiple headers in chain

giflw:
- Add IpHeaderChainExtractor to search for ip header
- Add tests


leleuj:
In fact, I'd prefer to only have the `IpExtractor`: can you integrate your changes in this existing class? Thanks

giflw:
I made the merge in a backward compatible manner, and deprecated the set/get alternateHeader. I also added and option to set the proxy IP, so the source of the request can be checked before searching the headers. This way a fake or malicious proxy can be detected and blocked. I just don't know if this should be enable by default and we should use a single proxy IP or a list. For proxy check, could also be added an options to check an token on an custom header, but I don't know if this is useful either. I'm just concerned with security in this case.

What do you think about it?

Thanks

pac4j-http/src/main/java/org/pac4j/http/credentials/extractor/IpExtractor.java
```diff
@@ -37,16 +59,77 @@ public TokenCredentials extract(WebContext context) throws HttpAction {
...
+        return Collections.unmodifiableList(this.alternateIpHeaders);
+    }
+
+    /**
+     * @param alternateIpHeaders Sets alternate headers to search for IP.
+     *                           The first match will be returned as specified for {@code enhanced for} iteration over arrays.
+     * @since 2.1.0
+     */
+    public void setAlternateIpHeaders(final String... alternateIpHeaders) {
+        Objects.requireNonNull(alternateIpHeaders, "Ip headers must be not null");
```
leleuj: Use `CommonHelper.assertNotNull` for consistency with other checks.

leleuj:
I don't think we need to mark the simple setter/getter as `@Deprecated`, they still make sense as one header is the most common use case.

I see your check on the proxy IP: if ever we have defined a `proxyIp`, it must be the source IP unless we don't get the client IP from the header(s).
As the `proxyIp` is empty by default, it's backwards compatible so I have no problem with this change.

You should add that in the documentation as well.

I always have the feelings that such check (I mean about the proxy) should be down at the network level (in low-level equipment) but anyway, we can have this check in the code.

giflw:
I've put `@Deprecated` on the getter/setter as the getter wolud not be correct in case the user has setted more than 1 header. So, to avoid problems, they should no be used, unless on old code, where that was the only option available. This deprecation helps to avoid errors case the developer forget the `s` on the getter and lets these change be backward compatible.

leleuj:
OK. Let's keep the `@Deprecated`, we have till v3.0 to finally decide what to do.

I can't merge the PR though, the build is broken due to a test failure. Would you mind fixing it? Thanks

PR #930, opened 2017-06-07
Ini instantiable FromAttributesAuthorizationGenerator

andreaTP:
Added an empty constructor to let instantiate `FromAttributesAuthorizationGenerator` from ini configuration.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/authorization/generator/FromAttributesAuthorizationGenerator.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/authorization/generator/FromAttributesAuthorizationGenerator.java
@@ -4,6 +4,7 @@ import org.pac4j.core.context.WebContext;
 import org.pac4j.core.profile.CommonProfile;
 
 import java.util.Arrays;
+import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


pac4j-core/src/main/java/org/pac4j/core/authorization/generator/FromAttributesAuthorizationGenerator.java
```diff
@@ -17,12 +17,17 @@
...
-    private final Collection<String> roleAttributes;
+    private Collection<String> roleAttributes;
 
-    private final Collection<String> permissionAttributes;
+    private Collection<String> permissionAttributes;
 
     private String splitChar = ",";
 
+    public FromAttributesAuthorizationGenerator() {
+        this.roleAttributes = Arrays.asList(new String[] {});
```
leleuj: Can you use `new ArrayList<String>();`?

andreaTP: yes

leleuj:
Thanks

PR #932, opened 2017-06-07
ini friendly constructor for FromAttributesAuthorizationGenerator

andreaTP:
Let instatiate `FromAttributesAuthorizationGenerator` from shiro.ini


leleuj:
Thanks

PR #934, opened 2017-06-10
RestAuthenticator

### LINKED ISSUE #933: RestAuthenticator
Before reporting any issue, please use the mailing lists: http://www.pac4j.org/mailing-lists.html

leleuj:
Closes #933


pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
```diff
@@ -218,6 +218,7 @@ public static String toString(final Class<?> clazz, final Object... args) {
      * @param name the name
      * @return the prefix
      */
+    @Deprecated
```
victornoel: maybe add a doc to explain what is to be used instead?

pac4j-couch/pom.xml
```diff
@@ -26,6 +26,11 @@
 			<artifactId>org.ektorp</artifactId>
 			<version>${ektorp.version}</version>
 		</dependency>
+        <dependency>
```
victornoel: indentation is funny here and in other pom.xml... I think it is because it was using tabs before, but now we have an editorconfig file that enforces 2-spaces indentation... maybe it would be good to reindent all pom at one point (or change the editorconfig setting to 1-tab indentation, but it's not the best usually for xml-based language, because it's not rendered identically everywhere)

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/RestAuthenticator.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public void validate(final UsernamePasswordCredentials credentials, final WebContext context) throws HttpAction, CredentialsException {
+        init(context);
+
+        final String username = credentials.getUsername();
+        final String password = credentials.getPassword();
+        if (CommonHelper.isBlank(username) || CommonHelper.isBlank(password)) {
+            logger.error("Empty username or password");
```
victornoel: should this be an error? if a user provides the wrong data, are we going to log it everytime? for me it is an expected situation, not an error... at most a warning, but I would go for an information or even a debug log: the only reason I would need to see that information is if I'm trying to understand why something is not working as expected, not everytime a user provides wrong data in prod...
at the same time, you could also argue that it is the responsibility of another component of the system to enforce that? your choice, I just wanted to raise the point :)

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/RestAuthenticator.java
```diff
@@ -0,0 +1,142 @@
...
+ * @since 2.1.0
+ */
+public class RestAuthenticator extends ProfileDefinitionAware<RestProfile> implements Authenticator<UsernamePasswordCredentials> {
+
+    private static final Logger logger = LoggerFactory.getLogger(RestAuthenticator.class);
+
+    private static ObjectMapper mapper;
+
+    static {
+        mapper = new ObjectMapper();
```
victornoel: this should be injectable I think (with a default lazy-constructed static instance in case none is provided), because many applications have their own objectmapper configured...

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/RestAuthenticator.java
```diff
@@ -0,0 +1,142 @@
...
+
+        final Map<String, String> headers = new HashMap<>();
+        headers.put(HttpConstants.AUTHORIZATION_HEADER, HttpConstants.BASIC_HEADER_PREFIX + basicAuth);
+
+        HttpURLConnection connection = null;
+        try {
+            connection = HttpUtils.openPostConnection(new URL(url), headers);
+            int code = connection.getResponseCode();
+            if (code != 200) {
+                logger.error("Failed response: {}", HttpUtils.buildHttpErrorMessage(connection));
```
victornoel: idem with the error log: is that really a good idea to log everytime a user can't authenticate against the REST service?

pac4j-core/src/main/java/org/pac4j/core/util/CommonHelper.java
leleuj: These are methods I forgot to remove when I refactor the `Resource` support for the v2. In fact, they are `protected` so I will remove them, instead of deprecating them.

pac4j-couch/pom.xml
leleuj: Indeed, the `pom.xml` files are just a mess in terms of indentation. No problem to fix that.

That said, I'd prefer to have a 4-spaces indentation for XML files in the editorconfig as this is what was used so far (when it was used).

leleuj: Reformated all `pom.xml` files BTW

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/RestAuthenticator.java
leleuj: I turned the `mapper` as an instance of the `RestAuthenticator`and settable.

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/RestAuthenticator.java
leleuj: Yes, it's an expected error so it's better to log it at the `INFO` level (despite it could be handled before).

Changed.

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/RestAuthenticator.java
leleuj: You're right again. I slightly changed the code to log at the `INFO` level the 401 and 403 status codes and at the `WARN` level other status codes (`DEBUG`log for 200 status code).

leleuj:
Thanks for the feedbacks. PR updated.

pac4j-couch/pom.xml
victornoel: Actually, I think it is 4 by default for every file!

If it's not what happened, then maybe you need to install a plugin for your IDE that takes editorconfig into account...

pac4j-http/src/main/java/org/pac4j/http/credentials/authenticator/RestAuthenticator.java
victornoel: seems good yep! :)

victornoel:
np :)

pac4j-couch/pom.xml
victornoel: it seems to be good actually, all is perfect then :)

PR #944, opened 2017-06-20
InMemoryProfileService

### LINKED ISSUE #942: in-memory ProfileService
I think it would be very useful to have an InMemoryServiceProvider to mock a service provider in tests. This would simplify the task of test writing quite a lot. Currently I'm using a CouchServiceProvider using a mocked CouchDB to get this result, but it's uselessly cumbersome

eroux:
this is an attempt to fix #942 but two things are missing:

First, the test has an error I cannot understand: `getProfileDefinition()` called in `AbstractProfileService.internalInit()` returns `null`, is there anything obvious I missed?

I think it would be good do add something in the documentation but I cannot find where... wdyt?

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
@@ -21,71 +21,71 @@ import java.util.stream.Collectors;
 
 public class InMemoryProfileService<U extends CommonProfile> extends AbstractProfileService<U>  {
 
-	public Map<String,Map<String,Object>> profiles;
-	public Function<Object[], U> profileFactory;
+    public Map<String,Map<String,Object>> profiles;
+    public Function<Object[], U> profileFactory;
 
-	public InMemoryProfileService(final Function<Object[], U> profileFactory) {
-		this(new HashMap<>(), profileFactory);
-	}
+    public InMemoryProfileService(final Function<Object[], U> profileFactory) {
+        this(new HashMap<>(), profileFactory);
+    }
 
-	public InMemoryProfileService(final Map<String,Map<String,Object>> profiles, final Function<Object[], U> profileFactory) {
-		this.profiles = profiles;
-		this.profileFactory = profileFactory;
-	}
+    public InMemoryProfileService(final Map<String,Map<String,Object>> profiles, final Function<Object[], U> profileFactory) {
+        this.profiles = profiles;
+        this.profileFactory = profileFactory;
+    }
 
-	@Override
-	protected void internalInit(final WebContext context) {
-		CommonHelper.assertNotNull("passwordEncoder", getPasswordEncoder());
-		defaultProfileDefinition(new CommonProfileDefinition<U>(profileFactory));
-		super.internalInit(context);
-	}
+    @Override
+    protected void internalInit(final WebContext context) {
+        CommonHelper.assertNotNull("passwordEncoder", getPasswordEncoder());
+        defaultProfileDefinition(new CommonProfileDefinition<U>(profileFactory));
+        super.internalInit(context);
+    }
 
-	@Override
-	protected void insert(final Map<String, Object> attributes) {
-		final String id = (String) attributes.get(getIdAttribute());
-		logger.debug("Inserting doc id: {} with attributes: {}", id, attributes);
-		profiles.put(id, attributes);
-	}
+    @Override
+    protected void insert(final Map<String, Object> attributes) {
+        final String id = (String) attributes.get(getIdAttribute());
+        logger.debug("Inserting doc id: {} with attributes: {}", id, attributes);
+        profiles.put(id, attributes);
+    }
 
-	@Override
-	protected void update(final Map<String, Object> attributes) {
-		final String id = (String) attributes.get(getIdAttribute());
-		logger.debug("Updating id: {} with attributes: {}", id, attributes);
-		profiles.put(id, attributes);
-	}
+    @Override
+    protected void update(final Map<String, Object> attributes) {
+        final String id = (String) attributes.get(getIdAttribute());
+        logger.debug("Updating id: {} with attributes: {}", id, attributes);
+        profiles.put(id, attributes);
+    }
 
-	@Override
-	protected void deleteById(final String id) {
-		logger.debug("Delete id: {}", id);
-		profiles.remove(id);
-	}
+    @Override
+    protected void deleteById(final String id) {
+        logger.debug("Delete id: {}", id);
+        profiles.remove(id);
+    }
 
-	private Map<String, Object> populateAttributes(final Map<String, Object> rowAttributes, final List<String> names) {
+    private Map<String, Object> populateAttributes(final Map<String, Object> rowAttributes, final List<String> names) {
         return rowAttributes.entrySet().stream()
                 .filter(p -> names == null || names.contains(p.getKey()))
                 // not using Collators.toMap because of
                 // https://stackoverflow.com/questions/24630963/java-8-nullpointerexception-in-collectors-tomap
                 .collect(HashMap::new, (m,v)->m.put(v.getKey(), v.getValue()), HashMap::putAll);
-	}
+    }
 
-	@Override
-	protected List<Map<String, Object>> read(final List<String> names, final String key, final String value) {
-		logger.debug("Reading key / value: {} / {}", key, value);
-		final List<Map<String, Object>> listAttributes;
-		if (key.equals(getIdAttribute())) {
-			listAttributes = new ArrayList<>();
-			final Map<String,Object> profile = profiles.get(value);
-			if (profile != null) {
-				listAttributes.add(populateAttributes(profile, names));
-			}
-		} else {
-			listAttributes = profiles.entrySet().stream()
-					.filter(p -> p.getValue().get(key).equals(value))
-					.map(p -> populateAttributes(p.getValue(), names))
-					.collect(Collectors.toList());
-		}
-		logger.debug("Found: {}", listAttributes);
-		return listAttributes;
-	}
+    @Override
+    protected List<Map<String, Object>> read(final List<String> names, final String key, final String value) {
+        logger.debug("Reading key / value: {} / {}", key, value);
+        final List<Map<String, Object>> listAttributes;
+        if (key.equals(getIdAttribute())) {
+            listAttributes = new ArrayList<>();
+            final Map<String,Object> profile = profiles.get(value);
+            if (profile != null) {
+                listAttributes.add(populateAttributes(profile, names));
+            }
+        } else {
+            listAttributes = profiles.entrySet().stream()
+                    .filter(p -> p.getValue().get(key).equals(value))
+                    .map(p -> populateAttributes(p.getValue(), names))
+                    .collect(Collectors.toList());
+        }
+        logger.debug("Found: {}", listAttributes);
+        return listAttributes;
+    }
 
 }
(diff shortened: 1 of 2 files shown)


eroux:
The error is the following:

```
<error message="profileDefinition cannot be null" type="org.pac4j.core.exception.TechnicalException">org.pac4j.core.exception.TechnicalException: profileDefinition cannot be null
        at org.pac4j.core.profile.service.InMemoryProfileServiceTests.setUp(InMemoryProfileServiceTests.java:53)
</error>
```

eroux:
I've fixed the error. I'm not sure it's the best way to fix it, but I couldn't find a better solution...

eroux:
This is ready for review, any comment?

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
```diff
@@ -0,0 +1,101 @@
...
+ * In-memory profile service.
+ *
+ * @author Elie Roux
+ * @since 2.0.0
+ */
+
+public class InMemoryProfileService<U extends CommonProfile> extends AbstractProfileService<U>  {
+
+	public Map<String,Map<String,Object>> profiles;
+	//public AuthenticatorProfileCreator<UsernamePasswordCredentials, U> creator;
```
victornoel: remove commented line :)

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
```diff
@@ -0,0 +1,101 @@
...
+	public Class<U> typeArgumentClass;
+
+	public InMemoryProfileService(Class<U> typeArgumentClass) {
+		this(new HashMap<String,Map<String,Object>>(), typeArgumentClass);
```
victornoel: Use the diamond operator instead of the type parameters (`new HashMap<>()`

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * @author Elie Roux
+ * @since 2.0.0
+ */
+
+public class InMemoryProfileService<U extends CommonProfile> extends AbstractProfileService<U>  {
+
+	public Map<String,Map<String,Object>> profiles;
+	//public AuthenticatorProfileCreator<UsernamePasswordCredentials, U> creator;
+	public Class<U> typeArgumentClass;
```
victornoel: I think it would be cleaner to use a factory here… for example a `Supplier<U>`

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
victornoel: Or even better, a `Function<Object[], U>` which is exactly what the `CommonProfileDefinition` is expecting…

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	@Override
+	protected void deleteById(String id) {
+		logger.debug("Delete id: {}", id);
+		profiles.remove(id);
+	}
+
+	private Map<String, Object> populateAttributes(final Map<String, Object> rowAttributes, final List<String> names) {
+		final Map<String, Object> newAttributes = new HashMap<>();
```
victornoel: could be maybe a bit more elegant to use the Stream API to do that? :)

eroux: I tried

```java
	private Map<String, Object> populateAttributes(final Map<String, Object> rowAttributes, final List<String> names) {
        return rowAttributes.entrySet().stream()
                .filter(p -> names == null || names.contains(p.getKey()))
                .collect(Collectors.toMap(p -> p.getKey(), p -> p.getValue()));
	}
```

but it raises a `NullPointerException` due to [this](https://stackoverflow.com/questions/24630963/java-8-nullpointerexception-in-collectors-tomap) very weird "feature", and working around the feature makes the code quite cumbersome...

eroux:
I've addressed all your comments, tell me what you think!

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
victornoel: I didn't know this one! too bad then :) thanks for the explanation

pac4j-core/src/test/java/org/pac4j/core/profile/service/InMemoryProfileServiceTests.java
```diff
@@ -0,0 +1,138 @@
...
+
+	public final static PasswordEncoder PASSWORD_ENCODER = new ShiroPasswordEncoder(new DefaultPasswordService());
+	public final static InMemoryProfileService<CommonProfile> inMemoryProfileService = new InMemoryProfileService<CommonProfile>(x -> new CommonProfile());
+	static {
+		inMemoryProfileService.setPasswordEncoder(PASSWORD_ENCODER);
+	}
+
+
+	@BeforeClass
+	public static void setUp() {
```
victornoel: use a simple `@Before` and a non-static `InMemoryProfileService` so that you are 100% sure there is no interferences between tests :)

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
```diff
@@ -0,0 +1,90 @@
...
+ * @author Elie Roux
+ * @since 2.0.0
+ */
+
+public class InMemoryProfileService<U extends CommonProfile> extends AbstractProfileService<U>  {
+
+	public Map<String,Map<String,Object>> profiles;
+	public Function<Object[], U> profileFactory;
+
+	public InMemoryProfileService(final Function<Object[], U> profileFactory) {
```
victornoel: I'm not sure but would it make sense to provide also a parameter-less constructor with a default profile factory for `CommonProfile`?

Actually, I would even:
 - remove any `profileFactory` from the constructor
 - call defaultProfileDefinition in the `internalInit()` method with a stupid `CommonProfileDefinition` initialised using an anonymous `CommonProfile` factory.
 - let the user call setProfileDefinition() if changing it is needed.

WDYT?

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
```diff
@@ -0,0 +1,90 @@
...
+                // https://stackoverflow.com/questions/24630963/java-8-nullpointerexception-in-collectors-tomap
+                .collect(HashMap::new, (m,v)->m.put(v.getKey(), v.getValue()), HashMap::putAll);
+	}
+
+	@Override
+	protected List<Map<String, Object>> read(final List<String> names, final String key, final String value) {
+		logger.debug("Reading key / value: {} / {}", key, value);
+		final List<Map<String, Object>> listAttributes = new ArrayList<>();
+		if (key.equals(getIdAttribute())) {
+			final Map<String,Object> profile = profiles.get(value);
```
victornoel: use stream api here too? and this time, there is no reason for `null` values to be present in `profiles` :)

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
eroux: I tried that, but I cannot make something like that work: Java is not happy when I give it a `CommonProfile`, it says `Type mismatch: cannot convert from CommonProfile to U`. I did not manage to find an actual solution, please suggest one.

eroux:
I've addressed your comments except the one with the signature for which I have no solution

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
victornoel: Ah yes, you are right, of course.
The only possible way to do that would be to introduce a static method that calls the constructor to do it. It's too much hassle I guess, let's forget about it :)

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
```diff
@@ -70,18 +71,18 @@ protected void deleteById(final String id) {
 	@Override
 	protected List<Map<String, Object>> read(final List<String> names, final String key, final String value) {
 		logger.debug("Reading key / value: {} / {}", key, value);
-		final List<Map<String, Object>> listAttributes = new ArrayList<>();
+		final List<Map<String, Object>> listAttributes;
 		if (key.equals(getIdAttribute())) {
+			listAttributes = new ArrayList<>();
 			final Map<String,Object> profile = profiles.get(value);
```
victornoel: sorry to insist, but why not use the stream api on this one too? :)

eroux: Well, first, I really need to ask: what is your motivation to use the Stream API in general (performance? readability?)? Also, does it have to be used everywhere? I can't help having the feeling that it's just a new toy... It can be useful in some places, but if it was the holy grail it would have been there for a long time...

Also, the reason why I think Stream is useless here, is that what I'm doing here is a `.get(myId)`, I'm not walking through the list. I can walk through the list and make a filter to check if the key is `myId`, but that would mean replacing the very efficient `.get` by something completely inefficient. And it would be, in my opinion, less readable. So, what I'm perceiving here is that it would harm performance and readability, but would use the new shiny toy... and I'm not willing to do that.

Please correct me and tell me why the stream api should be used here.

victornoel:
I noticed you use tabs for indentation, please make sure you conform to the editorconfig configuration of the project (stated in the `.editorconfig` at the root of the project). The easiest is to use an editorconfig plugin for your editor :)

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
victornoel: It's just a personal preference, mostly for readability yes, but it's ok if you prefer not to use it.

Regarding this particular use case, you are completely right, I read too fast and misunderstood what was written :)

eroux:
I realize I forgot to reinstall the editorconfig plugin, thanks for noticing, I fixed the indentation

victornoel:
Thanks for your time :) @leleuj maybe a last review to be sure things are well integrated with how pac4j work? For me it's good now :)

eroux:
Thank you very much! If you have comments on #945, don't hesitate! The couchdb mocking doesn't allow to test it though, so I don't think I can add some tests...

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * In-memory profile service.
+ *
+ * @author Elie Roux
+ * @since 2.0.0
```
leleuj: It will be 2.1.0.

pac4j-core/src/test/java/org/pac4j/core/profile/service/InMemoryProfileServiceTests.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.*;
+
+/**
+ * Tests the {@link InMemoryProfileService}.
+ *
+ * @author Elie Roux
+ * @since 2.0.0
```
leleuj: 2.1.0 as well.

leleuj:
A tiny comment, but that looks good to me!

eroux:
Done, thanks!

PR #951, opened 2017-06-26
#964 Support secret as byte[] via constructor

gbatalski:
Added possibility to set the secret via byte[]

### DIFF
--- a/pac4j-core/src/test/java/org/pac4j/core/util/TestsConstants.java
+++ b/pac4j-core/src/test/java/org/pac4j/core/util/TestsConstants.java
@@ -20,6 +20,8 @@ public interface TestsConstants {
     String FAKE_VALUE = "fakeValue";
     String SECRET = "secret";
     String MAC_SECRET = "12345678901234567890123456789012";
+    String BASE64_256_BIT_ENC_SECRET = "MvwjfoJ4LocaE_7n2Y-M7-X7Cq75dbzsuHWNB8XOQvo";
+    String BASE64_512_BIT_SIG_SECRET = "4pqomdEAl6l4aLJzR52fhHY01_sG1V60y7Vc3gAUTrvrkc4moALkPAe8rwe4-EJ2sVnZx8INO4oONJhH7BERGQ";
     String TOKEN = "token";
     String VERIFIER = "verifier";
     String NAME = "name";
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/config/encryption/SecretEncryptionConfiguration.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/config/encryption/SecretEncryptionConfiguration.java
@@ -5,6 +5,7 @@ import com.nimbusds.jose.crypto.AESDecrypter;
 import com.nimbusds.jose.crypto.AESEncrypter;
 import com.nimbusds.jose.crypto.DirectDecrypter;
 import com.nimbusds.jose.crypto.DirectEncrypter;
+import com.nimbusds.jose.util.Base64;
 import org.pac4j.core.exception.TechnicalException;
 import org.pac4j.core.util.CommonHelper;
 
@@ -99,6 +100,22 @@ public class SecretEncryptionConfiguration extends AbstractEncryptionConfigurati
 
     public void setSecret(final String secret) {
         this.secret = secret.getBytes(UTF_8);
+    }
+
+    public byte[] getSecretBytes() {
+        return  Arrays.copyOf(secret,secret.length);
+    }
+
+    public void setSecretBytes(final byte[] secretBytes) {
+        this.secret = Arrays.copyOf(secretBytes,secretBytes.length);
+    }
+
+    public String getSecretBase64() {
+        return Base64.encode(secret).toString();
+    }
+
+    public void setSecretBase64(final String secret) {
+        this.secret = new Base64(secret).decode();
     }
 
     @Override
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/config/encryption/SecretEncryptionConfiguration.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/config/encryption/SecretEncryptionConfiguration.java
@@ -8,7 +8,7 @@ import com.nimbusds.jose.crypto.DirectEncrypter;
 import org.pac4j.core.exception.TechnicalException;
 import org.pac4j.core.util.CommonHelper;
 
-import java.io.UnsupportedEncodingException;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Secret encryption configuration.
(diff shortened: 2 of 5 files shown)


gbatalski:
The PR is ready but there are some problems i see:

1. Mutalbility of instances as i mentioned before and necessity to preserve backward compatibility for getter and setter. But the improvement in compatible now.
1. set/getSecret(byte[] secret)  may be implemented, but again, i'm not so happy with a mutability 
1. my basic problem caused this issue was the missing possibility to accept the Base64 (Url) encoded keys for example generated via recommended cas key generator.

I didn't find an ultimate method to recognize the base64 string as such. All of methods i've considered were false positiv. The only possibility to recognize could be 
```java

Base64.Decoder decoder = Base64.getUrlDecoder();
try {
    decoder.decode(secretString);
} catch(IllegalArgumentException iae) {
    // Not a base64 string, use as plain one
}
```

In this case the caller should use  valid Base64 url encoded and fully padded base64 string to be recognized correctly. What do you mean?

victornoel:
@gbatalski maybe you should introduce multiple setters for each type of secret but store only the last set?

`setBytes`, `setBase64`, etc.

I don't think mutability is a problem, the setters are here only for the object to be javabean-like but nobody is expected (by implicit contract) to change the secret while things are initialised.

gbatalski:
**Base64.getUrlDecoder()** doesn't work, also false positive: org.pac4j.core.util.TestsConstants#MAC_SECRET is recognized as Base64 :-(

@victornoel I'm not very happy with the already existing getter/setter but should preserve it for backward compat. If you see any necessity, i could introduce more getter/setter. Then please propose a desired naming convention.

My problem is the implementation of org.apereo.cas.token.authentication.TokenAuthenticationHandler line 84 ```a.setSignatureConfiguration(new SecretSignatureConfiguration(signingSecret, signingAlg));``` and line 103 ```a.setEncryptionConfiguration(new SecretEncryptionConfiguration(encryptionSecret, encAlg, encMethod));``` 

@mmoayyed @leleuj may we change the symantic of jwtSigningSecret  and jwtEncryptionSecret  to be always base64 url encoded strings just like they are produced by key generator?

victornoel:
@gbatalski what I meant is that IF (maybe I misunderstood your need :) you need to distinguish setting a secret in string and a secret in base64, it could make sense to have different setters for each need instead of hacking something with a decoder :)

What is the problem with the setters (apart from the fact that it lets the user believe that these fields could be changed at any time, which is obviously false: changing them after initialisation has no effect I think)?

For the rest, I don't think I will be of much help :)

mmoayyed:
> may we change the symantic of jwtSigningSecret and jwtEncryptionSecret to be always base64 url encoded strings just like they are produced by key generator?

We may not, but I do very much like @victornoel's suggestion, that you should probably expose explicit setters for the kind of op you'd expect the class to accept. You want the intention to be very clear from the method, rather than "let me read the docs to figure out what's going on" :) 

So for the time being, a setBase64() sort of thing is perfectly fine. Progress over perfection.

gbatalski:
The problem is more focused on jwtSigningSecret and jwtEncryptionSecret.  I assume, this kind of JWT authentication isn't used too often, because my first trial with a generated key failed. The decoded key was also not correct because it contained non-printable chars not allowed within json config.  Could someone advice, what is the "correct" possibility in the meaning of CAS to teach TokenAuthenticationHandler to handle automatically the base64 keys? I hope, it is possible to solve the problem without patching the TokenAuthenticationHandler. But if not, i need a hint for patching because of backward compatibility in **cas-server-support-token-authentication**. 

My proposed functionality here would be an improvement enough,  although not really necessary: it helps only for direct usage of pac4j api with Base64 keys to avoid/reduce the ugly ```new String(Base64.decode(secret.getBytes(UTF_8)), UTF_8)``` ). The setter wouldn't resolve the core problem, even i could implement them within a few days. 

Sorry for too much noise.

mmoayyed:
No such thing as noise :) Thank you for following up.

Before we dig deeper, can you show an example of what you did on the CAS side that broke this configuration in the way it is?

gbatalski:
Hi, 

i've created the encryption key
```json
 java -jar jwk-gen.jar -a A128CBC_HS256 -t oct  -u enc -I -s 256
Full key:
{
  "kty": "oct",
  "use": "enc",
  "k": "MvwjfoJ4LocaE_7n2Y-M7-X7Cq75dbzsuHWNB8XOQvo",
  "alg": "A128CBC_HS256"
}
```
Then configured the service

```json
 "jwtEncryptionSecret": {
            "@class": "org.apereo.cas.services.DefaultRegisteredServiceProperty",
            "values": [
                "java.util.HashSet",
                [
                    "MvwjfoJ4LocaE_7n2Y-M7-X7Cq75dbzsuHWNB8XOQvo"
                ]
            ]
        },
        "jwtEncryptionSecretAlg": {
            "@class": "org.apereo.cas.services.DefaultRegisteredServiceProperty",
            "values": [
                "java.util.HashSet",
                [
                    "dir"
                ]
            ]
        },
        "jwtEncryptionSecretMethod": {
            "@class": "org.apereo.cas.services.DefaultRegisteredServiceProperty",
            "values": [
                "java.util.HashSet",
                [
                    "A128CBC-HS256"
                ]
            ]
        },
```
if i try to authenticate me with existing jwt token, below exception is thrown.

```bash
2017-06-27 13:22:41,608 ERROR [org.apereo.cas.integration.pac4j.authentication.handler.support.AbstractWrapperAuthenticationHandler] - <Failed to validate credentials>                                                  cas_1             | org.pac4j.core.exception.TechnicalException: com.nimbusds.jose.KeyLengthException: The Content Encryption Key length must be 128 bits (16 bytes), 192 bits (24 bytes), 256 bits (32 bytes), 384 bits (48 bytes) or 512 bites (64 bytes)                                                                                                                                                                                                                               cas_1             |     at org.pac4j.jwt.config.encryption.SecretEncryptionConfiguration.buildDecrypter(SecretEncryptionConfiguration.java:82) ~[pac4j-jwt-2.0.0.jar!/:?]       
```

 in debugger i see the direct usage of encoded key.
![grafik](https://user-images.githubusercontent.com/765808/27589893-3ba5aeac-5b4d-11e7-9e73-c642b18e1b98.png)

Same algorithm with ASCII key eg. 01234567890123456789012345678901 works well. I think, the necessity of byte[] constructor is given by problematic conversion from Base64 ->  UTF8 String -> byte[]. I think, the jwtEncryptionSecret and jwtSigningSecret should be also/only possible as Base64 encoded just like created via jwk-gen.

gbatalski:
Do we have something like Base64UrlEncodedRegisteredServiceProperty?

leleuj:
For the record, we have setters and getters in most pac4j components to be able to configure them in the Shiro framework in the shiro.ini file.
Various setter names seem good to me to have various configuration options.

leleuj:
This PR looks good to me. Specific setters are still expected before the merge I think...

mmoayyed:
Question: is there any way to "detect" whether a given String is base-64ed so that you can decode it automatically before getting its bytes, etc? That should remove the need for extra setters?

mmoayyed:
https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html#isBase64(java.lang.String) perhaps?

ghenadiibatalski:
I'm afraid no, **isBase64** only checks the alphabet. So 01234567890123456789012345678901 were also recognized as a correct base64 string in the meaning of this method. I've tried also to decode the string directly in a constructor and to catch illegal argument, but MAC_SECRET was also converted successfully :-((

mmoayyed:
Bummer. OK. So we have the extra setters fine, and that is great for the first leg of the work. Then I imagine the 2nd leg would to work this behavior into CAS so it can support "native" as well as "base64ed" strings somehow?

gbatalski:
optional prefix ```{base64}``` or the like?

gbatalski:
sorry, multiple browser with firm and private logins...

mmoayyed:
> optional prefix {base64} or the like?

Yes something like that, or perhaps another property that says "keysAreBase64" equals true and then we'd know what setter to call.

leleuj:
Setters are in place now.

@mmoayyed are you in line to merge this PR?

mmoayyed:
Yes. Thanks!

leleuj:
Merging

ghenadiibatalski:
@mmoayyed  I'll take a look at the "keysAreBase64" next week.

mmoayyed:
Good for you. Thank you.

PR #956, opened 2017-07-07
Azure Ad OIDC delegation client remove client_name from redirect URL

gbatalski:
This is the proposal for solving the problem observed with at least Office 365 Azure AD Open ID Connect delagation client. The problem is described  in https://github.com/apereo/cas/issues/2711.

As initially discussed with  @mmoayyed , the `client_name` could be replaced by `/delegatedAuthn/<clientType>/<clientName>` context path. 

@leleuj please take a look at the problem. The proposal is just a first trial. I need some assistance at combining the redirect url (1. resolution of cas server context path (now */cas* hardcoded) 2. possibly, replacement of client type (here */oidc*) by constant/enum to make the solution more generic?)

The counterpart of this is proposed to cas for extending the cas-server-support-pac4j-clients accordingly .

Any review/feedback is highly appreciated!

Regards, Gena

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/client/AzureAdClient.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/client/AzureAdClient.java
@@ -17,14 +17,14 @@ import org.pac4j.oidc.profile.azuread.AzureAdProfileCreator;
  * Azure AD v2.0. Replace {@code tenantid} with the ID of the tenant to authenticate against. To
  * find this ID, fill in your tenant's domain name. Your tenant ID is the UUID in
  * {@code authorization_endpoint}.
- * 
+ *
  * For authentication against an unknown (or dynamic tenant), use {@code common} as ID.
  * Authentication against the common endpoint results in a ID token with a {@code issuer} different
  * from the {@code issuer} mentioned in the discovery data. This class uses to special validator
  * to correctly validate the issuer returned by Azure AD.
  *
  * More information at: https://msdn.microsoft.com/en-us/library/azure/dn645541.aspx
- * 
+ *
  * @author Emond Papegaaij
  * @since 1.8.3
  */
@@ -34,6 +34,9 @@ public class AzureAdClient extends OidcClient<AzureAdProfile> {
 
     public AzureAdClient(final OidcConfiguration configuration) {
         super(configuration);
+        // https://github.com/apereo/cas/issues/2711
+        // don't include clientName because its ignored and causes validation failure
+        setIncludeClientNameInCallbackUrl(false);
     }
 
     @Override
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/client/AzureAdClient.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/client/AzureAdClient.java
@@ -1,6 +1,7 @@
 package org.pac4j.oidc.client;
 
 import org.pac4j.core.context.WebContext;
+import org.pac4j.core.http.RelativeUrlResolver;
 import org.pac4j.core.util.CommonHelper;
 import org.pac4j.oidc.client.azuread.AzureAdResourceRetriever;
 import org.pac4j.oidc.config.OidcConfiguration;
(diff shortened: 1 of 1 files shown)


gbatalski:
I've closed pull request https://github.com/apereo/cas/issues/2772 to apereo/cas until Jérôme reviewed and commented this proposal. But the discussion there still be interesting. The counterpart can be seen here https://github.com/gbatalski/cas/pull/1 . Misagh proposes to integrate also the cas part into pac4j. But i don't see any Controller within the pac4j implementation..

mmoayyed:
This is probably the spot you need to review, though @leleuj must confirm: https://github.com/pac4j/spring-webmvc-pac4j

gbatalski:
@mmoayyed 
I've placed the proposed controller next to `DelegatedClientAuthenticationAction`, because it "decodes" the callback url and forwards requests to DelegatedClientAuthenticationAction webflow. I think, it belongs somehow  to pac4j client support within the cas (as usage?), even if the strategy for that should be generally approved by @leleuj and you. 

The fact is, this seems to be the only way (not the implementation) to delegate to OIDC for Azure AD and as such,  i would love to have it within the standard and not as my own patch or addon.  I also suppose, other, at least oidc, clients could remove the support for additional url parameter for redirect urls, as its seems to be an unsafe practice. I also had problems to verify the token against Azure, because the verification url registered there is not allowed to contain any url parameter (also .* or * wildcards didn't help).

mmoayyed:
> I think, it belongs somehow to pac4j client support within the cas (as usage?), even if the strategy for that should be generally approved by @leleuj and you.

Not just yet. The reason I hesitate is because this is a rather specialized strategy in dealing with one particular client. I generally dislike devising specialized strategies :) So while this is a working solution, the overall theme might be that perhaps pac4j needs to review how callback urls are constructed and change its way so there is no need for special decoding. 

From the usage perspective, all clients within CAS must behave and act exactly the same configuration-wise. All special logic must sit outside CAS and that's where pac4j comes in. That's why Pac4j is used; Otherwise there would be no point. CAS is not and should not be in the business of understanding how to deal with Azure AD as a client. That is pac4j's business. 

>The fact is, this seems to be the only way (not the implementation) to delegate to OIDC for Azure AD and as such, i would love to have it within the standard and not as my own patch or addon. 

I could not agree more. Maybe it turns out that we gracefully accept the original PR into CAS for the time being. You may also have to live with the patch for a while. You can also talk to Azure AD and yell at them. Before any of that ever happens, I'd like to focus on the "right" thing a bit more and not do a quick solution. 

> I also suppose, other, at least oidc, clients could remove the support for additional url parameter for redirect urls, as it seems to be an unsafe practice.

Exactly. That's what I had in mind that the strategy needs to be consistent. If Pac4j ends up changing the way parameters are built our roadmap would be to:

1. Change pac4j's callback url semantics
2. Test, new release.
3. Have CAS upgrade and it should all work

While you in the meanwhile live with the patch.

leleuj:
I'm in the middle of a relocation, so I'll take a closer look at it next week.

I understand your concern and indeed, there is a problem with Azure and room for a more generic solution.

Your solution is too specific (by embedding the CAS URL: `/cas/delegatedAuthn/oidc/`).

leleuj:
Things are not easy here, we have two main components involved:

- the `UrlResolver` at the `IndirectClient` level whose role is to switch the host/port/domain depending on the use cases. This behavior is specific to a client.

- the `Clients` which computes the callback URL for all clients and retrieves the appropriate indirect client based on a given parameter. This behavior is specific to all clients defined for this component.

Maybe we should update the `UrlResolver` to make it compute the **full** callback URL **and** be able to say if it is the right client on the callback URL (some `supports` method).

This would be a breaking change. I need to think a little more about it.

gbatalski:
**/cas/delegatedAuthn/oidc/** is only proposal. The proposal for cas support modification/extension may be found in https://github.com/gbatalski/cas/pull/1 

It must not be a breaking change, as we could support both treatment depending on client configuration (as backward compatibility) in saying 5.3. But you are right, the cleaner solution would be to change the handling globally as a breaking change in 6.0.

leleuj:
I think you should implement a quick workaround inside CAS with a `CasAzureAdClient`(your proposal) and we'll be able to remove it when CAS uses pac4j 3.

Closed in favor of: https://github.com/pac4j/pac4j/issues/967

PR #958, opened 2017-07-11
Fixes InMemoryProfileService.update

eroux:
The `InMemoryProfileService.update()` function used to erroneously assume that all attributes were present during an update so it just replaced the previous profile entry with the asked one, overwriting attritutes that were not meant to be overwritten. This PR fixes this behavior and adds a test. I'll send a separate PR for CouchProfileService which is doing the same mistake.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
@@ -21,71 +21,71 @@ import java.util.stream.Collectors;
 
 public class InMemoryProfileService<U extends CommonProfile> extends AbstractProfileService<U>  {
 
-	public Map<String,Map<String,Object>> profiles;
-	public Function<Object[], U> profileFactory;
+    public Map<String,Map<String,Object>> profiles;
+    public Function<Object[], U> profileFactory;
 
-	public InMemoryProfileService(final Function<Object[], U> profileFactory) {
-		this(new HashMap<>(), profileFactory);
-	}
+    public InMemoryProfileService(final Function<Object[], U> profileFactory) {
+        this(new HashMap<>(), profileFactory);
+    }
 
-	public InMemoryProfileService(final Map<String,Map<String,Object>> profiles, final Function<Object[], U> profileFactory) {
-		this.profiles = profiles;
-		this.profileFactory = profileFactory;
-	}
+    public InMemoryProfileService(final Map<String,Map<String,Object>> profiles, final Function<Object[], U> profileFactory) {
+        this.profiles = profiles;
+        this.profileFactory = profileFactory;
+    }
 
-	@Override
-	protected void internalInit(final WebContext context) {
-		CommonHelper.assertNotNull("passwordEncoder", getPasswordEncoder());
-		defaultProfileDefinition(new CommonProfileDefinition<U>(profileFactory));
-		super.internalInit(context);
-	}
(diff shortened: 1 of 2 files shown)


eroux:
@leleuj @victornoel any comment on this one and #959 ?

leleuj:
I'm not convinced there is a mistake here: it's more a matter of what behavior we want for the update.

Previously, update meant erase if the profile already existed and create if it didn't.
What you propose is to merge the new profile and the existing one if it already exists and not to create it if it doesn't exist.

It's a breaking change also which fits badly with a minor version like v2.1.

What use case pushed you into that direction?

eroux:
About the change: if you look at the [DbProfileService](https://github.com/pac4j/pac4j/blob/master/pac4j-sql/src/main/java/org/pac4j/sql/profile/service/DbProfileService.java#L103) it clearly just updates the asked fields without removing the others when in legacy mode.

About the use case: I just want to update a profile without updating the password. Or is there a way to do that that I missed?

eroux:
BTW, the use case is reflected in the additions in the tests, something like

```java
// update with no password update
 inMemoryProfileService.update(profile, null);
 // check credentials
inMemoryProfileService.validate(credentials2, null);
CommonProfile profile4 = credentials.getUserProfile();
assertNotNull(profile4);
```

was failing before (with the correct credentials), but doesn't fail now.

leleuj:
I guess you're right about the password. It would be erased if not provided during the update.

But if the profile does not already exist, the update won't be done, right?

eroux:
No it would not, should it? Maybe adding some documentation on the function would help... I remember trying to find a path in the code where a profile would be updated without being created first, but I coudln't find any, so I assumed that update was always called on a profile already existing in the database... should it raise an exception if it isn't? or fail silently? or create the profile?

leleuj:
My feeling is that we should create the profile at the update if it doesn't already exist to be backward compatible. But please add a TODO to keep track that we need to revisit that for the version 3.0.

eroux:
Ok, I'll change that, but if the update behavior should be revisited, I guess it should be revisited for all plugins, so maybe opening a separate issue would be better than writing a TODO in this file?

leleuj:
Yes, please open an issue. Forget about the TODO idea...

eroux:
I've updated the PR with the requested change, is there anything else?

pac4j-core/src/main/java/org/pac4j/core/profile/service/InMemoryProfileService.java
```diff
@@ -51,7 +51,12 @@ protected void insert(final Map<String, Object> attributes) {
     protected void update(final Map<String, Object> attributes) {
         final String id = (String) attributes.get(getIdAttribute());
         logger.debug("Updating id: {} with attributes: {}", id, attributes);
-        profiles.put(id, attributes);
+        final Map<String,Object> profile = profiles.get(id);
+        if (profile != null) {
+            profile.putAll(attributes);
+        } else {
+            insert(attributes);
```
leleuj: Cannot we just do: `profiles.put(id, attributes);`?

eroux: if you prefer

eroux:
done

leleuj:
Thanks

leleuj:
Can you do the same change for the other PR for CouchDB?

eroux:
sure, I'll do that later today

PR #959, opened 2017-07-11
Fix CouchProfileService.update()

eroux:
This PR is the equivalent of https://github.com/pac4j/pac4j/pull/958 for CouchProfileService. It also reformats the files correctly so the global diff is difficult to read, the real changes are in a separate commit though.


eroux:
@leleuj done

leleuj:
Thanks

PR #960, opened 2017-07-11
SAML2Client: Allow passive authn option

mmoayyed:
Title says all.


mmoayyed:
Any objections before merging?

leleuj:
I guess it's a non-breaking change with `passive` being `false` by default. Updating the documentation as well would have been a good thing: https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/saml.md

PR #961, opened 2017-07-13
Add Authn-related attribute map to UserProfile, and specific Authn attributes for SAML.

### LINKED ISSUE #947: Expose (Issuer) EntityID and AuthnContext from recieved SAML 2 assertion as attributes
Add mechanism to return the issuer EntityID and AuthnContext via the UserProfile so they can be used by downstream authentication and/or authorization processes.

Doing this unconditionally is a fairly straightforward change (I think) to saml/sso/impl/SAML2DeafultResponseValidator; making it optional might take a bit of thought.

rrenomeron:
Fixes #947 and #948.

### DIFF
--- a/documentation/docs/clients/saml.md
+++ b/documentation/docs/clients/saml.md
@@ -134,6 +134,15 @@ You can generate the SP metadata in two ways:
 - either programmatically using the `SAML2Client`: `String spMetadata = client.getServiceProviderMetadataResolver().getMetadata();`
 - or by defining the appropriate configuration: `cfg.setServiceProviderMetadata(new FileSystemResource("/tmp/sp-metadata.xml"));`
 
+## 4) Authentication Attributes
+
+The following authentication attributes are populated by this client:
+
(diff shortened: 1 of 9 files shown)


leleuj:
Excellent!

I would appreciate if you could update the documentation as well (it will be on this PR). I mean:

- here: https://github.com/pac4j/pac4j/blob/master/documentation/docs/user-profile.md to explain the user profile has authentication attributes
- and here: https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/saml.md to explain new attributes are available at the `SAML2Profile` level
- and there: https://github.com/pac4j/pac4j/blob/master/documentation/docs/release-notes.md to track that change for v2.1.

See the process to browse and update the documentation: http://www.pac4j.org/docs/contribute.html#2-documentation

leleuj:
Thanks for the documentation update.

This will go into the version 2.2 as the version 2.1 has been released.

PR #962, opened 2017-07-15
Use a more generic param type for keycloak config

mmoayyed:


leleuj:
The build is broken due to a forgotten import.

What's the use case behind this change?

mmoayyed:
Use case: ability to use a generic OIDC configuration up front and then cast down to set more specific settings, rather than creating a whole new config object and copy everything into it. See https://github.com/apereo/cas/blob/master/support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/support/authentication/Pac4jAuthenticationEventExecutionPlanConfiguration.java#L300

Issue #955, opened 2017-07-05
About coding style


tianyin:
I read `CONTRIBUTING.md` and checked out `.editorconfig`.
(it seems that we use EditorConfig for code-style control)
However, there're still a number of style violations.

I run `CheckStyle` with the `LineLength < 140` rule (specified in `.editorconfig`) and there are `57` lines longer than `140` chars.

Similarly, the indent is not always consistent. 
I run `CheckStyle` with the "indent as 4 space" rule (specified in `.editorconfig`) and there're 18 violations.
For example, in `pac4j-core/src/main/java/org/pac4j/core/client/Clients.java`,
line 39 uses a `tab` as indentation instead of spaces; same as Line 138--140, Line 173--177. 
(there're also violations of other rules such as `end_of_line = lf`, etc)

As a more general question, shall we use `CheckStyle` or something alike to enforce the code style as a part of integration testing?
Relying on Editor/IDE settings seems not very effective :(

Thanks!

victornoel:
That can't be a bad thing to enforce these things :)

leleuj:
I generally have the feeling that using Checkstyle is mostly a waste of time, but I'll follow the agreement on this.

victornoel:
To me, it's useful only when there is multiple people working on the same project and you don't want to bother with verifying and each and every PR for style.

Checkstyle itself, I don't know well… would be good enough to run an editorconfig utility and fail the build if it's not valid w.r.t. the editorconfig file? Something like this: https://github.com/slang800/editorconfig-tools (not sure there is an official tool… I thought there was but can't find it anymore…)

victornoel:
Actually maybe https://github.com/jedmao/eclint is better…

tianyin:
`eclint` is cool as we've already adopted `.editorconfig`.

The nice thing about CheckStyle is that it is the de facto style checker for Maven projects,
so it can be directly integrated into the build process.
For `eclint`, we perhaps need to add an ad hoc scripts if we want to do this.

@leleuj what's the problem of CheckStyle? (just want to understand a little bit more)

leleuj:
"ugly side of Checkstyle" is a bit too much, but indeed I don't think Checkstyle is so necessary.

You can properly format your source code and break a few style rules of a project, it's not a big deal from my POV...

That being said, I'll accept a PR using Checkstyle to match our `.editorconfig`.

leleuj:
[...]

PR #966, opened 2017-07-19
Integrate Checkstyle for style checks (Issue #955)

tianyin:
This PR integrates CheckStyle using `maven-checkstyle-plugin` (added in `pom.xml`).

I create two CheckStyle configurations:
- `checkstyle.xml` is a strict translation of `.editorconfig` in the code base (https://github.com/pac4j/pac4j/blob/master/.editorconfig). I suppose it's the style that are supposed to be followed. Unfortunately, as shown in #955 , the current code base has a large number of violations against these style specifications.
- `checkstyle-zero-vio.xml` is the actual style inferred from the existing code base, and thus is coherent with the code. If you run Checkstyle against this specification, there is 0 violations. 

The command to run Checkstyle is:
```
$ mvn checkstyle:check
```

As I use `checkstyle-zero-vio.xml` in `pom.xml`, the PR would __not__ affect the build process. However, changing it to `checkstyle.xml` would results in many violations (fixing them needs to revise the code).


victornoel:
I think we should directly use the correct checkstyle configuration and fix the problems in one go :)

But are the violations related to indentation and stuffs like that or also to other stylistic issues (because the zero-vio file seems to contain much more than just things about indentation style…)

tianyin:
@victornoel then we can apply `checkstyle.xml`
The violations include indentation, line length, and trailing white space (basically those 5 rules specified in `.editorconfig`; all of these rules are violated)

The zero-violation file is generated via a tool that infers the style of an existing code base
(it's strictly consistent with the current code base)
There're more rules than what is specified in `.editorconfig` (which are somehow followed by the current code)

victornoel:
ok, so seems good to me then, could you apply the changes to the code and we will see how it looks like? :)

leleuj:
Yes, I'm in line, we should follow our own `.editorconfig` and fix all violations.

tianyin:
done.  you can run 
```
$ mvn checkstyle:check
```
to examine and fix the violations.

victornoel:
@tianyin but could you actually apply the changes and put the in a commit? As well as enabling checkstyle in the travis build?

leleuj:
@tianyin Can we complete this and integrate the checkstyle verification in the build? As well as the appropriate changes? Thx

leleuj:
Thanks

PR #974, opened 2017-08-17
Allow different timeouts per configuration instance, using HttpUtils as default

papegaaij:
This PR restores the read and connect timeout settings in OAuthConfiguration and OidcConfiguration. HttpUtils now serves as default.

### DIFF
--- a/pac4j-oauth/src/main/java/org/pac4j/oauth/config/OAuthConfiguration.java
+++ b/pac4j-oauth/src/main/java/org/pac4j/oauth/config/OAuthConfiguration.java
@@ -6,9 +6,9 @@ import com.github.scribejava.core.model.SignatureType;
 import com.github.scribejava.core.model.Token;
 import com.github.scribejava.core.oauth.OAuthService;
 import org.pac4j.core.client.IndirectClient;
+import org.pac4j.core.context.HttpConstants;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.util.CommonHelper;
-import org.pac4j.core.util.HttpUtils;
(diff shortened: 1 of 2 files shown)


pac4j-core/src/main/java/org/pac4j/core/util/HttpUtils.java
```diff
@@ -18,9 +18,9 @@
  */
 public final class HttpUtils {
 
-    private static int connectTimeout = HttpConstants.DEFAULT_CONNECT_TIMEOUT;
+    private static int defaultConnectTimeout = HttpConstants.DEFAULT_CONNECT_TIMEOUT;
```
leleuj: I'm not sure to understand why we now have default timeouts instead of timeouts...

leleuj:
The `HttpUtils` class should not be changed. Only timeouts should be added back for the OAuth and OpenID Connect classes.

papegaaij:
I decided to change it this way to prevent a change in behavior when people already changed their code to use HttpUtils.

papegaaij:
I can change the PR to leave HttpUtils untouched and initialize the timeouts to the constants, but that would break existing applications using HttpUtils to set the timeout.

leleuj:
In fact, this change was already a breaking one (as all timeouts were set to one value), which is something we do not allow for minor versions.
So we definitely need to revert this change as it was in 2.0.

papegaaij:
I've rewritten the original commit (so the change still is only 1 commit) to only change the configuration classes.

leleuj:
Thanks. We'll see if other people report an issue with a unique timeout when testing the version 2.1.

PR #975, opened 2017-08-24
Integration of logout mechanism for SPA applications and oidc

andreaTP:
This is a PR to enable SPA applications to handle the logout mechanism in Oidc context.
I followed what we already did for login functionality, here I used code 403 ....

Please note that I'm introducing an API incompatibility changing the signature of a public method in Client that now can throw `HttpAction`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
 	<artifactId>pac4j</artifactId>
 	<packaging>pom</packaging>
 	<name>pac4j</name>
-	<version>2.2.0-SNAPSHOT</version>
+	<version>3.0.0-SNAPSHOT</version>
 	<description>Profile &amp; Authentication Client for Java</description>
 	<url>https://github.com/pac4j/pac4j</url>
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ It provides a comprehensive set of [**concepts and components**](http://www.pac4
 
 ## Versions
 
-The version **2.2.0-SNAPSHOT** is under development. Maven artifacts are built via Travis: [![Build Status](https://travis-ci.org/pac4j/pac4j.png?branch=master)](https://travis-ci.org/pac4j/pac4j) and available in the [Sonatype snapshots repository](https://oss.sonatype.org/content/repositories/snapshots/org/pac4j).
+The version **3.0.0-SNAPSHOT** is under development. Maven artifacts are built via Travis: [![Build Status](https://travis-ci.org/pac4j/pac4j.png?branch=master)](https://travis-ci.org/pac4j/pac4j) and available in the [Sonatype snapshots repository](https://oss.sonatype.org/content/repositories/snapshots/org/pac4j).
 
 The source code can be cloned and locally built via Maven:
 
--- a/pac4j-cas/pom.xml
+++ b/pac4j-cas/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.pac4j</groupId>
         <artifactId>pac4j</artifactId>
-        <version>2.2.0-SNAPSHOT</version>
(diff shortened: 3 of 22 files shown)


leleuj:
Thanks for this contribution. It completely makes sense.

Though, as it holds a breaking change, it cannot go into version 2.2, but only 3.0. The development of version 3.0 will start in September.

I will review the PR carefully.

pac4j-oidc/src/main/java/org/pac4j/oidc/logout/AjaxAwareOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+import org.pac4j.oidc.profile.OidcProfile;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * Logout action builder for OpenID Connect that takes care of ajax calls.
+ *
+ * @author Andrea Peruffo
+ * @since 2.2.0
```
leleuj: It will be 3.0.0.

pac4j-oidc/src/main/java/org/pac4j/oidc/logout/AjaxAwareOidcLogoutActionBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+                final JWT idToken = currentProfile.getIdToken();
+
+                LogoutRequest logoutRequest;
+                if (CommonHelper.isNotBlank(targetUrl)) {
+                    logoutRequest = new LogoutRequest(endSessionEndpoint, idToken, new URI(targetUrl), null);
+                } else {
+                    logoutRequest = new LogoutRequest(endSessionEndpoint, idToken);
+                }
+
+                if (ajaxRequestResolver.isAjax(context)) {
```
leleuj: Why not put this change in the already existing `OidcLogoutActionBuilder`?

pom.xml
```diff
@@ -444,6 +444,11 @@
 						</supportedProjectTypes>
 					</configuration>
 				</plugin>
+        <plugin>
```
leleuj: I'm not sure to see why we need this plugin.

pom.xml
```diff
@@ -504,4 +509,17 @@
 		</profile>
 	</profiles>
 
+  <distributionManagement>
```
leleuj: This should definitely not be here.

pac4j-oidc/src/main/java/org/pac4j/oidc/logout/AjaxAwareOidcLogoutActionBuilder.java
andreaTP: ok, going to merge into it

pom.xml
andreaTP: my bad, publishing locally ...

pom.xml
andreaTP: yes, sorry ...

andreaTP:
updated according to your comments!

leleuj:
OK. I'll merge that as soon as we start working on the version 3.0.0.

andreaTP:
Thanks! :-)

andreaTP:
Any update on this?

leleuj:
As I said, I'll merge it as soon as we start working on 3.0 (likely next week), then a few months later, I'll release the 3.0.0 version. Be prepared to be patient.

andreaTP:
argh, wow ... months ... is maybe possible to have published a `-PRE` release sooner?

leleuj:
We'll certainly have RCs before, but still, it won't be before several weeks.

andreaTP:
Thanks! :-)

PR #982, opened 2017-09-07
Add support for configuration of SAML2 ACS index

mmoayyed:
Going to let CIs do their thing so I can spot and fix errors, if any.


leleuj:
+1 Looks good to me, even if I'm not very familiar with all SAML subtilities.

Thanks for the documentation update as well.

mmoayyed:
Cool. Thanks for the review.

Short of it is that (I believe) an authn request does either send an ACS url for the followup post, or an index which the idp will use to look up the ACS url. Figured we might have support for the ACS index now that we also have support for attribute idx.

PR #983, opened 2017-09-12
Remove the setSessionStore method and other deprecateds

### LINKED ISSUE #873: Refactor WebContext.setSessionStore
Since it makes no sense for the session store associated with a WebContext instance to change during the lifetime of the WebContext, there should not be a setter for this. All it does is create API noise, and the only sensible implementation is an UnsupportedOperationException.

It's reasonable (and actually should be enforced) that the SessionStore should be set during the construction of the WebContext instance and not be changed during the lifetime of the instance. So we should deprecate this method in 2.0 and remove it in 3.0.

I'm happy to hear any use-cases where we think we should change the SessionStore during the lifetime of the WebContext and be persuaded that I'm wrong.

leleuj:
@millross Removing the `setSessionStore` method was finally easy. The `ProfileManager` now takes a `SessionStore` if necessary (or use the one of the `WebContext` otherwise) to perform its operations.

Closes #873


millross:
Thank you, that's nice :)

PR #990, opened 2017-09-15
Make HttpAction unchecked

### LINKED ISSUE #984: Make HttpAction unchecked
To make evolutions and asynchronism easier.

leleuj:
Closes #984 

@millross : for async also, this is when I realize how checked exceptions clutter the whole pac4j source code.

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -7,6 +7,7 @@ title: Release notes&#58;
 
 - Handle AJAX requests in the OpenID Connect logout
 - All session interactions are done via the `SessionStore` (retrieved from the `WebContext`)
+- All exceptions (especially `HttpAction`) are unchecked
 
 **v2.2.0**:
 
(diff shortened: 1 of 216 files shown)


millross:
I think one of the things about HttpAction is there's a tension between making it part of a method signature (checked) so that _someone_ will have to handle it, and the fact that we in many cases let it bubble out to be dealt with by an HttpActionAdapter.

The rationale (someone will have to handle it) is a good one. However, the main (arguably) only reason to use a checked exception is so that an immediate caller can handle it. Once you start propagating checked exceptions from inner code outwards, then they should probably be changed to unchecked. As I said, HttpAction is slightly borderline (because we want to force someone to handle it and we also use it as a return type). However, in terms of the way we actually use it (lots of propagation outwards) I think unchecked is correct. It became very painful in the Async api building very quickly so I locally changed it to unchecked for convenience, but the more I think about it, the more I think it should be unchecked because of the way we propagate.

leleuj:
I fully agree.

PR #1000, opened 2017-10-04
Introduce a default implementation for Principal

### LINKED ISSUE #996: Provide a default java.security.Principal implementation
Following a discussion in pac4j/j2e-pac4j#44, we decided with @leleuj to move jax-rs-pac4j's `java.security.Principal` implementation into pac4j-core since it was only specific to pac4j and could be useful for other  frameworks.

@leleuj question: where do I put it in the source tree? I was thinking in the `profile` package…

victornoel:
Closes #996 

@leleuj validate the name and package are chosen as desired :)

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/Pac4JPrincipal.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/Pac4JPrincipal.java
@@ -1,0 +1,59 @@
+package org.pac4j.core.profile;
+
+import java.security.Principal;
+import java.util.Objects;
+
+import org.pac4j.core.util.CommonHelper;
+
+/**
+ * 
+ * Default implementation for {@link Principal} based on a pac4j
+ * {@link UserProfile}.
+ * 
+ * It determines the name based on the profile's username, and fallbacks to id
+ * if it doesn't exist.
+ * 
+ * @author Victor Noel
+ * @since 3.0.0
+ */
+public class Pac4JPrincipal implements Principal {
+
+    private final String name;
+
+    public Pac4JPrincipal(CommonProfile profile) {
+        String username = profile.getUsername();
+        if (CommonHelper.isNotBlank(username)) {
+            this.name = username;
+        } else {
+            this.name = profile.getId();
+        }
+    }
+
+    @Override
+    public String getName() {
+        return this.name;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(getName());
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        final Pac4JPrincipal principal = (Pac4JPrincipal) o;
+        return CommonHelper.areEquals(this.getName(), principal.getName());
+    }
+
+    @Override
+    public String toString() {
+        return CommonHelper.toNiceString(this.getClass(), "profileId", this.name);
+    }
+}
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/CommonProfile.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/CommonProfile.java
@@ -4,6 +4,7 @@ import org.pac4j.core.context.Pac4jConstants;
 import org.pac4j.core.profile.definition.CommonProfileDefinition;
 
 import java.net.URI;
+import java.security.Principal;
 import java.util.Locale;
 
 /**
@@ -109,5 +110,9 @@ public class CommonProfile extends UserProfile {
      */
     public String getLocation() {
         return (String) getAttribute(CommonProfileDefinition.LOCATION);
+    }
+
+    public Principal asPrincipal() {
+        return new Pac4JPrincipal(this);
     }
 }


pac4j-core/src/main/java/org/pac4j/core/profile/Pac4JPrincipal.java
```diff
@@ -0,0 +1,61 @@
...
+    private final Set<String> roles;
+
+    public Pac4JPrincipal(LinkedHashMap<String, CommonProfile> profiles) {
+        Optional<CommonProfile> optProfile = ProfileHelper.flatIntoOneProfile(profiles);
+        if (!optProfile.isPresent()) {
+            throw new IllegalArgumentException();
+        }
+        this.profile = optProfile.get();
+        Set<String> rs = new HashSet<>();
+        profiles.values().stream().forEach(p -> rs.addAll(p.getRoles()));
```
leleuj: We gather the roles from all profiles, but we all return the identifier of only one profile.

pac4j-core/src/main/java/org/pac4j/core/profile/Pac4JPrincipal.java
```diff
@@ -0,0 +1,61 @@
...
+public class Pac4JPrincipal implements Principal {
+
+    private final CommonProfile profile;
+
+    private final Set<String> roles;
+
+    public Pac4JPrincipal(LinkedHashMap<String, CommonProfile> profiles) {
+        Optional<CommonProfile> optProfile = ProfileHelper.flatIntoOneProfile(profiles);
+        if (!optProfile.isPresent()) {
+            throw new IllegalArgumentException();
```
leleuj: Let's throw a `TechnicalException` with a clear message to stay in the "pac4j" field...

pac4j-core/src/main/java/org/pac4j/core/profile/Pac4JPrincipal.java
victornoel: Should we return instead only the roles of the one profile selected by `flatIntoOneProfile`?

leleuj: Finally, I think it's better to return all roles. Keep things that way.

victornoel: That raises the issue of why we don't do that when we return a profile using `ProfileHelper.flatIntoOneProfile()` in all pac4j... Maybe it depends where it is used.

leleuj: For me, it's not the same use case: using `flatIntoOneProfile`, we choose one profile out of the several ones we may have. It's a complete profile with roles and identifier. Here, we have no choice, we try to get as much as we can from the pac4j information: all roles, but only one identifier as we can only return one name.

I think we should not keep a profile variable in this principal but only a name property, built in the constructor from the profile identifier.
We could even imagine to have this name built from all profile identifiers.

victornoel: Yep, that makes sense for not keeping the profile.
I don't know about the name being a compound of the all the profile's name thought (because usually I think people would be expecting the name to be a username for example).

I think we should go back to the reason why we are building a principal. Right now it's not clear when this information is extracted and why it is needed, so it's not clear what should be kept or not. I will try to think a bit about it in the context of jax-rs-pac4j and get back on that :)

leleuj: A compound of all the profile's names was just an example, it clearly makes more sense to have a username.

I would keep all roles, the id of the `flatIntoOneProfile`-returned profile and keep the id property instead of the whole profile.

victornoel: Actually, I just realised that the `getRoles()` method is something I introduced but is not part of the `Principal` interface (I stored it there because it made sense in jax-rs-pac4j).

So in the end, the principal is just a name (and an `implies` relation with a `Subject`)… and thus, I wonder if we shouldn't do thing the other way around: a pac4j profile IS-A Principal and should simply implements the interface.

What do you think?

leleuj: Actually, it makes a lot more sense to have it that way.

victornoel: Hm, actually, that's not so obvious, because implementing Principal has strong implications in terms of semantic:
 - first there is the method names that are very generic and will clash with CommonProfile
 - also there is the hashCode/equals method that may not be exactly what we would expect for a UserProfile

So in the end, I think I will continue with the idea of having a separate class… I'm going to rework this branch and you will tell me what you think :)

victornoel: ok, it's done

pac4j-core/src/main/java/org/pac4j/core/profile/Pac4JPrincipal.java
```diff
@@ -0,0 +1,43 @@
+package org.pac4j.core.profile;
+
+import java.security.Principal;
+import java.util.Objects;
+
+import org.pac4j.core.util.CommonHelper;
+
+public class Pac4JPrincipal implements Principal {
```
leleuj: Please add the Javadoc, the `@author` and the `@since`...

pac4j-core/src/main/java/org/pac4j/core/profile/Pac4JPrincipal.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final String id;
+
+    public Pac4JPrincipal(UserProfile profile) {
```
leleuj: Just wondering whether it would be better to have the identifier directly... as indeed, the whole `UserProfile` is not necessary... but at the same time, this makes this `Pac4JPrincipal` component fully tied to pac4j...

If we keep the `UserProfile` in the constructor, we should add a `CommonHelper.assertNotNull` on it...

victornoel: My idea was that the principal was derived from profile, so we pass the profile to it, and it is the responsibility of this implementation to take from it what it needs.
Like this we are sure that everybody using this implementation will use the same information from the profile to build the principal.

Also I am wondering if it should use `id` or `username` from the profile…

OK for the rest :)

leleuj: About the id/username, the question is whether the `getName()` must return a unique identifier or something more meaningful.

victornoel: I was more thinking about `getUsername()` since it is what represents user identity… there is some chance for `getId()` to be a db id or something like this…

leleuj: OK. Let's do that. And if the username is blank, let's use the identifier...

victornoel: makes sense!

victornoel: Of course, things are never easy :)
The `UserProfile` does not have a `getUsername()` method!

But I don't see any reason to work at the `CommonProfile` level, the Principal should be available for any profile retrieved from a provider.

So I will use `getId()` and we will see later if some other need arises.

leleuj: We never use the `UserProfile` which is abstract. Why not move things at the `CommonProfile` level?

victornoel: ok, that is an acceptable alternative, but why does UserProfile exist then?
I will use CommonProfile :)

leleuj: Its methods could have been merged into `CommonProfile`, but it's a design split: `UserProfile` contains generic user profile methods while `CommonProfile` contains common attributes methods.
As we want to enforce people to use common profiles, `UserProfile` is abstract and never really used.

victornoel: ok, make sense :)

so I pushed the last changes, I think you can validate and finally merge if I didn't forget anything :)

PR #1003, opened 2017-10-06
Logout OK should be a 204

### LINKED ISSUE #701: Logout OK should be a 204

leleuj:
Closes #701 

cc @victornoel 

@mmoayyed : I have been able to use the 204 status code for the CAS back-channel logout call, but not in the proxy callback.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/exception/HttpAction.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/exception/HttpAction.java
@@ -49,17 +49,6 @@ public class HttpAction extends TechnicalException {
     }
 
     /**
-     * Build an HTTP Ok without any content.
-     *
-     * @param message message
-     * @param context context
-     * @return an HTTP ok
-     */
-    public static HttpAction ok(final String message, final WebContext context) {
-        return ok(message, context, "");
-    }
-
-    /**
      * Build an HTTP Ok.
      *
      * @param message message
@@ -67,10 +56,23 @@ public class HttpAction extends TechnicalException {
      * @param content content
      * @return an HTTP ok
      */
-    public static HttpAction ok(final String message, final WebContext context, String content) {
+    public static HttpAction ok(final String message, final WebContext context, final String content) {
         context.setResponseStatus(HttpConstants.OK);
         context.writeResponseContent(content);
         return new HttpAction(message, HttpConstants.OK);
+    }
+
+    /**
+     * Build an HTTP No content.
+     *
+     * @param message message
+     * @param context context
+     * @return an HTTP No content
+     */
+    public static HttpAction noContent(final String message, final WebContext context) {
+        context.setResponseStatus(HttpConstants.NO_CONTENT);
+        context.writeResponseContent("");
+        return new HttpAction(message, HttpConstants.NO_CONTENT);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


mmoayyed:
Thanks. I'll keep this mind when I switch to v3 for testing in about a month or so.

leleuj:
@mmoayyed I remember adding the 204 HTTP status code as acceptable codes for the logout, but I guess we should do the same for proxy checking...

pac4j-core/src/main/java/org/pac4j/core/exception/HttpAction.java
```diff
@@ -49,28 +49,30 @@ public static HttpAction redirect(final String message, final WebContext context
...
      * @param context context
-     * @param content content
-     * @return an HTTP ok
+     * @return an HTTP No content
      */
-    public static HttpAction ok(final String message, final WebContext context, String content) {
-        context.setResponseStatus(HttpConstants.OK);
-        context.writeResponseContent(content);
-        return new HttpAction(message, HttpConstants.OK);
+    public static HttpAction noContent(final String message, final WebContext context) {
```
victornoel: I'm wondering, do we really need this `message`? Same for `ok()` actually...

pac4j-core/src/test/java/org/pac4j/core/engine/DefaultLogoutLogicTests.java
```diff
@@ -110,7 +110,7 @@ public void testLogoutPerformed() {
         profiles.put(NAME, new CommonProfile());
         addProfilesToContext();
         call();
-        assertEquals(200, context.getResponseStatus());
+        assertEquals(204, context.getResponseStatus());
         assertEquals("", context.getResponseContent());
```
victornoel: Theoretically, an empty body is not the same than no body (null content if you want to talk in programming language terms). Not sure if this is important or not though :)

leleuj: Does it change something at the HTTP level (between empty and null)?

pac4j-core/src/main/java/org/pac4j/core/exception/HttpAction.java
leleuj: You're right, I'm not sure this is really useful. I should do some experiments...

I opened: https://github.com/pac4j/pac4j/issues/1004 to test the change.

pac4j-core/src/test/java/org/pac4j/core/engine/DefaultLogoutLogicTests.java
victornoel: With an empty content, the response can have a content-type as well as a content-length (of 0), while with no body, there should be none of the two… so if you do set an empty body, you have the risk of generating a response with this incoherent meaning…
Also it may depends on the lib that handles the response generation, maybe they won't include the body at all in case there is a 204 status.

Ok, I realise I'm nitpicking, let's not bother about that :)

leleuj: OK. So you're in line to merge it "as is"?

PR #1009, opened 2017-10-10
Use Dropbox V2 API because Dropbox has retired V1

mattes3:
Dropbox API V1 stopped working in September 2017 (it has been deprecated for a year now), and pac4j was still using it. So, OAuth with Dropbox did not work anymore.

I wrote some code to use Dropbox API V2 so that OAuth with Dropbox will work again!

Could you please merge it into master and release a new version? Would be cool.

Best wishes...
Matthias

### DIFF
--- a/pac4j-oauth/pom.xml
+++ b/pac4j-oauth/pom.xml
@@ -9,6 +9,7 @@
     </parent>
 
     <artifactId>pac4j-oauth</artifactId>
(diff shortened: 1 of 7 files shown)


leleuj:
Thanks for the upgrade. It makes sense indeed.

That said, this pull request is not synced with the master: "This branch has conflicts that must be resolved": can you please fix them?

PR #1010, opened 2017-10-11
Nicely handle the authentication errors

### LINKED ISSUE #908: Nicely handle the authentication errors
Maybe by redirecting to an error page.

What kind of errors should we catch?

leleuj:
Closes #908

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -11,6 +11,7 @@ title: Release notes&#58;
 - Upgraded to Scribe v4
 - Added "multi-tenancy" capabilities: you can dynamically define multiple callback URLs for the same client, except for SAML for which you need as many `SAML2Client` as the number of different callback URLs you want
 - The `CallbackUrlResolver` computes the callback URL (using a query parameter or a path parameter to define the client), based on the `UrlResolver` which computes a URL
+- You can define an error (page) URL at the logics' level to handle unexpected exceptions
 
 **v2.2.0**:
 
--- a/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
@@ -68,7 +68,7 @@ public class CasClient extends IndirectClient<TokenCredentials, CommonProfile> {
 
     @Override
     public String toString() {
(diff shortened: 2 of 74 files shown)


victornoel:
@leleuj in this PR you removed the `throws` from `SecurityGrantedAccessAdapter`, any reason?

Because in j2e-pac4j the adapter delegate to some servlet `FilterChain` and those `throws IOException, ServletException`!

leleuj:
@victornoel Previously, it was `throws Throwable` which is by nature a bad practice. So I completely removed it. Maybe I should have kept: `throws Exception`. We can change that easily as I tend to think that it would make more sense indeed.

Agreed?

victornoel:
Completely :)

leleuj:
Here it is: https://github.com/pac4j/pac4j/pull/1031

victornoel:
great, thanks :)

millross:
Personally, I'd ask one question. Will the J2E implementation, when it sees IOException/ServletException do anything other than wrap them in a TechnicalException? If so, I'd say the adapter should take that responsibility. Unless the J2E-pac4j actually does specific things for those exceptions which it wouldn't do for a TechnicalException, I'd expect the SecurityGrantedAccessAdapter to do the wrapping. If there will be specific, and different, handling for the checked exception cases, then it makes sense to use throws Exception

victornoel:
@millross I agree with you, but at the same time, we know that `SecurityGrantedAccessAdapter` implementations are going to wrap the exception in a `TechnicalException`, so why bother doing that for every integration lib instead of acknowledging this situation at pac4j level?

I don't think there is a good answer, just a developer's preference :)

leleuj:
In fact, if you see: https://github.com/pac4j/pac4j/pull/1031/files, I wrap the `Exception` into a `RuntimException` and not a `TechnicalException` as I consider that it's more an exception of the core framework than of pac4j itself (like an `IOException` thrown by the `FilterChain`).

Like @victornoel , I'd rather be inclined to let the pac4j core handle the wrapping than each pac4j implementation (unless there is a good reason to do the opposite).

kmandeville:
Any chance of this being in any of the 2.0 versions?

leleuj:
No, our efforts will now be focused on pac4j v3.x and upgrading the pac4j libraries to use it.

PR #1012, opened 2017-10-11
Binding not properly generated in SAML metadata

### LINKED ISSUE #989: Binding not properly generated in SAML metadata
After some investigation I noticed that even if i alter the Pac4j config object by setting: cfg.setDestinationBindingType(SAMLConstants.SAML2_REDIRECT_BINDING_URI);
the generated meta-data file still does not alter the assertionBinding. Which i can see the SAML2MetaDataGenerator line 197.. it's hardcoded to only do SAML2_POST_BINDING_URI.
The resulting assertion binding is set to: <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location=" ... " index="0"/>

leleuj:
Closes #989

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2ServiceProviderMetadataResolver.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2ServiceProviderMetadataResolver.java
@@ -48,23 +48,14 @@ public class SAML2ServiceProviderMetadataResolver implements SAML2MetadataResolv
 
     public SAML2ServiceProviderMetadataResolver(final SAML2ClientConfiguration configuration, final String callbackUrl,
                                                 final CredentialProvider credentialProvider) {
-        this(configuration.getServiceProviderMetadataResource(), callbackUrl,
-                configuration.getServiceProviderEntityId(), configuration.isForceServiceProviderMetadataGeneration(), credentialProvider,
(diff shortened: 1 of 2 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2ServiceProviderMetadataResolver.java
```diff
@@ -45,28 +44,27 @@
...
+                configuration.isAuthnRequestSigned(), configuration.getWantsAssertionsSigned(), configuration.getDestinationBindingType());
     }
 
-    private SAML2ServiceProviderMetadataResolver(final WritableResource spMetadataResource,
-                                                 final String callbackUrl,
-                                                 @Nullable final String spEntityId,
-                                                 final boolean forceSpMetadataGeneration,
-                                                 final CredentialProvider credentialProvider,
-                                                 boolean authnRequestSigned, boolean wantsAssertionsSigned) {
+    private SAML2ServiceProviderMetadataResolver(final WritableResource spMetadataResource, final String callbackUrl,
```
mmoayyed: This all looks good. 

Question: Do we really need a constructor to pass bits and pieces from `configuration` individually?

leleuj: Yes, indeed, the private constructor is useless. Will remove it.

PR #1014, opened 2017-10-11
DropBoxClient based on DropBox API V2, without conflicts on master branch

mattes3:
See https://github.com/pac4j/pac4j/pull/1009

Hi Jérôme,

I have resolved the two conflicts that I had with the master branch. They occurred because I created my patch based on V2.1.x instead of master

HOWEVER: Now, I get a NullPointerException when I run the code!

java.lang.NullPointerException
	at org.pac4j.core.client.IndirectClient.getRedirectAction(IndirectClient.java:95) ~[pac4j-core-3.0.0-SNAPSHOT.jar:?]
	at org.pac4j.core.client.IndirectClient.redirect(IndirectClient.java:68) ~[pac4j-core-3.0.0-SNAPSHOT.jar:?]
	at ratpack.pac4j.RatpackPac4j.lambda$initiateAuthentication$13(RatpackPac4j.java:518) ~[ratpack-pac4j-2.0.0.jar:?]

The source line looks like this:
final String attemptedAuth = (String) context.getSessionStore().get(context, getName() + ATTEMPTED_AUTHENTICATION_SUFFIX);

As far as I can see, the result of context.getSessionStore() is null.

Alas, I do not know Pac4j well enough to solve that problem! Could you please find out why the code in pull request #1009 works inside the V2.1.0 enivronment but this code here does not work in 3.0.0-SNAPSHOT?

Thanks...
Matthias

### DIFF
--- a/pac4j-oauth/src/main/java/org/pac4j/oauth/client/DropBoxClient.java
+++ b/pac4j-oauth/src/main/java/org/pac4j/oauth/client/DropBoxClient.java
@@ -1,8 +1,9 @@
 package org.pac4j.oauth.client;
 
 import org.pac4j.core.redirect.RedirectAction;
-import org.pac4j.oauth.profile.dropbox.DropBoxProfileDefinition;
 import org.pac4j.oauth.profile.dropbox.DropBoxProfile;
+import org.pac4j.oauth.profile.dropbox.DropBoxProfileCreator;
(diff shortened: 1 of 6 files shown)


leleuj:
Things have changed between version 2.1.0 and 3.0.0 for several reasons. I'll take a look at that...

PR #1019, opened 2017-10-16
loadProfileFromSession

### LINKED ISSUE #1015: loadProfileFromSession
The `saveProfileInSession` property is used to save the profiles into the session after a web services authentication.

It would make sense to have a `loadProfileFromSession` property as well.

leleuj:
Closes #1015 

cc @millross : is this what you had in mind?


millross:
Not quite, I'd embed the code which actually does the save and load into the strategy. I think it's also a ProfileStorageStrategy (e.g. save in session, no save would be two possible strategy implementations). 

See https://github.com/millross/pac4j-async/blob/master/pac4j-async-core/src/main/java/org/pac4j/async/core/logout/AsyncLocalLogoutStrategy.java and https://github.com/millross/pac4j-async/blob/master/pac4j-async-core/src/main/java/org/pac4j/async/core/logout/AsyncLocalLogout.java for something closer to what I have in mind.

millross:
That approach lets you provide a set of default implementations via an enum, while allowing a coder to create their own implementations if desired.

leleuj:
`ProfileStorageStrategy` is a better name. We generally have an interface and basic implementations, but in that case, as it's fairly straight-forward, I preferred to only have one implementation.

millross:
But there are really two implementations:
NoSave (equivalent of save in session, load from session being false)
and SaveToSession (equivalent of both being true).

leleuj:
In fact, we have at least three: the default one (false, false) and the two use cases I gave you: (true, false) and (false, true). I'll try to figure out something...

leleuj:
@millross is it better now?

millross:
Not really, the logic as to whether to load/save should be embedded in the strategy definition. Look at the URLs I linked to for example.

Also, what's the use case for the two boolean values being different? When would you say "never save to session but do load from session" or vice versa. 

If you really want to have independence on load/save (but I don't see the clear use case for that) then I would suggest a ProfileLoadStrategy which wraps the logic in loadProfilesFromSession and the equivalent for save. 

Different types means that you won't be as likely to accidentally assign false or true to the wrong parameter. Throwing around lots of boolean switches is far more likely to lead to error, completely unnecessarily.

leleuj:
I did take a look at your classes and tried to replicate the logic, but to avoid creating an interface and an enum with values like you did, I only created a class with a protected constructor and properties representing default implementations.
It seems very similar to me. Developers won't accidentally assign fale or true to the wrong parameter as they choose meaningfull properties.

About the use decoupling read/write, in fact, the only use case which makes sense for web services is to only use the request to load or save data thus {read: false, write: false}. So I think other options are completelly open.

millross:
In effect your strategy class behaves as an interface/enum combination anyway. You should move the save/load logic into the strategy class rather than have it just be queried for the booleans. That's rather the point of the strategy pattern, to delegate the relevant logic to the strategy. Someone can always provide an overridden strategy class if they want to change the behaviour. 

In other words, rather than asking the strategy whether you should save/load to session you should tell the strategy to perform the save/load and move that save/load logic from the SecurityLogic to the StorageStrategy. Does that clarify better? The responsibility here is not determining whether to save/load, it is the save/load.

millross:
If you want the strategy to determine the save/load combination only, but not to call ProfileManager with the correct param, then you should be passing the Strategy to ProfileManager and let the profile manager use the strategy directly. However, for me, I would let the strategy determine the parameter with which to call ProfileManager and then call ProfileManager. That would mean that if someone wanted to offer a more complex save approach (e.g. by triggering side effects such as a notification) they can do so from the strategy after a successful call to ProfileManager.

millross:
Could you also elaborate a use-case for using load from session but not save from session (or vice versa) please? I'm still not clear when someone would want to do that and your other answer just says "completely open" without enumerating when we would expect those other options to be used.

leleuj:
OK. I see your point: in that case, the `loadProfilesFromSession` and `saveProfileInSession` methods of the `DefaultSecurityLogic` should compose the `ProfileStorageStrategy`.

About the use cases, it's the ones I added on the issue:

1) you perform a UI authentication and want to call your REST API (single page application): you need to load the profiles from the session
2) you perform a web service authentication and want to use the UI then (authentication by token from a link in an email for example): you want to save the profiles into the session.

These are the cases where we need to save or load, but load and save would work in that case too.

So I came to the idea we should have a more simple solution: a single `useSessionForWebServices` boolean in the `DefaultSecurityLogic` to control the load/save of the session.

Are you in line with that?

millross:
I don't think multiple boolean switches is more simple for the consumer of our code. In my view the signature of perform is too complex already, and adding more parameters to try and support more cases is not the right answer. I think we should bite the bullet and admit that DefaultSecurityLogic should not be all things to all people, it's a default implementation. Also Web services accessing the session is _not_ a conventional pattern. I don't believe we should support it in our Default offering (other than as above by offering a strategy-based approach).

If we want to go down the "DefaultSecurityLogic is all things to all people" route then we should make it more composable via strategies for some of this stuff, not keep adding parameters to the signature of perform. 

Implementation of it for async was already a major headache, partly because of the way the logic was already constructed. We shouldn't be making it more complex, we should be seeking to reduce that complexity via delegation of responsibilities more clearly.

leleuj:
The idea is not to have a new boolean, but to replace the existing `loadProfilesFromSession` with something more generic: `useSessionForWebServices`.
And it won't be passed as an argument of the `perform` method, it will be just set/get at the class level.

Customizations happen more or less often. For "less often" use cases, I'm happy to let people override the class and do whatever they want, but this customization has been discussed and requested many times already so I tend to make things easier (with that single boolean) for this kind of use cases.

And yes, the `perform` method has too many parameters and I'll be more than happy to revisit that when we'll make pac4j asynchronous.

millross:
I don't think we're going to agree on this, but ultimately it's your codebase. I still think it's better to compose through strategies than lots of Boolean switches - it's cleaner and you can test the strategies more directly rather than trying to test more corner cases and combinations on your overall logic. You will have seen strategies and decision classes appearing in the async stuff, which reflect my desire to cut down on the boolean switches and make it easier to work with the code. I'm not going to bust a gut to accommodate new features in pac4j-async anyway, as it's effectively a candidate implementation. I'll let you manage that complexity when you integrate an async API into pac4j.

leleuj:
- "I don't think we're going to agree on this"

Quite the opposite in fact. I agree that "it's better to compose through strategies than lots of Boolean switches". See the properties inside the `DefaultSecurityLogic`:

```java
private ClientFinder clientFinder = new DefaultSecurityClientFinder();
private AuthorizationChecker authorizationChecker = new DefaultAuthorizationChecker();
private MatchingChecker matchingChecker = new RequireAllMatchersChecker();
private boolean saveProfileInSession;
```

There is only one `boolean`. And strategies to find the clients, check the authorizations and the check the matchers.

And I know that there will be refactoring on the logics more globally.

I had in mind that we could just change the meaning of the `boolean` to accommodate this need **before** diving into something bigger because I'm not completely sure of what we will do and how long it will take.

- "but ultimately it's your codebase", "I'm not going to bust a gut to accommodate new features in pac4j-async anyway, as it's effectively a candidate implementation. I'll let you manage that complexity when you integrate an async API into pac4j."

It sounds less collaborative that I would have expected.

I'm not sure to understand how one `boolean` could be such a showstopper towards the async integration, especially after the huge step made by pac4j v3 towards `pac4j-async`, but anyway, I'll try to come up with something more acceptable.

millross:
My point is: if our default position to keep adding flags rather than manage the complexity now, all you're doing is kicking the can into pac4j-async's court. Making DefaultSecurityLogic async compatible took weeks (based on the limited time commitment I can make) and involved further decomposition. We shouldn't be going in the other way. 

As regards your "collaborative" comment, for the candidate implementation of pac4j-async) I won't be reflecting updates in underlying pac4j-core classes, because if I do we'll never complete a candidate implementation. I wasn't aware that v3 had taken a huge step forward yet. How many of the "duplicated" classes will I be able to remove if I switch to building against the latest 3.x? 

For the record, if you want to throw the word "collaborative" out, I'd like to remind you that this candidate implementation has been a one-person effort. It shows (already) how to create the async API, and the changes that need to happen to core classes, their organisation and structure, to achieve that. What I'm not prepared to do is solely adopt the burden of forward integration of new functionality into that model. In other words, I do not wish (and never have) to become sole maintainer of the pac4j async API, because I do not have bandwidth to do so properly.

millross:
I'm about to eat some of my words. I've just reviewed the relevant segment of pac4j-async and in the async stuff I broke the decision on loading from session out into its own class. This class has no magic switching. 

I would suggest you create an analogue of org.pac4j.async.core.logic.decision.AsyncLoadProfileFromScheduleDecision, and feel free to add your boolean flag to its constructor (and consume it during the check in make). If you believe that lots of people are already overriding the making of that decision inline, then you can offer the same by exposing an interface they can override, I didn't do that in the pac4j-async stuff for speed of progress. Use this in DefaultSecurityLogic. Not only does this satisfy my engineering preferences, it also brings a tiny piece of the decomposition performed during production of the async security logic into core. So we both gain.

leleuj:
I haven't made any progress on this yet. I will review your latest technical comments to follow the right direction. I've been fairly busy lately (as we all are).

The idea has never been to push back work into the pac4j-async project. Far from it! If adding a `boolean` goes into the wrong direction, let's find a more acceptable solution.

To go further on this, you have achieved an incredible work to make pac4j asynchronous with the pac4j-async project. This is something we must use as a foundation to make pac4j asynchronous. 
Though, the target has never been to leave you alone maintaining on your own the pac4j-async project. The goal is to bring pac4j and pac4j-async so close that we can drop pac4j-async and implement the async nature into the pac4j project itself (maintained by all of us) without sacrificing the relative simplicity we have.

And this is what I have started to do with most of the changes brought by pac4j version 3 (without adding any real complexity): all exceptions are unchecked, the `SessionStore` is never accessed through the `WebContext` except via the `getSessionStore` method, the `InitializableWebObject` and its `init(WebContext context)` method have been removed.
With these changes, you should be able to remove all the duplicated pac4j classes, likely turn the `AsyncWebContext` into a `AsyncSessionStore` and remove the classes related to exceptions handling, as far as my understanding goes.

millross:
Not quite, we will have to type-parameterise a number of classes (because WebContext will need to be type parameterised for sync or async session store in particular) to enable easier conversion from existing classes. Just the WebContext doesn't do enough, we cannot drop in the existing versions of the duplicated classes and expect them to work (part of this is because we need to deal with anything which works against the session store in different ways).

millross:
I also missed removal of InitializableWebObject, I will schedule that in my
code as well. That's a fantastic change from an engineering point of view
(I don't like any initialization that depends on the first request to hit,
it's not very deterministic).

victornoel:
See #1032 for some discussion related to this PR.

leleuj:
@millross one more update, how does it sound this time?

millross:
Works for me, minor suggestion would be to change Strategy to Decision, as we don't actually execute the save/load, we just decide on it. But I'm pretty relaxed about that. One other minor suggestion, I don't like checks on instanceOf, especially as we will be adding new Async versions of clients. So I prefer the idea of a common base client which covers async and sync which exposes isAnonymous and isIndirect methods, then we can reuse the same concept directly between sync and async. This is impossible with tight coupling of runtime logic to sync client types, so we would end up with duplication.

millross:
However, the type checking can be moved to a separate issue, it's not directly related to the principle of this change.

leleuj:
I renamed the classes with `Decision` instead of `Strategy` for a better meaning.

Regarding the class type check, indeed, this is something that will be handled later on if ever we come with new types for Async clients.

PR #1021, opened 2017-10-17
Remove the client name from the credentials

### LINKED ISSUE #1017: Remove the client name from the credentials
We don't need it the way pac4j works. Another cleaning @victornoel

leleuj:
Closes #1017 

cc @victornoel


victornoel:
that's cool :)

PR #1028, opened 2017-10-29
Improve security logic

victornoel:
This depends on #1027.

It is needed to implement pac4j/j2e-pac4j#44 and I think it makes sense to give this information here since it's complex to retrieve it from the `WebContext` itself (because from the `WebContext`, you need to know if session as used or not and stuffs like this that only the `SecurityLogic` knows!).

I know I will exploit this also in jax-rs-pac4j for example.

This maybe highlights a problem with the current architecture too and this solution is maybe over simplistic for the problem of retrieving the profiles of a request...

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
@@ -20,7 +20,7 @@ import org.pac4j.core.matching.MatchingChecker;
 import org.pac4j.core.profile.CommonProfile;
 import org.pac4j.core.profile.ProfileManager;
 
-import java.util.List;
+import java.util.*;
 
 import static org.pac4j.core.util.CommonHelper.*;
 
@@ -131,7 +131,7 @@ public class DefaultSecurityLogic<R, C extends WebContext> extends AbstractExcep
                     logger.debug("authorizers: {}", authorizers);
                     if (authorizationChecker.isAuthorized(context, profiles, authorizers, config.getAuthorizers())) {
                         logger.debug("authenticated and authorized -> grant access");
-                        return securityGrantedAccessAdapter.adapt(context, parameters);
+                        return securityGrantedAccessAdapter.adapt(context, profiles, parameters);
                     } else {
                         logger.debug("forbidden");
                         action = forbidden(context, currentClients, profiles, authorizers);
@@ -150,7 +150,7 @@ public class DefaultSecurityLogic<R, C extends WebContext> extends AbstractExcep
             } else {
 
                 logger.debug("no matching for this request -> grant access");
-                return securityGrantedAccessAdapter.adapt(context, parameters);
+                return securityGrantedAccessAdapter.adapt(context, Arrays.asList(), parameters);
             }
 
         } catch (final Exception e) {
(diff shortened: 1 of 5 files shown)


victornoel:
See #1032 for discussion potentially invalidating this PR.

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileHelper.java
```diff
@@ -122,41 +122,25 @@ public static CommonProfile buildUserProfileByClassCompleteName(final String com
...
-            while (profilesList.hasNext()) {
-                final U nextProfile = profilesList.next();
-                if (profile == null || profile instanceof AnonymousProfile) {
-                    profile = nextProfile;
-                }
-            }
-            return Optional.of(profile);
-        }
+    public static <U extends CommonProfile> Optional<U> flatIntoOneProfile(final Collection<U> profiles) {
+        return profiles.stream().filter(p -> p != null && !(p instanceof AnonymousProfile)).findFirst();
```
leleuj: This is where we appreciate Java 8.

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
```diff
@@ -150,7 +150,7 @@ public R perform(final C context, final Config config, final SecurityGrantedAcce
             } else {
 
                 logger.debug("no matching for this request -> grant access");
-                return securityGrantedAccessAdapter.adapt(context, parameters);
+                return securityGrantedAccessAdapter.adapt(context, Arrays.asList(), parameters);
```
leleuj: Thinking about this once more, I think we should load the profiles in that case too and pass them to the `securityGrantedAccessAdapter`...

victornoel: In that case, I'm not clear how I should load them... from session or not?

Also, in this branch of the if, we are in the case where the matcher does NOT apply, why would we thus load the profiles? I think we shouldn't and consider that pac4j should not interfere at all in that case with the request.

leleuj: You should load them using the `loadProfilesFromSession` method. In fact, you just need to move the first lines of source code before the `if (matchingChecker.matches(context, matchers, config.getMatchers())) {`.

I had the same reasoning that you have. Then, I wondered what would happen if I use the `ProfileManager` even if the `DefaultSecurityLogic` has not applied? The profiles would be loaded if they existed. This is why I came to this conclusion.

victornoel: But then maybe the question is opposite: why would we want to load profiles even with the `ProfileManager` if pac4j is not meant to have played into the processing of the request?
Currently none of the security logic behaviour is applied when the filter matcher does not apply, why would we want to depart from that for the profiles?
To me, if the filter's security logic's matcher does not apply, it should behave exactly the same as if I haven't had set a filter at all.
Actually, the marchers are only there because it's not possible to conditionally apply filters with most of the frameworks pac4j is integrated with, if it was possible, then there would be no matchers in pac4j's security logic IMHO.

So I don't agree with your conclusion, but if you think it's still better to load the profile, then no problem, it's yours to make the choice, I just want to be clear about the fact I think it is counter-intuitive to do so :)

leleuj: OK. It makes sense. I'm not always right, far from it ;-) I will merge it as is.

victornoel: hehe ;) thanks!

Issue #1026, opened 2017-10-25
SAML2MetadataGenerator not honoring NameIdPolicyFormat on client configuration


scoldwell:
The SAML2MetadataGenerator outputs all NameId formats in the generated metadata even though you can specify the specific NameIdFormatPolicy via SAML2ClientConfiguration.setNameIdPolicyFormat(String).  That policy is never passed to SAML2MetadataGenerator and as such SAML2MetadataGenerator.buildNameIDFormat() ends up including all formats.

leleuj:
Thanks for reporting. Would you mind submitting a pull request to fix that? Thanks

PR #1029, opened 2017-10-30
Issue #1026

scoldwell:
Issuing pull request to fix this issue


pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2ServiceProviderMetadataResolver.java
```diff
@@ -44,12 +44,14 @@
     private final boolean forceSpMetadataGeneration;
     private boolean authnRequestSigned;
     private boolean wantsAssertionsSigned;
+    private final String nameIdPolicyFormat;
```
leleuj: In fact, we don't use that property anywhere else, it is just passed to the metadata generator. I think we could remove it.

leleuj:
Looks good overall with a minor comment.

leleuj:
And thanks for your contribution.

leleuj:
Merging. I'll make the small change by myself.

leleuj:
Forget my comment. The property was used in another method.

PR #1034, opened 2017-11-04
Clients does not seem to be initialized anymore in DefaultCallbackLogic

### LINKED ISSUE #1033: Clients does not seem to be initialized anymore in DefaultCallbackLogic
Hi,

In the past, the `DefaultCallbackLogic` would call `Clients.findClient` to find a client.
This would initialize the Clients object if needed and all the clients configured.

Now there is this `DefaultCallbackClientFinder` used and it doesn't seem to do this initialization anymore.

I think that's a bug, no? I suppose it was caught before because usually people would first go through the `DefaultSecurityLogic` (which does initialize the `Clients` object via `DefaultSecurityClientFinder`'s calls to `Clients.findClient`) first. But with jax-rs-pac4j, it happens that someone would directly go through a `DefaultCallbackLogic`.

leleuj:
Closes #1033


victornoel:
So simple :) thanks!

Issue #926, opened 2017-06-02
Update API to not return nulls, but sensitive defaults or optionals


giflw:
With the advent of​ lambda in Java and the influence of functional paradigm, and the problems that NPE can arise, it's interesting to avoid null as much as possible, changing nulls to sensitive default values or using optionals.

leleuj:
@giflw : can you elaborate on your proposal?

giflw:
Yes :relaxed:. I'll try to do until next Sunday.

giflw:
I'll try to explain what I was thinking using some examples from pac4j itself (commit 682a374f8e9e18310fd49f0b4a3bb4bb5ee77c00). As it is a security framework, I think its worthy to reason about this:

+ CommonHelper class
    ```
    /**
     * Add a new parameter to an url.
     *
     * @param url   url
     * @param name  name of the parameter
     * @param value value of the parameter
     * @return the new url with the parameter appended
     */
    public static String addParameter(final String url, final String name, final String value) {
        if (url != null) {
            final StringBuilder sb = new StringBuilder();
            sb.append(url);
            if (name != null) {
                if (url.indexOf("?") >= 0) {
                    sb.append("&");
                } else {
                    sb.append("?");
                }
                sb.append(name);
                sb.append("=");
                if (value != null) {
                    sb.append(urlEncode(value));
                }
            }
            return sb.toString();
        }
        return null;
    }
    ```
    Is this example from pac4j code, we hava a method that is supposed to return **String**, but can return null too, but the null return is unexpected, as it's not documented (@return in javadoc). In this case, the method could throw `IllegalArgumentException` if url is null, or return an `Optional`, avoiding `NullPointerException` when trying to use the returned value.
    
+ ParameterExtractor class
	```
     /**
     * Extract the right credentials. It should throw a {@link CredentialsException} in case of failure.
     *
     * @param context the current web context
     * @return the credentials
     */
    
    @Override
    public TokenCredentials extract(WebContext context) {
        final String method = context.getRequestMethod();
        if (HTTP_METHOD.GET.name().equalsIgnoreCase(method) && !supportGetRequest) {
            throw new CredentialsException("GET requests not supported");
        } else if (HTTP_METHOD.POST.name().equalsIgnoreCase(method) && !supportPostRequest) {
            throw new CredentialsException("POST requests not supported");
        }

        final String value = context.getRequestParameter(this.parameterName);
        if (value == null) {
            return null;
        }

[...]

PR #1035, opened 2017-11-06
Update API to not return nulls but Optional or sensitive defaults instead

giflw:
See #926

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/authorization/generator/AuthorizationGenerator.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/authorization/generator/AuthorizationGenerator.java
@@ -3,8 +3,6 @@ package org.pac4j.core.authorization.generator;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.profile.CommonProfile;
 
-import java.util.Optional;
-
 /**
  * Generate the authorization roles and permissions for this user profile.
  *
@@ -20,5 +18,5 @@ public interface AuthorizationGenerator<U extends CommonProfile> {
      * @param profile the user profile for which to generate the authorization information.
      * @return the updated profile or a new one
      */
-    Optional<U> generate(WebContext context, U profile);
+    U generate(WebContext context, U profile);
 }
--- a/pac4j-core/src/main/java/org/pac4j/core/authorization/generator/AuthorizationGenerator.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/authorization/generator/AuthorizationGenerator.java
@@ -3,9 +3,11 @@ package org.pac4j.core.authorization.generator;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.profile.CommonProfile;
 
+import java.util.Optional;
+
 /**
  * Generate the authorization roles and permissions for this user profile.
- * 
+ *
  * @author Jerome Leleu
  * @since 1.5.0
  */
@@ -18,5 +20,5 @@ public interface AuthorizationGenerator<U extends CommonProfile> {
      * @param profile the user profile for which to generate the authorization information.
      * @return the updated profile or a new one
      */
-    U generate(WebContext context, U profile);
+    Optional<U> generate(WebContext context, U profile);
 }
--- a/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
@@ -87,16 +87,16 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
             return Optional.empty();
         }
 
-        U profile = retrieveUserProfile(credentials, context);
-        if (profile != null) {
-            profile.setClientName(getName());
+        Optional<U> profile = retrieveUserProfile(credentials, context);
+        if (profile.isPresent()) {
+            profile.get().setClientName(getName());
             if (this.authorizationGenerators != null) {
                 for (AuthorizationGenerator<U> authorizationGenerator : this.authorizationGenerators) {
-                    profile = authorizationGenerator.generate(context, profile);
+                    profile = authorizationGenerator.generate(context, profile.get());
                 }
             }
         }
-        return Optional.of(profile);
+        return profile;
     }
 
     /**
@@ -106,8 +106,8 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
      * @param context the web context
      * @return the user profile
      */
-    protected final U retrieveUserProfile(final C credentials, final WebContext context) {
-        final U profile = this.profileCreator.create(credentials, context);
+    protected final Optional<U> retrieveUserProfile(final C credentials, final WebContext context) {
+        final Optional<U> profile = this.profileCreator.create(credentials, context);
         logger.debug("profile: {}", profile);
         return profile;
     }
--- a/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
@@ -14,7 +14,10 @@ import org.pac4j.core.util.InitializableObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Optional;
 
 /**
  * <p>This class is the default implementation of an authentication client (whatever the mechanism). It has the core concepts:</p>
@@ -38,7 +41,7 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
 
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
-    private String name = this.getClass().getSimpleName();
+    private String name;
 
     // As the most common access would be through iteration then random, LinkedList is more efficient
     private List<AuthorizationGenerator<U>> authorizationGenerators = new LinkedList<>();
@@ -49,6 +52,10 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
 
     private ProfileCreator<C, U> profileCreator = AuthenticatorProfileCreator.INSTANCE;
 
+    public BaseClient() {
+        this.setName(null);
+    }
+
     /**
      * Retrieve the credentials.
      *
@@ -58,13 +65,15 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
     protected Optional<C> retrieveCredentials(final WebContext context) {
         try {
             final Optional<C> credentials = this.credentialsExtractor.extract(context);
-            final long t0 = System.currentTimeMillis();
-            try {
-                credentials.ifPresent(c -> this.authenticator.validate(c, context));
-            } finally {
-                final long t1 = System.currentTimeMillis();
-                logger.debug("Credentials validation took: {} ms", t1 - t0);
-            }
+            credentials.ifPresent(c -> {
+                final long t0 = System.currentTimeMillis();
+                try {
+                    this.authenticator.validate(c, context);
+                } finally {
+                    final long t1 = System.currentTimeMillis();
+                    logger.debug("Credentials validation took: {} ms", t1 - t0);
+                }
+            });
             return credentials;
         } catch (CredentialsException e) {
             logger.info("Failed to retrieve or validate credentials: {}", e.getMessage());
@@ -86,7 +95,9 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
             profile.get().setClientName(getName());
             if (this.authorizationGenerators != null) {
                 for (AuthorizationGenerator<U> authorizationGenerator : this.authorizationGenerators) {
-                    profile = authorizationGenerator.generate(context, profile.get());
+                    profile = Optional.of(
+                        authorizationGenerator.generate(context, profile.get())
+                    );
                 }
             }
         }
@@ -106,17 +117,17 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
         return profile;
     }
 
+    @Override
+    public String getName() {
+        return this.name;
+    }
+
     public void setName(final String name) {
         if (CommonHelper.isBlank(this.name)) {
             this.name = this.getClass().getSimpleName();
         } else {
             this.name = name;
         }
-    }
-
-    @Override
-    public String getName() {
-        return this.name;
     }
 
     /**
@@ -156,6 +167,10 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
         return credentialsExtractor;
     }
 
+    public void setCredentialsExtractor(final CredentialsExtractor<C> credentialsExtractor) {
+        this.credentialsExtractor = credentialsExtractor;
+    }
+
     protected void defaultCredentialsExtractor(final CredentialsExtractor<C> credentialsExtractor) {
         if (this.credentialsExtractor == null) {
             this.credentialsExtractor = credentialsExtractor;
@@ -164,6 +179,10 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
 
     public Authenticator<C> getAuthenticator() {
         return authenticator;
+    }
+
+    public void setAuthenticator(final Authenticator<C> authenticator) {
+        this.authenticator = authenticator;
     }
 
     protected void defaultAuthenticator(final Authenticator<C> authenticator) {
@@ -176,22 +195,14 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
         return profileCreator;
     }
 
+    public void setProfileCreator(final ProfileCreator<C, U> profileCreator) {
+        this.profileCreator = profileCreator;
+    }
+
     protected void defaultProfileCreator(final ProfileCreator<C, U> profileCreator) {
         if (this.profileCreator == null || this.profileCreator == AuthenticatorProfileCreator.INSTANCE) {
             this.profileCreator = profileCreator;
         }
-    }
-
-    public void setCredentialsExtractor(final CredentialsExtractor<C> credentialsExtractor) {
-        this.credentialsExtractor = credentialsExtractor;
-    }
-
-    public void setAuthenticator(final Authenticator<C> authenticator) {
-        this.authenticator = authenticator;
-    }
-
-    public void setProfileCreator(final ProfileCreator<C, U> profileCreator) {
-        this.profileCreator = profileCreator;
     }
 
     @Override
--- a/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
@@ -1,9 +1,5 @@
 package org.pac4j.core.client;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-
 import org.pac4j.core.authorization.generator.AuthorizationGenerator;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.credentials.Credentials;
@@ -17,6 +13,11 @@ import org.pac4j.core.util.CommonHelper;
 import org.pac4j.core.util.InitializableObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
 
 /**
  * <p>This class is the default implementation of an authentication client (whatever the mechanism). It has the core concepts:</p>
@@ -56,11 +57,11 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
      * @param context the web context
      * @return the credentials
      */
-    protected C retrieveCredentials(final WebContext context) {
+    protected Optional<C> retrieveCredentials(final WebContext context) {
         try {
             final C credentials = this.credentialsExtractor.extract(context);
             if (credentials == null) {
-                return null;
+                return Optional.empty();
             }
             final long t0 = System.currentTimeMillis();
             try {
@@ -69,21 +70,21 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
                 final long t1 = System.currentTimeMillis();
                 logger.debug("Credentials validation took: {} ms", t1 - t0);
             }
-            return credentials;
+            return Optional.of(credentials);
         } catch (CredentialsException e) {
             logger.info("Failed to retrieve or validate credentials: {}", e.getMessage());
             logger.debug("Failed to retrieve or validate credentials", e);
 
-            return null;
+            return Optional.empty();
         }
     }
 
     @Override
-    public final U getUserProfile(final C credentials, final WebContext context) {
+    public final Optional<U> getUserProfile(final C credentials, final WebContext context) {
         init();
         logger.debug("credentials : {}", credentials);
         if (credentials == null) {
-            return null;
+            return Optional.empty();
         }
 
         U profile = retrieveUserProfile(credentials, context);
@@ -95,7 +96,7 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
                 }
             }
         }
-        return profile;
+        return Optional.of(profile);
     }
 
     /**
--- a/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
@@ -14,10 +14,7 @@ import org.pac4j.core.util.InitializableObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 
 /**
  * <p>This class is the default implementation of an authentication client (whatever the mechanism). It has the core concepts:</p>
@@ -41,9 +38,10 @@ public abstract class BaseClient<C extends Credentials, U extends CommonProfile>
 
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
-    private String name;
+    private String name = this.getClass().getSimpleName();
 
-    private List<AuthorizationGenerator<U>> authorizationGenerators = new ArrayList<>();
+    // As the most common access would be through iteration then random, LinkedList is more efficient
+    private List<AuthorizationGenerator<U>> authorizationGenerators = new LinkedList<>();
 
     private CredentialsExtractor<C> credentialsExtractor;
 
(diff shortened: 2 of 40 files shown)


pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
```diff
@@ -95,7 +96,7 @@ public final U getUserProfile(final C credentials, final WebContext context) {
                 }
             }
         }
-        return profile;
+        return Optional.of(profile);
```
victornoel: ofNullable maybe, no?

victornoel: (because can be null at that point.
And actually why `retrieveUserProfile` does not return an `Optional`?

pac4j-core/src/main/java/org/pac4j/core/client/Client.java
```diff
@@ -56,7 +58,14 @@
      * @param context web context
      * @return the user profile
      */
-    U getUserProfile(C credentials, WebContext context);
+    Optional<U> getUserProfile(C credentials, WebContext context);
+
+    default Optional<U> getUserProfile(Optional<C> credentials, WebContext context) {
+        if(credentials.isPresent()) {
```
victornoel: use map()

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
```diff
@@ -135,9 +137,9 @@ public final C getCredentials(final WebContext context) {
     }
 
     @Override
-    public final RedirectAction getLogoutAction(final WebContext context, final U currentProfile, final String targetUrl) {
+    public final Optional<RedirectAction> getLogoutAction(final WebContext context, final U currentProfile, final String targetUrl) {
         init();
-        return logoutActionBuilder.getLogoutAction(context, currentProfile, targetUrl);
+        return Optional.of(logoutActionBuilder.getLogoutAction(context, currentProfile, targetUrl));
```
victornoel: ofNullable?

pac4j-core/src/main/java/org/pac4j/core/context/session/J2ESessionStore.java
```diff
@@ -59,7 +59,7 @@ public Object getTrackableSession(final J2EContext context) {
         if (trackableSession != null) {
             return new J2EProvidedSessionStore((HttpSession) trackableSession);
         } else {
-            return null;
+            return (SessionStore<J2EContext>) SessionStore.EMPTY;
```
victornoel: I have no opinion on this, but is it really what we would want? to throw exception while before we would be returning `null`? (I haven't finished reading, so maybe I will get an answer in the rest of the diff : )

pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
```diff
@@ -56,11 +57,11 @@
      * @param context the web context
      * @return the credentials
      */
-    protected C retrieveCredentials(final WebContext context) {
+    protected Optional<C> retrieveCredentials(final WebContext context) {
         try {
             final C credentials = this.credentialsExtractor.extract(context);
```
millross: Why not make credentialsExtractor return an optional?

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
millross: If we're going to temporarily facade null returns with an Optional until we can replace those null returns with Optional returns (as it appears), then I agree with @victornoel - we should be using ofNullable.

Ideally logoutActionBuilder would return an Optional.

pac4j-core/src/main/java/org/pac4j/core/context/ContextHelper.java
```diff
@@ -19,15 +20,15 @@
      * @param name the name of the cookie
      * @return the cookie
      */
-    public static Cookie getCookie(final Collection<Cookie> cookies, final String name) {
+    public static Optional<Cookie> getCookie(final Collection<Cookie> cookies, final String name) {
         if (cookies != null) {
```
millross: Surely this can be done better using streaming
cookies.stream.filter(cookie != null && CommonHelper.equals...).findFirst()

This will return an Optional directly, or empty.

giflw:
In the case of Optional.of(null), it throws null pointer exception, as null is not allowed as Optional value. But the benefit is the NPE will be thrown on the origin of null value.

victornoel:
> In the case of Optional.of(null), it throws null pointer exception, as null is not allowed as Optional value. But the benefit is the NPE will be thrown on the origin of null value.

This is ok except for the cases where we actually want to convert from nullable to an optional. Those were the cases where I proposed to change to `ofNullable` :)

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
victornoel: yes, if we are going to change things drastically, then also change the return type of everything you think is potentially returing null (and then don't use `ofNullable` in that case of course).

millross:
In the case of Optional.of(null), it throws null pointer exception, as null is not allowed as Optional value. But the benefit is the NPE will be thrown on the origin of null value.

But that assumes that the null value is incorrect. In most cases null/empty is a legitimate result. I would prefer if we're doing something where null isn't permitted that we do an Objects.requireNonNull at that position as an explicit statement that null is not allowed here, rather than have the behaviour being implicit via Optional.of.

giflw:
@victornoel That's the main goal :). I was trying to reply to @milross, but pressed the wrong button. My point in that was just Optional.of(null) throws exception instead of returning a Optional with the null value.

giflw:
@millross I agree. Those cases we must identify to use the right call. In this first change I just replace all `return null` in the core by Optional. Now we need to choose between `of` and `ofNullable`. In fact, if we find that in somewhere `null` is not acceptable, we should throw an appropriated exception.

pac4j-core/src/main/java/org/pac4j/core/context/session/J2ESessionStore.java
leleuj: We should return an `Optional` here as it might be `null` when this capability is not supported.

pac4j-core/src/main/java/org/pac4j/core/context/session/SessionStore.java
```diff
@@ -10,62 +11,78 @@
...
     /**
      * Get the object from its key in store.
      *
      * @param context the web context
-     * @param key the key of the object
+     * @param key     the key of the object
      * @return the object in store
      */
-    Object get(C context, String key);
+    default Object get(C context, String key) {
```
leleuj: Shouldn't we have an `Optional` here?

pac4j-core/src/main/java/org/pac4j/core/context/session/SessionStore.java
```diff
@@ -10,62 +11,78 @@
...
+    }
 
     /**
      * Get the native session as a trackable object.
      *
      * @param context the web context
      * @return the trackable object or <code>null</code> if this is not supported
      */
-    Object getTrackableSession(C context);
+    default Object getTrackableSession(C context) {
```
leleuj: It might not be implemented: an `Optional` could be returned...

pac4j-core/src/main/java/org/pac4j/core/context/session/SessionStore.java
```diff
@@ -10,62 +11,78 @@
...
     /**
      * Build a new session store from a trackable session.
      *
-     * @param context the web context
+     * @param context          the web context
      * @param trackableSession the trackable session
      * @return the new session store or <code>null</code> if this is not supported
      */
-    SessionStore<C> buildFromTrackableSession(C context, Object trackableSession);
+    default SessionStore<C> buildFromTrackableSession(C context, Object trackableSession) {
```
leleuj: `Optional` here also?

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
```diff
@@ -6,37 +6,40 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
 
 import static org.pac4j.core.util.CommonHelper.*;
 
 /**
  * <p>Default security logic:</p>
- *
+ * <p>
```
leleuj: Why adding a single `<p>` here?

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
```diff
@@ -163,12 +166,25 @@ public R perform(final C context, final Config config, final SecurityGrantedAcce
...
 
     /**
+     * Load the profiles from the web context if no clients are defined or if the first client is an indirect one
+     * or the {@link AnonymousClient}.
+     *
+     * @param context        the web context
+     * @param currentClients the current clients
+     * @return whether the profiles must be loaded from the web session
+     */
+    protected boolean loadProfilesFromSession(final C context, final List<Client> currentClients) {
```
leleuj: This has been changed recently, you are not up-to-date with the master...

pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
giflw: In fact I think the Optional should come from `ProfileCreator` and `AuthorizationGenerator`, which are used by `retrieveUserProfile`.

pac4j-core/src/main/java/org/pac4j/core/client/Client.java
giflw: I'm questioning which `getUserProfile` should be default or removed: with optional or accepting null?

pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
giflw: In fact this change should be applied to `ProfileCreator` and `AuthorizationGenerator`

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
giflw: Sorry, that was my IDE

giflw:
I've made the changes in core API as revision above, but couldn't finish the changes in the implementations. As soon I can I'll propagate API changes to implementations.

pac4j-core/src/main/java/org/pac4j/core/client/Client.java
leleuj: Why do we have two `getUserProfile` methods in the `Client` interface? This is the most important class of *pac4j*: we should keep it as clean as possible.

pac4j-core/src/main/java/org/pac4j/core/context/session/J2ESessionStore.java
```diff
@@ -30,8 +31,8 @@ public String getOrCreateSessionId(final J2EContext context) {
     }
 
     @Override
-    public Object get(final J2EContext context, final String key) {
-        return getHttpSession(context).getAttribute(key);
+    public Optional get(final J2EContext context, final String key) {
+        return Optional.of(getHttpSession(context).getAttribute(key));
```
victornoel: OfNullable I suppose ?

pac4j-core/src/main/java/org/pac4j/core/context/session/J2ESessionStore.java
```diff
@@ -50,17 +51,13 @@ public boolean destroySession(final J2EContext context) {
     }
 
     @Override
-    public Object getTrackableSession(final J2EContext context) {
-        return getHttpSession(context);
+    public Optional getTrackableSession(final J2EContext context) {
+        return Optional.of(getHttpSession(context));
```
victornoel: ofNulllable I suppose?

pac4j-core/src/main/java/org/pac4j/core/context/session/J2ESessionStore.java
```diff
@@ -50,17 +51,13 @@ public boolean destroySession(final J2EContext context) {
...
     }
 
     @Override
-    public SessionStore<J2EContext> buildFromTrackableSession(final J2EContext context, final Object trackableSession) {
-        if (trackableSession != null) {
-            return new J2EProvidedSessionStore((HttpSession) trackableSession);
-        } else {
-            return null;
-        }
+    public Optional<SessionStore<J2EContext>> buildFromTrackableSession(final J2EContext context, final Object trackableSession) {
```
victornoel: shouldn't trackableSession be an Optional?

Either they are both or none are for me… actually, I don't think we really need to introduce Optional here, it is the caller that should be careful of not calling this with a null value!

pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/Authenticator.java
```diff
@@ -14,10 +14,11 @@
 public interface Authenticator<C extends Credentials> {
 
     /**
-     * Validate the credentials. It should throw a {@link CredentialsException} in case of failure.
+     * Validate the credentials. It must throw a {@link CredentialsException} in case of failure.
+     * As the exception is the validation mecanism, the absent of it indicates valid credentials.
```
victornoel: the absence

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/BasicAuthExtractor.java
```diff
@@ -28,26 +29,28 @@ public BasicAuthExtractor(final String headerName, final String prefixHeader) {
...
-
-        final int delim = token.indexOf(":");
-        if (delim < 0) {
-            throw new CredentialsException("Bad format of the basic auth header");
-        }
-        return new UsernamePasswordCredentials(token.substring(0, delim),
-                token.substring(delim + 1));
+    public Optional<UsernamePasswordCredentials> extract(WebContext context) {
+        final Optional<TokenCredentials> credentials = this.extractor.extract(context);
+        return credentials.flatMap( c -> {
```
victornoel: why flatMap if you end up always returning a non-empty optional in the lambda?

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/CredentialsExtractor.java
```diff
@@ -4,6 +4,8 @@
 import org.pac4j.core.credentials.Credentials;
 import org.pac4j.core.exception.CredentialsException;
 
+import java.util.Optional;
+
 /**
  * An extractor gets the {@link Credentials} from a {@link WebContext} and should return <code>null</code> if no credentials are present
```
victornoel: remove the reference to null

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/FormExtractor.java
```diff
@@ -21,14 +23,14 @@ public FormExtractor(final String usernameParameter, final String passwordParame
     }
 
     @Override
-    public UsernamePasswordCredentials extract(WebContext context) {
+    public Optional<UsernamePasswordCredentials> extract(WebContext context) {
         final String username = context.getRequestParameter(this.usernameParameter);
```
victornoel: I'm surprised that getRequestParameter does not return an Optional :)

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
```diff
@@ -22,17 +24,17 @@ public HeaderExtractor(final String headerName, final String prefixHeader) {
     }
 
     @Override
-    public TokenCredentials extract(WebContext context) {
+    public Optional<TokenCredentials> extract(WebContext context) {
         final String header = context.getRequestHeader(this.headerName);
```
victornoel: I'm surprised that getRequestHeader does not return an optional :)

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/ParameterExtractor.java
```diff
@@ -33,20 +35,16 @@ public ParameterExtractor(final String parameterName, final boolean supportGetRe
...
         }
 
-        final String value = context.getRequestParameter(this.parameterName);
-        if (value == null) {
-            return null;
-        }
-
-        return new TokenCredentials(value);
+        return Optional.ofNullable(context.getRequestParameter(this.parameterName))
+            .map((v) -> new TokenCredentials(v));
```
victornoel: TokenCredentials::new

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java
```diff
@@ -78,10 +78,10 @@ public R perform(final C context, final Config config, final HttpActionAdapter<R
             logger.debug("client: {}", client);
             assertNotNull("client", client);
 
-            final Credentials credentials = client.getCredentials(context);
+            final Optional<Credentials> credentials = client.getCredentials(context);
             logger.debug("credentials: {}", credentials);
 
-            final CommonProfile profile = client.getUserProfile(credentials, context);
+            final Optional<CommonProfile> profile = credentials.flatMap( c -> client.getUserProfile(c, context));
```
victornoel: nice :)

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java
```diff
@@ -129,7 +136,7 @@ protected void renewSession(final C context, final Config config) {
     }
 
     protected HttpAction redirectToOriginallyRequestedUrl(final C context, final String defaultUrl) {
-        final String requestedUrl = (String) context.getSessionStore().get(context, Pac4jConstants.REQUESTED_URL);
+        final String requestedUrl = (String) context.getSessionStore().get(context, Pac4jConstants.REQUESTED_URL).get();
```
victornoel: .get() alone…

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
```diff
@@ -202,19 +218,19 @@ protected void saveRequestedUrl(final C context, final List<Client> currentClien
...
      *
-     * @param context the web context
+     * @param context        the web context
      * @param currentClients the current clients
      * @return the performed redirection
      */
     protected HttpAction redirectToIdentityProvider(final C context, final List<Client> currentClients) {
         final IndirectClient currentClient = (IndirectClient) currentClients.get(0);
-        return currentClient.redirect(context);
+        return (HttpAction) currentClient.redirect(context).get();
```
victornoel: .get() alone…

pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
```diff
@@ -78,7 +78,7 @@ public ProfileManager(final WebContext context, final SessionStore sessionStore)
             }
         }
         if (readFromSession) {
-            final Object sessionAttribute = this.sessionStore.get(this.context, Pac4jConstants.USER_PROFILES);
+            final Object sessionAttribute = this.sessionStore.get(this.context, Pac4jConstants.USER_PROFILES).get();
```
victornoel: .get() alone…

pac4j-core/src/main/java/org/pac4j/core/profile/creator/ProfileCreator.java
```diff
@@ -4,6 +4,8 @@
 import org.pac4j.core.credentials.Credentials;
 import org.pac4j.core.profile.CommonProfile;
 
+import java.util.Optional;
+
 /**
  * This interface is responsible to create a {@link CommonProfile} from a {@link Credentials}.
  * Return <code>null</code> if no profile can be found / created.
```
victornoel: remove references to null :)

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/FormExtractor.java
giflw: As the return type is String, we could use empty one to denote absence. What you think?

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
giflw: Same as above...

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/FormExtractor.java
victornoel: That it is a very bad idea ^^
I think that a request parameter can be present and have no value furthermore, but anyway, `Optional` is a good idea!

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
victornoel: same as above :)

leleuj:
@giflw @victornoel what's the status of this PR? Have all comments been accounted for?

victornoel:
@leleuj my comments are still valid, there is some work to finish this. I can try to take over if @giflw doesn't have the time.

pac4j-core/src/main/java/org/pac4j/core/authorization/generator/AuthorizationGenerator.java
```diff
@@ -18,5 +20,5 @@
      * @param profile the user profile for which to generate the authorization information.
      * @return the updated profile or a new one
      */
-    U generate(WebContext context, U profile);
+    Optional<U> generate(WebContext context, U profile);
```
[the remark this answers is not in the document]
giflw: In fact its assuming that `U profile` is not null. So NPE would be raised before `Optional.of`.

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java
giflw: Thanks =D

pac4j-core/src/main/java/org/pac4j/core/authorization/generator/AuthorizationGenerator.java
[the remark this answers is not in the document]
victornoel: exactly, so what's the point in wrapping it into an `Optional`? The idea is not to make everything `Optional`, is to use `Optional` when things CAN be `null`, not the opposite.

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
giflw: How should we deal with no HttpAction returned? I've thrown an Error by now.

giflw:
Hi @leleuj and @victornoel! Sorry for taking so long, but I'm having no much time lately. I reviewed all comments and the core is compiling. I'll try to make any further changes faster.

pac4j-core/src/main/java/org/pac4j/core/authorization/generator/AuthorizationGenerator.java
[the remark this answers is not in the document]
giflw: Changed back in commit d86cee4b7690b28a400a8daefb8d6efeb2eb9483

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
leleuj: I don't think it makes sense to have an optional `HttpAction`  here. The signature of the `redirect` method should not have an `Optional`.
As we are working on the API, I'm wondering if this `redirect` should belong to the `IndirectClient`  class instead of the `Client` interface.

pac4j-core/src/main/java/org/pac4j/core/client/Client.java
```diff
@@ -38,7 +40,7 @@
      * @param context the current web context
      * @return the performed redirection
      */
-    HttpAction redirect(WebContext context);
+    Optional<HttpAction> redirect(WebContext context);
```
leleuj: I don't think we should make this output `Optional`, functionally speaking, it doesn't make any sense.

And I'd like to asses the move of this method in the `IndirectClient` class.

pac4j-core/src/main/java/org/pac4j/core/client/DirectClient.java
```diff
@@ -32,18 +33,18 @@ protected final void internalInit() {
     protected abstract void clientInit();
 
     @Override
-    public final HttpAction redirect(final WebContext context) {
-        throw new TechnicalException("direct clients do not support redirections");
+    public final Optional<HttpAction> redirect(final WebContext context) {
+        return Optional.empty();
```
leleuj: Let's keep the old code here or let's try to move this method in the `IndirectClient`?

pac4j-core/src/main/java/org/pac4j/core/context/session/SessionStore.java
```diff
@@ -48,16 +51,16 @@
...
     /**
      * Build a new session store from a trackable session.
      *
-     * @param context the web context
+     * @param context          the web context
      * @param trackableSession the trackable session
-     * @return the new session store or <code>null</code> if this is not supported
+     * @return the new session store
      */
     SessionStore<C> buildFromTrackableSession(C context, Object trackableSession);
```
leleuj: This should be an `Optional<SessionStore<C>>` as a new session store could be built from an optional `trackableSession`.

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
leleuj: Did you clean the added `<p>`? I still see it in the Github diff...

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
leleuj: Sync with master!

leleuj:
I knew it would be a hard job to do... Thanks for your efforts.

I have made a few comments here and there, but my main concern is the `Optional` on the `redirect` method: this doesn't make any sense and we should move this method in the `IndirectClient`.

giflw:
What about an IndirectActions object?

```java
interface Client {
    String getName();
    Optional<C> getCredentials(...);
    Optional<U> getUserProfile(...);
    Optional<IndirectActions> getIndirectActions();
}

interface IndirectActions {
    HttpAction redirect(...);
    Optional<RedirectAction> getLogoutAction(...);
}
```

This way, the difference between direct and indirect clients would be the return value from `getIndirectActions()`, allowing easy interchange between direct and indirect clients.

pac4j-core/src/main/java/org/pac4j/core/credentials/authenticator/Authenticator.java
```diff
@@ -14,10 +14,11 @@
...
     /**
-     * Validate the credentials. It should throw a {@link CredentialsException} in case of failure.
+     * Validate the credentials. It must throw a {@link CredentialsException} in case of failure.
+     * As the exception is the validation mecanism, the absence of it indicates valid credentials.
      *
      * @param credentials the given credentials
      * @param context the web context
      */
-    void validate(C credentials, WebContext context);
+    void validate(C credentials, WebContext context) throws  CredentialsException;
```
leleuj: It's a runtime exception, why keeping it in the `throws` clause?

pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
```diff
@@ -257,10 +257,10 @@ public boolean containsAuthenicationAttribute(final String name) {
      * @param <T> The type of the authentication attribute
      * @return the named attribute
      */
-    public <T> T getAuthenticationAttribute(final String name, final Class<T> clazz)
+    public <T> Optional<T> getAuthenticationAttribute(final String name, final Class<T> clazz)
```
leleuj: Why not add an `Optional` to the `getAttribute` and `getAttributeByType` methods?

leleuj:
The `IndirectActions` is not a bad idea, but we already have the `IndirectClient`, `DirectClient` and ` Client` components for a long time, so I'd prefer to stick with them.
That said, I realize it's a new change, distinct from the "Optional" one. So we should better tackle it in a new PR. I opened: https://github.com/pac4j/pac4j/issues/1069

@victornoel are you ok with this PR? I see you "requested changes"...

PR #1037, opened 2017-11-10
enforce renew when validating CAS tickets

frett:
When renew was enabled in the `CasConfiguration` it would add the parameter to the login redirect, but not enforce that the ticket returned requires a new session when validating the ticket.

This PR will configure the default `TicketValidators` to respect the renew setting in `CasConfiguration`


leleuj:
Good catch! I think the bug has always been there...

frett:
possibly, stumbled across it as I was doing some integration testing

Issue #1043, opened 2017-11-24
Retrieve Access Token from Refresh Token stored in OIDC profile


mores:
Here are the docs to add support for the AzureAdClient: https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-protocols-oauth-code#refreshing-the-access-tokens

If you can provide some direction ( gist ) I can make a PR.

leleuj:
Would you mind submitting a pull request? Thanks

mores:
DiscoveryURI is "https://login.microsoftonline.com/" + microsoftTenant + "/.well-known/openid-configuration"
AcceessTokenFromRefreshURI is "https://login.microsoftonline.com/" + microsoftTenant + "/oauth2/token"

What would be the best way to configure the microsoftTenant inside of pac4j ?

leleuj:
No, I don't have a better idea, it seems the best option.

kunalsumbly:
Hi ..wanted to know if we have the feature of access token and refresh token available in normal oidc client as well. e.g. if we want to check the expiry of the access token which we got from an OIDC server and use refresh token to get a new access token after the previous access token is no more valid.

leleuj:
Please use the mailing list for questions. And yes: https://github.com/pac4j/pac4j/blob/3.6.x/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/OidcProfile.java#L107

PR #1047, opened 2017-11-29
Get access token from refresh token

mores:
https://github.com/pac4j/pac4j/issues/1043

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
@@ -1,6 +1,20 @@
 package org.pac4j.oidc.profile.azuread;
 
+import org.pac4j.core.context.HttpConstants;
+import org.pac4j.core.exception.TechnicalException;
+import org.pac4j.core.util.CommonHelper;
+import org.pac4j.core.util.HttpUtils;
+import org.pac4j.oidc.config.AzureAdOidcConfiguration;
 import org.pac4j.oidc.profile.OidcProfile;
+
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * <p>This class is the user profile for Azure AD (using OpenID Connect protocol) with appropriate getters.</p>
@@ -13,14 +27,10 @@ public class AzureAdProfile extends OidcProfile {
 
     private static final long serialVersionUID = -8659029290353954198L;
 
-    private String tenant = null;
+    private AzureAdOidcConfiguration config = null;
 
-    public String getTenant() {
-        return this.tenant;
-    }
-
-    public void setTenant(String tenant) {
-        this.tenant = tenant;
+    public void setConfiguration(AzureAdOidcConfiguration config) {
+        this.config = config;
     }
 
     public String getIdp() {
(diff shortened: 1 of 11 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfileCreator.java
```diff
@@ -13,12 +15,19 @@
  * @author Jerome Leleu
  * @since 1.9.2
  */
-public class AzureAdProfileCreator extends OidcProfileCreator<AzureAdProfile> {
+public class AzureAdProfileCreator extends OidcProfileCreator<AzureAdProfile,AzureAdOidcConfiguration> {
```
leleuj: In fact, we don't care about the configuration type, do we really need this change?

leleuj:
Looks good to me, just a minor comment.

Can you update the doucmentation as well? -> https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/openid-connect.md

I remember a concern when updating the OIDC support for CAS: @mmoayyed : are you in line with this PR?

pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfileCreator.java
mores: I can make it work without this change, but then I would store the tenant in the OidcConfiguration not in the AzureAdOidcConfiguration. I don't think that is what you want.

Can you suggest a way to make this work without this change ?

pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
```diff
@@ -46,4 +66,33 @@ public String getUsername() {
         return (String) getAttribute(AzureAdProfileDefinition.UPN);
     }
 
+    public String getAccessTokenFromRefreshToken() {
```
mores: Currently the String that is returned is json with attributes of:
token_type, scope, expires_in, ext_expires_in, expires_on, not_before, resource, access_token, refresh_token

Should it be left as is, or return some sort of AzureAdResponse ?

mmoayyed:
Reviewed. Looks fine to me.

pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
leleuj: In fact, I missed the point here but this method should be in the `AzureAdClient`, taking an `AzureAdProfile` as input and returning only an access token.

pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfileCreator.java
leleuj: For sure, we need the `AzureAdOidcConfiguration` to hold the tenant.

My point is: should we have the `OidcProfileCreator` parametrized with `? extends OidcConfiguration`?

pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
mores: How should the end user get a reference to the AzureAdClient ?

leleuj: The `AzureAdClient` has been defined in the `Config`  component, so `config.findClient("AzureAdClient")` will do the job...

mores: I don't think that is 100% correct....Here is the (verbose) gist as I see it:

`
org.pac4j.core.config.Config config = org.pac4j.core.config.ConfigSingleton.getConfig();
                org.pac4j.core.client.Clients clients = config.getClients();
                org.pac4j.oidc.client.AzureAdClient azureAdClient = clients.findClient( org.pac4j.oidc.client.AzureAdClient.class );
String accessToken = azureAdClient.getAccessTokenFromRefreshToken( azureAdProfile );
`

Or the one liner gist if you prefer:

`String accessToken = org.pac4j.core.config.ConfigSingleton.getConfig().getClients().findClient( org.pac4j.oidc.client.AzureAdClient.class ).getAccessTokenFromRefreshToken( azureAdProfile );`

If you approve this gist I will open up a second Pull Request.

leleuj: In fact, you don't need to open a second PR: just move the `getAccessTokenFromRefreshToken` method from the `AzureAdProfile` to the `AzureAdClient`. Thanks

PR #1050, opened 2017-12-05
Merge both 'authRequestSigned'?

### LINKED ISSUE #1044: Merge both "authRequestSigned"?
I noticed we have: `forceSignRedirectBindingAuthnRequest` and `authnRequestSigned`: shouldn't we merge both in a `mustSignAuthRequest`?

@mmoayyed : what do you think?

leleuj:
Closes #1044

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
@@ -65,8 +65,6 @@ public class SAML2ClientConfiguration extends InitializableObject {
     private boolean forceAuth = false;
     private boolean passive = false;
 
-    private boolean forceSignRedirectBindingAuthnRequest;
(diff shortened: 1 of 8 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2ClientConfiguration.java
```diff
@@ -81,7 +79,7 @@
 
     private SAMLMessageStorageFactory samlMessageStorageFactory = new EmptyStorageFactory();
 
-    private boolean authnRequestSigned = true;
+    private boolean authnRequestSigned;
```
leleuj: @mmoayyed : Just wondering, should we sign authentication requests by default?

leleuj: OK. So I will merge this PR as is. Thanks

PR #1058, opened 2017-12-14
add issue time set to current date to claimset matching the comment

andre-doherty:
add issue time to claimset.

All tests are running ok


leleuj:
Thanks

PR #1060, opened 2017-12-14
add issue time set to current date to claimset matching the comment

andre-doherty:
add issuetime to current date to jwt claim builder


leleuj:
Thanks

Issue #1064, opened 2017-12-16
[AzureAdClient] reauthenticate when id_token is expired


mores:
The AzureAdProfile should only be valid while the ID token is valid.

https://docs.microsoft.com/en-us/azure/active-directory/active-directory-configurable-token-lifetimes#id-tokens

PR #1065, opened 2017-12-17
AzureAD id_token is refreshed when it is expired.

mores:
https://github.com/pac4j/pac4j/issues/1064

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
@@ -21,6 +21,16 @@ public class AzureAdProfile extends OidcProfile {
 
     private static final long serialVersionUID = -8659029290353954198L;
 
+    private int idTokenExpireAdvance;
+
+    public AzureAdProfile() {
+        this.idTokenExpireAdvance = 10;
+    }
+
+    public AzureAdProfile(int idTokenExpireAdvance) {
+        this.idTokenExpireAdvance = idTokenExpireAdvance;
+    }
+
     public String getIdp() {
         return (String) getAttribute(AzureAdProfileDefinition.IDP);
     }
@@ -62,7 +72,7 @@ public class AzureAdProfile extends OidcProfile {
             Date expiresOn = claims.getExpirationTime();
 
             Calendar now = Calendar.getInstance();
-            now.add( Calendar.MINUTE, 5 );
+            now.add( Calendar.SECOND, idTokenExpireAdvance );
 
             if (expiresOn.before(now.getTime())) {
                 return true;
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
@@ -1,5 +1,13 @@
 package org.pac4j.oidc.profile.azuread;
 
+import java.text.ParseException;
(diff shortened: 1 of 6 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
```diff
@@ -46,4 +54,23 @@ public String getUsername() {
...
 
+    @Override
+    public boolean isExpired() {
+        try {
+            JWT jwt = this.getIdToken();
+            JWTClaimsSet claims = jwt.getJWTClaimsSet();
+            Date expiresOn = claims.getExpirationTime();
+
+            Calendar now = Calendar.getInstance();
+            now.add( Calendar.MINUTE, 5 );
```
leleuj: Why add 5 minutes for now?

leleuj:
I think the implementation is really smart: to make the expiration a general concept of the user profile is a great idea!
I have one small comment about the source code. And the documentation should be updated as well: https://github.com/pac4j/pac4j/blob/master/documentation/docs/user-profile.md

pac4j-oidc/src/main/java/org/pac4j/oidc/profile/azuread/AzureAdProfile.java
mores: Use case when there is no buffer time.....
   id_token expires 10:00:00
current time is 9:59:59
   id_token is passed to the web app
   web app passes it_token to AzureAd 10:00:01 - sorry token has expired call fails.

leleuj: OK. I see, but isn't 5 minutes a bit too much? 10 seconds maybe?

mores: Ideally this value should be configured by the implementer.
I agree 5 minutes might be too long, I figured better safe than sorry. The end result is that my id_token gets refreshed every 55 minutes instead of every hour.
Depending on load/activity or other logic processes that may be written I feel 10 seconds may not be long enough for ALL cases.

leleuj: OK. I understand. As we have an `AzureAdProfileDefinition`, I think it should hold this delay and use it when building the `AzureAdProfile`...

mores: In a previous PR ( https://github.com/pac4j/pac4j/pull/1047 ) I was setting the config in the profile. It was not needed for that specific issue and was left out. Should that change be brought into this one, or is there a better way to bring in the delay configuration ?

leleuj: I don't think we should put the config in the profile. It should be the responsibility of the profile definition to know that and pass it when building the profile...

leleuj:
We're almost good here. Just update the documentation and we'll be ready for the merge. Thanks.

dmitriyvolk:
Can you please tell me how the refresh process is kicked off? All I can see is that a profile with an expired Id token gets ignored when read from session. What happens then? Is pac4j actually refreshes the token or just logs the user out so that they are forced to go through the login process and bring back an updated token?

leleuj:
You get it right: the profile is removed when read if it is expired. Then, no profile is returned in the security filter/logic which should trigger a new login process.

Issue #1070, opened 2017-12-20
pac4j incompatible with javax.faces.partial.ajax


mores:
Today during an ajax request, pac4j is returning: 
[AzureAdClient:86] () AJAX request detected -> returning 401
This will not work for JSF application ( I use primefaces. )
This same issue was brought up and fixed in CAS: https://issues.jasig.org/browse/CASC-206

PR #1073, opened 2017-12-26
Gist needed to address issue 1070

mores:
Here is one way to fix issue 1070. Should I continue will all the other code to make this work, or is there a different approach you would like to see ?

https://github.com/pac4j/pac4j/issues/1070

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
@@ -80,8 +80,10 @@ public abstract class IndirectClient<C extends Credentials, U extends CommonProf
      */
     public RedirectAction getRedirectAction(final WebContext context) {
         init();
+        if(ajaxRequestResolver.isFacesPartialAjax(context)) {
+            logger.info("Faces Partial AJAX request detected");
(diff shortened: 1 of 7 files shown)


leleuj:
I'd like to see something more generic as I think the AJAX logic should be encapsulated inside the `AjaxRequestResolver`. So to give you an idea of what I have in mind, I would see something like:

```java
if (ajaxRequestResolver.isAjax(context)) {
      logger.info("AJAX request detected -> returning the appropriate action");
      RedirectAction action = redirectActionBuilder.redirect(context);
      cleanRequestedUrl(context);
      throw ajaxRequestResolver.buildAjaxResponse(action);
}
````

mores:
If the request isAjax, buildAjaxResponse will need to either:
throw HttpAction ( ajax faces redirect )
or 
return RedirectAction ( non faces ajax redirect )
......
Are you saying that RedirectAction should now extend TechnicalException to bring it into alignment with HttpAction ?

leleuj:
Let's return a `RedirectAction` in the regular use case and throw an `HttpAction` for AJAX Faces requests, it will be properly handled in the `DefaultSecurityLogic`.

PR #1079, opened 2018-01-02
Default clients list in the Config

### LINKED ISSUE #987: Default clients list in the Config
Currently with pac4j, in the `Config` object, one can define a default `Client` object.

It could/would make sense to be able, instead, to pass a default clients' name list, in the same format that what is passed to the `SecurityLogic` object.

I already implemented it in jax-rs-pac4j at the annotation/filter level because I found it useful, and some people are using it too (see pac4j/jax-rs-pac4j#28) so it could be a desired feature to generalise to pac4j itself.

leleuj:
Closes #987 

cc @victornoel

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/client/finder/DefaultCallbackClientFinder.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/finder/DefaultCallbackClientFinder.java
@@ -34,7 +34,7 @@ public class DefaultCallbackClientFinder implements ClientFinder {
         }
 
         // fallback: we didn't find any client on the URL and have a default client
-        if (result.size() == 0 && CommonHelper.isNotBlank(clientNames)) {
+        if (result.isEmpty() && CommonHelper.isNotBlank(clientNames)) {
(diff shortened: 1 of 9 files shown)


pac4j-core/src/main/java/org/pac4j/core/client/finder/DefaultCallbackClientFinder.java
```diff
@@ -40,23 +33,11 @@ public DefaultCallbackClientFinder(final IndirectClient defaultClient) {
             }
         }
 
-        if (result.size() == 0 && defaultClient != null) {
-            return Arrays.asList(defaultClient);
-        } else {
-            return result;
+        // fallback: we didn't find any client on the URL and have a default client
+        if (result.size() == 0 && CommonHelper.isNotBlank(clientNames)) {
```
victornoel: prefer isEmpty to test emptyness :)

leleuj: Sure. I changed it.

Issue #1075, opened 2017-12-28
In the Profile, non-character types of fields will be lost


cabbage89:
Unable to specify field type，passed in null value
https://github.com/pac4j/pac4j/blob/b2766f1d9286fe3f3e2216ba98d6a31eb70dfd5d/pac4j-oauth/src/main/java/org/pac4j/oauth/client/GenericOAuth20Client.java#L50

The default is StringConverter

https://github.com/pac4j/pac4j/blob/1cfcf75ce128ca5851e70b9f199987f1ce0d0660/pac4j-oauth/src/main/java/org/pac4j/oauth/profile/generic/GenericOAuth20ProfileDefinition.java#L105

https://github.com/pac4j/pac4j/blob/ddc4fd7d08bace1608dcf40e2d1acc6ad163cd7b/pac4j-core/src/main/java/org/pac4j/core/profile/converter/StringConverter.java

The StringConverter's internalConvert method inherits from the abstract parent class, which by default returns null

https://github.com/pac4j/pac4j/blob/79642ab995439c7a71495ae56e56ecfba8fc8eda/pac4j-core/src/main/java/org/pac4j/core/profile/converter/AbstractAttributeConverter.java#L45

So lost the value of non-character type field！

leleuj:
The idea of the `StringConverter` is to return a String value for attributes which are strings. Here we change the behavior to always return a String whatever the attribute type, so you'll end up with `MyType@45589` and other outputs like that.
We should find another way to being able to set the appropriate converter per profile attribute.

cabbage89:
Yeah,We should find another way to being able to set the appropriate converter per profile attribute.
For example, the type specified by the following expression:
`String|name`
Do not specify the type with the following expression:
`name`

We can use chain-of-responsibility pattern here，Create a chain of responsibility for all your converters：
https://github.com/pac4j/pac4j/blob/b2766f1d9286fe3f3e2216ba98d6a31eb70dfd5d/pac4j-oauth/src/main/java/org/pac4j/oauth/client/GenericOAuth20Client.java#L50

leleuj:
Yes, we could try something like that...

PR #1080, opened 2018-01-02
 fix(#1075) set the appropriate converter per profile attribute

cabbage89:
fix(#1075) set the appropriate converter per profile attribute

Compatible with the previous configuration


cabbage89:
update #1084

PR #1084, opened 2018-01-03
fix(#1075) set the appropriate converter per profile attribute

cabbage89:
（#1075）Compatible with the previous configuration

### DIFF
--- a/pac4j-oauth/src/main/java/org/pac4j/oauth/client/GenericOAuth20Client.java
+++ b/pac4j-oauth/src/main/java/org/pac4j/oauth/client/GenericOAuth20Client.java
@@ -1,9 +1,16 @@
 package org.pac4j.oauth.client;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
 import java.util.Map;
+import java.util.Vector;
 import java.util.logging.Logger;
+import java.util.stream.Stream;
 
 import com.github.scribejava.core.model.Verb;
+import org.pac4j.core.profile.converter.AbstractAttributeConverter;
 import org.pac4j.oauth.profile.OAuth20Profile;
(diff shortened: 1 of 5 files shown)


pac4j-oauth/src/main/java/org/pac4j/oauth/client/GenericOAuth20Client.java
```diff
@@ -56,6 +73,49 @@ protected void clientInit() {
...
+        if (converterClasses == null) {
+            try {
+                ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+                Class cla = classLoader.getClass();
+                while (cla != ClassLoader.class) {
+                    cla = cla.getSuperclass();
+                }
+                Field field = cla.getDeclaredField("classes");
+                field.setAccessible(true);
+                Vector<Class> classes = (Vector<Class>) field.get(classLoader);
```
leleuj: I don't like the idea of browsing all classes to find the right we want: what about the performance?

Cannot we just initialize a `Map`  with the existing converters instead of using reflection?

cabbage89: There is a cached object "converterClasses" in the GenericOAuth20Client instance that reflects only once,
If we use `Map`, that would be inflexible, and in the future, when we added the converter, we also need to add it here, and maybe we will miss it.

leleuj: Yes, I understand. Can you update the documentation then? -> https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/oauth.md

cabbage89: OK, I'll do it

cabbage89:
Has updated the document

leleuj:
Thanks

Issue #1085, opened 2018-01-03
[core] use of shiro's noSessionCreation raises DisabledSessionException


andre-doherty:
I have a rest backend  which is secured using pac4j saml2 indirectclient and headerclient directclient.
It has to be stateless so to comply with that i have used the shiro's dedicated filter (noSessionCreation).

It appears that some unnecessary session creation are done in several places (mainly by setting blank values for some session attributes).
I have focused on *logic classes and also in SAML2Client

There is also a hardcoded value in defaultcallbacklogic (saveUserProfile) that would require the creation of a supplementory parameter in callbackfilter (eg saveInSession)

I have already worked on a fix applying to the 2.2.x branch

leleuj:
Regarding the session, things are not easy: I'll see your PR...

Please start by targeting the master branch and then backport to the 2.2.x branch.

PR #1089, opened 2018-01-06
fix unnecessary session creation report in 2.2.x

andre-doherty:
Please find the pull request for issue #1085 

I am also working on pac4j 3.0.0 however it seems to be incompatible with the current bujiio master (eg HttpActionAdapter is not anymore in the http subpackage but in http.adapter subpackage which is not what Callbackfilter is waiting for

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
@@ -78,7 +78,7 @@ public class HeaderExtractor implements CredentialsExtractor<TokenCredentials> {
             throw new CredentialsException("Wrong prefix for header: " + this.headerName);
         }
 
-        String headerWithoutPrefix = header.substring(this.prefixHeader.length()).trim();
+        String headerWithoutPrefix = header.substring(this.prefixHeader.length());
 
         if (trimValue) {
             headerWithoutPrefix = headerWithoutPrefix.trim();
--- a/pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
@@ -36,7 +36,7 @@ public class HeaderExtractor implements CredentialsExtractor<TokenCredentials> {
             throw new CredentialsException("Wrong prefix for header: " + this.headerName);
         }
 
-        final String headerWithoutPrefix = header.substring(this.prefixHeader.length());
+        final String headerWithoutPrefix = header.substring(this.prefixHeader.length()).trim();
         return new TokenCredentials(headerWithoutPrefix, clientName);
     }
 }
--- a/pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
@@ -20,7 +20,7 @@ public class HeaderExtractor implements CredentialsExtractor<TokenCredentials> {
 
     private String clientName;
 
-    private Boolean trimValue;
+    private boolean trimValue = false;
 
     public String getHeaderName() {
         return headerName;
@@ -80,7 +80,7 @@ public class HeaderExtractor implements CredentialsExtractor<TokenCredentials> {
 
         String headerWithoutPrefix = header.substring(this.prefixHeader.length()).trim();
 
-        if (trimValue != null && trimValue) {
+        if (trimValue) {
             headerWithoutPrefix = headerWithoutPrefix.trim();
         }
         return new TokenCredentials(headerWithoutPrefix, clientName);
--- a/pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
@@ -4,6 +4,7 @@ import org.pac4j.core.context.WebContext;
 import org.pac4j.core.exception.CredentialsException;
 import org.pac4j.core.credentials.TokenCredentials;
 import org.pac4j.core.exception.HttpAction;
+import org.pac4j.core.util.CommonHelper;
 
 /**
  * To extract header value.
@@ -13,11 +14,49 @@ import org.pac4j.core.exception.HttpAction;
  */
 public class HeaderExtractor implements CredentialsExtractor<TokenCredentials> {
 
-    private final String headerName;
+    private String headerName;
 
-    private final String prefixHeader;
+    private String prefixHeader;
 
-    private final String clientName;
+    private String clientName;
+
+    private Boolean trimValue;
+
+    public String getHeaderName() {
+        return headerName;
+    }
+
+    public void setHeaderName(String headerName) {
+        this.headerName = headerName;
+    }
+
+    public String getPrefixHeader() {
+        return prefixHeader;
+    }
+
+    public void setPrefixHeader(String prefixHeader) {
+        this.prefixHeader = prefixHeader;
+    }
+
+    public String getClientName() {
+        return clientName;
+    }
+
+    public void setClientName(String clientName) {
+        this.clientName = clientName;
+    }
+
+    public Boolean getTrimValue() {
+        return trimValue;
+    }
+
+    public void setTrimValue(Boolean trimValue) {
+        this.trimValue = trimValue;
+    }
+
+    public HeaderExtractor() {
+        // empty constructor as needed to be instanciated by beanutils
+    }
 
     public HeaderExtractor(final String headerName, final String prefixHeader, final String clientName) {
         this.headerName = headerName;
@@ -27,6 +66,9 @@ public class HeaderExtractor implements CredentialsExtractor<TokenCredentials> {
 
     @Override
     public TokenCredentials extract(WebContext context) throws HttpAction, CredentialsException {
+        CommonHelper.assertNotBlank("headerName", this.headerName);
+        CommonHelper.assertNotNull("prefixHeader", this.prefixHeader);
+
         final String header = context.getRequestHeader(this.headerName);
         if (header == null) {
             return null;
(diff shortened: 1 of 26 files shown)


pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
```diff
@@ -36,7 +36,7 @@ public TokenCredentials extract(WebContext context) throws HttpAction, Credentia
             throw new CredentialsException("Wrong prefix for header: " + this.headerName);
         }
 
-        final String headerWithoutPrefix = header.substring(this.prefixHeader.length());
+        final String headerWithoutPrefix = header.substring(this.prefixHeader.length()).trim();
```
leleuj: As I said in another comment, let's make the trimming an option driven by a boolean property, something like `trimValue`.

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java
```diff
@@ -47,6 +47,12 @@ public R perform(final C context, final Config config, final HttpActionAdapter<R
         } else {
             defaultUrl = inputDefaultUrl;
         }
+        final boolean saveInSession;
+        if (inputSaveInSession == null) {
```
leleuj: I'm a bit skeptical about the use case which justifies not to save the user profile into the web session after a remote login process: can you refresh my memory about your use case?

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
```diff
@@ -145,7 +149,9 @@ public R perform(final C context, final Config config, final SecurityGrantedAcce
                 } else {
                     if (startAuthentication(context, currentClients)) {
                         logger.debug("Starting authentication");
-                        saveRequestedUrl(context, currentClients);
+                        if (!ajaxRequestResolver.isAjax(context)) {
```
leleuj: I'd prefer to have this logic `if (!ajaxRequestResolver.isAjax(context)) {` inside the `saveRequestedUrl` to make this overridable.

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java
andre-doherty: Well the use case is the following, a spa relying on a jaxws backend request authent on a saml2 idp.
Upon auth completion the idp will trigger callback filter by the mean of the browser.
The callbackfilter will then produce a samlprofile and place it in the request context. 
This profile will be changed into a jwt token and sent back to the spa client.
In this use case as the final callback is processed in one single call there is no need for a session creation. 
The spa will authenticate itself using the jwt token for further requests.

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
andre-doherty: Sure, will make that change.

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java
leleuj: OK, it's a bit unusual, but it makes sense.

pac4j-core/src/main/java/org/pac4j/core/credentials/extractor/HeaderExtractor.java
```diff
@@ -13,11 +14,49 @@
...
-    private final String headerName;
+    private String headerName;
 
-    private final String prefixHeader;
+    private String prefixHeader;
 
-    private final String clientName;
+    private String clientName;
+
+    private Boolean trimValue;
```
leleuj: Just a detail, but I'd prefer to have a `boolean` here...

pac4j-http/src/main/java/org/pac4j/http/client/direct/HeaderClient.java
```diff
@@ -52,10 +52,12 @@ public HeaderClient(final String headerName, final String prefixHeader,
 
     @Override
     protected void clientInit(final WebContext context) {
-        CommonHelper.assertNotBlank("headerName", this.headerName);
-        CommonHelper.assertNotNull("prefixHeader", this.prefixHeader);
+        if (getCredentialsExtractor() == null) {
```
leleuj: In fact, you don't need that test as the `defaultCredentialsExtractor` already does that...

leleuj:
Two new small changes and then, we're good. You can also prepare a similar PR for the master. Thanks.

leleuj:
I realized this change is a breaking one, so we can't accept it for the 2.2.x branch. Though, we will accept it for the master branch. Thanks

leleuj:
Please open the same PR for the master and change this one to make it backward compatible with the 2.2.x branch (turn the `saveInProfile` input parameter into a property to keep the method signature and maybe some other adjustments). Thanks

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java
```diff
@@ -81,7 +93,7 @@ public R perform(final C context, final Config config, final HttpActionAdapter<R
 
             final CommonProfile profile = client.getUserProfile(credentials, context);
             logger.debug("profile: {}", profile);
-            saveUserProfile(context, config, profile, multiProfile, renewSession);
+            saveUserProfile(context, config, profile, saveInSession, multiProfile, renewSession);
```
leleuj: Can we avoid passing the `saveInSession` property here to keep the current signature?

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java
```diff
@@ -93,10 +105,10 @@ public R perform(final C context, final Config config, final HttpActionAdapter<R
     }
 
     protected void saveUserProfile(final C context, final Config config, final CommonProfile profile,
-                                   final boolean multiProfile, final boolean renewSession) {
+                                   final boolean saveInSession, final boolean multiProfile, final boolean renewSession) {
         final ProfileManager manager = getProfileManager(context, config);
         if (profile != null) {
-            manager.save(true, profile, multiProfile);
+            manager.save(saveInSession, profile, multiProfile);
```
leleuj: Same comment here: can we directly use the `saveInSession` property and no input parameter?

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
```diff
@@ -233,9 +237,11 @@ protected boolean startAuthentication(final C context, final List<Client> curren
      * @throws HttpAction whether an additional HTTP action is required
      */
     protected void saveRequestedUrl(final C context, final List<Client> currentClients) throws HttpAction {
-        final String requestedUrl = context.getFullRequestURL();
-        logger.debug("requestedUrl: {}", requestedUrl);
-        context.setSessionAttribute(Pac4jConstants.REQUESTED_URL, requestedUrl);
+        if (!ajaxRequestResolver.isAjax(context)) {
```
leleuj: This is also a breaking change at runtime. Can we avoid it?

leleuj:
@andre-doherty I added a few more comments to address.
I'm aware the process is a bit though on this PR, but I want to be sure we don't break anything as it's our commitment.

Regarding the change in Shiro, I just released buji-pac4j v3.1.0 so it should go into a new version 3.2.0.
As the right way to inject a new profile manager, you should rely on the `profileManagerFactory` property of the `Config` component and use that in the callback filter.

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java
andre-doherty: Sure, should have done it.

pac4j-core/src/main/java/org/pac4j/core/engine/DefaultCallbackLogic.java
andre-doherty: OK

andre-doherty:
Well yes, i still had one concern about the way to setup the profileManagerFactory.
As i didn't found a way to specify a lambda (ie : setProfileManagerFactory(ShiroProfileManager::new)) in shiro.ini (beanutils doens't seem to support the jdk8 and its new features), i was still wondering how to setup profileManagerFactory.

I might have to add another property (eg useShiroProfileManager) in CallbackFilter to trigger the use of shiroProfileManager in case a custom CallbackLogic implements ProfileManagerFactoryAware).

It would not break the compatibily but on the other side it would be a little too much "hardcoded" (what if one want to use a different ProfileManager).
I am still thinking about that

leleuj:
Please account for my comments on this PR and open a specific one for the master.

Regarding Shiro, my feeling is that the `ini` format is somehow too limited and sometimes, it's better to create a new class than try to define settings via the `ini` format.

andre-doherty:
Hello, 

Please see my latest commits. Sorry it took so long, i have been quite busy lately.
All the changes in api should be gone now, so it should be fully compatible with 2.2.x branch

The init of CallbackLogic and SecurityLogic (especially the need to set the ShiroProfileManager::new) have been addressed using two Custom classes where there only purpose is a clean setup of the above member. (Wish i had not to do it)
I havent added those to the buji lib as it was not quite clear if those belong to it, actually there is this strong dependency on buji but there natures should put those into the pac4j-core.
Please let me know if you believe they still can be added to buji. 
Note the init of the SecurityFilter and CallbackFilter would be much more cleaner than now.

Thank you

leleuj:
I'm in line with the changes in the PR.
I'm not sure to completely follow you regarding buji-pac4j, so please submit a PR to make things clearer.

With this change on the 2.x branch, we should upgrade to version 2.3.0 (per our versions policy), so please update the version.
And submit the related PR to the master as I don't want to be behind this feature on master.

andre-doherty:
Hi, 
I have created the 2 following pull requests : 
- one targeting the 3.0.0 pac4j branch including this change transposed into 3.0.0 branch
- one targeting the 3.2.0 buji branch (hope this pr make the above idea more explanatory).

I have also pushed some more commits on this PR (pom from 2.2.x=>2.3.0 as requested)

leleuj:
We're good on this one. Merging. Then, I'll make some tests before cutting a release...

I have one minor comment for the PR targeting the master branch. Please have a look.

I'll review shortly the PR for Shiro.

andre-doherty:
Hello,

Did you have the time to make the tests ? 
A release would be very much appreciated as we have apps relying on this branch going live by the end of the month.
Thank you very much

leleuj:
I just checked the 2.3.0-SNAPSHOT branch and it seems to be working correctly. So the release is in progress... The 2.3.0 release will show up in the Maven central repository in a few hours...

andre-doherty:
Thanks a lot!

Le ven. 16 mars 2018 à 16:39, LELEU Jérôme <notifications@github.com> a
écrit :

PR #1090, opened 2018-01-07
Alternate way to suppport javax.faces.partial.ajax

mores:
Address issue: https://github.com/pac4j/pac4j/issues/1070


victornoel:
this seems awfully specific to be in a generic class... don't you think you should introduce a specific `AjaxRequestResolver` for this use case instead of handling it in the default one?

victornoel:
(this is just an opinion, @leleuj will most certainly decide what is best :)

leleuj:
@victornoel I see your point, but I'm in line with the proposed implementation. In most cases, things should work out of the box: so if you use a JSF app, AJAX requests should be handled properly, like it is in this `DefaultAjaxResolver`. Overriding this component should only be reserved to advanced use cases.

PR #1093, opened 2018-01-16
Fix logger in Pac4jHTTPRedirectDeflateEncoder

jsimomaa:
Fixing logger in Pac4jHTTPRedirectDeflateEncoder to have logging done with correct class


leleuj:
Good catch!

PR #1094, opened 2018-01-17
Decouple the CallbackUrlResolver from the UrlResolver

### LINKED ISSUE #1024: Decouple the CallbackUrlResolver from the UrlResolver

leleuj:
cc @victornoel 

Closes #1024


victornoel:
Cool, it was one of the last missing pieces to update the jax-rs and dw implementations.

If I understand well, it has to be set on the client itself, it's not in the Config object, right?
Any reason not to do so? I don't think it's going to be a problem but I would I thought that maybe a user would want to set it globally?

leleuj:
I added a `UrlResolver` at the `Clients` level, to share it with all clients.

victornoel:
sorry I couldn't review earlier, anyway this last commit is perfectly answering my concern :)

PR #1095, opened 2018-01-22
Ensure private key password is used when generating saml metadata

mmoayyed:
The present method of generating a keystore for SAML2 uses the keystore password, rather than the private key password, which is of course a problem.


leleuj:
Just to be sure: we were using the password of the keystore for the certificate instead of a dedicated password, right?

mmoayyed:
Yes, technically it's used for the private key; not the certificate...which is wrong.

mmoayyed:
We should probably backport this to the 2.x branches as well.

leleuj:
Yes, it makes sense. I just backported it.

PR #1099, opened 2018-02-01
Allow optional assertion conditions

elektro-wolle:
As approved by Misagh on pac4j-users, conditions are optional.


leleuj:
Thanks for your contribution.

PR #1101, opened 2018-02-02
Remove use of unmodifiable maps

mmoayyed:
Ran into a problem with profile attributes constructed and returned as unmodifiable maps. If the profile is put inside a cache such as Memcached, the unmodifiable map instance cannot be reconstructed when it's fetched from the cache and populated.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
@@ -183,14 +183,14 @@ public abstract class UserProfile implements Serializable, Externalizable {
         return getImmutableAttributeMap(this.authenticationAttributes);
     }
 
-    private Map<String, Object> getImmutableAttributeMap(Map<String, Object> attributeMap) {
+    private static Map<String, Object> getImmutableAttributeMap(Map<String, Object> attributeMap) {
         final Map<String, Object> newAttributes = new HashMap<>();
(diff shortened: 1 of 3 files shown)


pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
```diff
@@ -183,14 +183,14 @@ public String getTypedId() {
         return getImmutableAttributeMap(this.authenticationAttributes);
     }
 
-    private Map<String, Object> getImmutableAttributeMap(Map<String, Object> attributeMap) {
+    private static Map<String, Object> getImmutableAttributeMap(Map<String, Object> attributeMap) {
```
leleuj: OK. So let's rename this method as `getAttributeMap` then.

leleuj:
One minor comment. Are you sure the immutable collections are properly registered in the Kyro serializer?

mmoayyed:
Yes, immutables are properly registered with Kryo. It seems like because they are registered, the immutable class is written to the cache. Then when it's read again, Kryp creates an instance of the same immutable structure and attempts to put things into it, which obviously fails. 

Immutables also cause problems if the structure is about to be serialized to JSON for instance. Same problem.

leleuj:
OK. So please rename the `getImmutableAttributeMap` method before I merge.

leleuj:
And the build must be fixed as well (I guess some tests expected immutable collections).

leleuj:
Thanks

PR #1104, opened 2018-02-15
Add UserProfile.setRoles and setPermissions for convenience

zenkiezhu:
So that the CommonProfile subclass can set new roles and permissions when admins change that


leleuj:
Indeed, we don't have the appropriate setters for the roles and permissions.

Can you make the methods `public`? Thx

zenkiezhu:
Sure, done

leleuj:
Thanks

PR #1107, opened 2018-02-19
Fix unnecessary session creation in 3.0

andre-doherty:
Please find the pull request following the change on branch 2.2.x
( [https://github.com/pac4j/pac4j/pull/1089](https://github.com/pac4j/pac4j/pull/1089) )


pac4j-core/src/main/java/org/pac4j/core/engine/DefaultSecurityLogic.java
```diff
@@ -53,6 +55,8 @@
 
     private ProfileStorageDecision profileStorageDecision = new DefaultProfileStorageDecision();
 
+    private AjaxRequestResolver ajaxRequestResolver = new DefaultAjaxRequestResolver();
```
leleuj: Can you add the setter/getter for this property?

Issue #1103, opened 2018-02-07
Check how authentication attributes are restored in addition to user attributes


leleuj:
In the `ProfileHelper` component...

jkacer:
Hi Jerome, I have create pull request #1110 for this issue. It's for branch 2.2.x. If you find it OK, I can add another one for master later.

I just noticed one thing: The 2.2.x branch now contains version 2.3.0-SNAPSHOT, so we should probably create a new branch after the merge. Or create it before and then target the PR against the new branch. I'll leave it up to you...

leleuj:
The PR is good, see my minor comments to address. Please open a new PR for the master before I merge anything (you don't need to make things backward-compatible for the master branch).

Yes, the 2.2.x is now at 2.3.0-SNAPSHOT as there is already a new backward-compatible change in it. I think I'll cut a 2.3.0 release after I merge and test your changes.

jkacer:
Hi! Thank you for the review, I'll have a look and fix the remaining issues, then make a new PR for master.

jkacer:
Hi Jerome, here is the pull request for master: https://github.com/pac4j/pac4j/pull/1112.
It contains the same changes, ported from 2.2.x, plus removal of the 3 deprecated methods and adequate adjustments in clients (one more parameter).

PR #1112, opened 2018-03-01
#1103: Check how authentication attributes are restored... [master branch]

jkacer:
This is a pull request to solve issue #1103 for the master branch - port of changes from branch 2.2.x.
I have enhanced the ProfileHelper and the ProfileDefinition with a few new methods that support authentication attributes.
Existing methods, deprecated in 2.2.x and 2.3.x, were removed from 3.0.x. Calls from clients and unit tests were adapted.

### DIFF
--- a/pac4j-cas/src/main/java/org/pac4j/cas/credentials/authenticator/CasAuthenticator.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/credentials/authenticator/CasAuthenticator.java
@@ -90,9 +90,9 @@ public class CasAuthenticator extends ProfileDefinitionAware<CommonProfile> impl
             if (configuration.getProxyReceptor() != null) {
                 profile = getProfileDefinition().newProfile(principal, configuration.getProxyReceptor());
                 profile.setId(ProfileHelper.sanitizeIdentifier(profile, id));
-                getProfileDefinition().convertAndAdd(profile, newAttributes);
+                getProfileDefinition().convertAndAdd(profile, newAttributes, null);
             } else {
-                profile = ProfileHelper.restoreOrBuildProfile(getProfileDefinition(), id, newAttributes, principal,
+                profile = ProfileHelper.restoreOrBuildProfile(getProfileDefinition(), id, newAttributes, null, principal,
                     configuration.getProxyReceptor());
             }
             logger.debug("profile returned by CAS: {}", profile);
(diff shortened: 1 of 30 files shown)


jkacer:
I'm not sure if I should care about the warnings from Code Climate, please tell me what you think. Thanks.

pac4j-cas/src/main/java/org/pac4j/cas/credentials/authenticator/CasAuthenticator.java
```diff
@@ -90,9 +90,9 @@ public void validate(final TokenCredentials credentials, final WebContext contex
             if (configuration.getProxyReceptor() != null) {
                 profile = getProfileDefinition().newProfile(principal, configuration.getProxyReceptor());
                 profile.setId(ProfileHelper.sanitizeIdentifier(profile, id));
-                getProfileDefinition().convertAndAdd(profile, newAttributes);
+                getProfileDefinition().convertAndAdd(profile, newAttributes, null);
```
leleuj: Instead of `null`, we should use the authentication attributes retrieved via `assertion.getAttributes()`.

leleuj:
You can generally ignore the warnings from code climate, or at least that's what I do most of the time ;-)

One minor comment and a question and we'll be good.

pac4j-cas/src/main/java/org/pac4j/cas/credentials/authenticator/CasAuthenticator.java
jkacer: Aha, OK, I'll have a look. I generally adapted all the calls by using profile attributes only to make it compile and act exactly like now (without authentication attributes). But you are right, we can improve it by using the authentication attributes where possible, since they are already available.

jkacer:
Hi! I have modified `CasAuthenticator` to count with authentication attributes as you suggested. I was not sure if they need the undergo the same transformation by the attribute handler but I decided to pass them through it too.
Concerning other possible uses, I was not able to find anything else, including SAML. In `SAML2Authenticator.validate()`, we operate with the attributes a lot but all attributes from `SAML2Credentials` are implicitly considered as profile attributes; authentication attributes are explicitly enumerated and added one-by-one. But if you are able to find a concrete call that could be updated, please tell me.
Would you like to have the fix of CAS Authenticator in 2.2.x too?

leleuj:
OK. Thanks for checking.

Regarding the 2.2.x branch, do it if you find some time, but it's not a priority.

leleuj:
Thank you for contributing!

PR #1113, opened 2018-03-06
Introduce specific saml exceptions - add UTC to date comparisons

mmoayyed:
This pull requests:

- Tries to handle date comparisons based on a common timezone (UTC) when SAML responses are validated.
- Introduces specific exceptions for SAML validation events so that integrations can catch each event separately and re-act to it differently (UI, etc).

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/exceptions/SAMAssertionSubjectException.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/exceptions/SAMAssertionSubjectException.java
@@ -1,0 +1,20 @@
+package org.pac4j.saml.exceptions;
+
+/**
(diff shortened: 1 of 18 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/exceptions/SAMAssertionSubjectException.java
```diff
@@ -0,0 +1,20 @@
+package org.pac4j.saml.exceptions;
+
+/**
+ * This is {@link SAMAssertionSubjectException}.
+ *
+ * @author Misagh Moayyed
+ */
```
leleuj: Can you add the `@since 3.0.0` annotation so we can track the change?

leleuj:
Thanks

PR #1114, opened 2018-03-07
Allow oidc configuration to indicate state predefined

mmoayyed:
Similar to the OAuth20Configuration, building the redirection url for OIDC client should have the ability to specify state where needed.


mmoayyed:
Are we all good on this one? Shall we proceed?

leleuj:
Great!

PR #1115, opened 2018-03-08
Allow CasClient to use custom parameters when building callback urls

mmoayyed:
Callback urls for CAS need to be customized in certain cases to include additional parameters that might convey "state" like values. Rather than building a new custom resolver, this PR allows the existing resolver to support custom parameters.


leleuj:
Good improvement. Thanks

PR #1117, opened 2018-03-08
Ensure state object is remembered by session-store

mmoayyed:
Instead of storing state data, ensure the State object is tracked by session store.


leleuj:
Awesome!

PR #1118, opened 2018-03-08
Chain SAML exceptions when validating responses

mmoayyed:
When validating a SAML response, the exceptions thrown as part of validating individual assertions are suppressed and are never reported back. This pull request collects them all, and then attempts to throw the first back if one is found before moving on.


leleuj:
Excellent! Merging

PR #1119, opened 2018-03-13
Add support for BaseClient custom properties

mmoayyed:
During the bootstrapping and construction of a client, it would be useful to provide a "bag of properties" that one can use to attach settings to the client in order to determine behavior elsewhere. An example would be: *This client should automatically redirect to the provider when the view layer begins processing things, etc*

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
@@ -2,7 +2,9 @@ package org.pac4j.core.client;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
```diff
@@ -206,10 +211,18 @@ public void setProfileCreator(final ProfileCreator<C, U> profileCreator) {
         this.profileCreator = profileCreator;
     }
 
+    public Map<String, Object> getCustomProperties() {
+        return customProperties;
+    }
+
+    public void setCustomProperties(final Map<String, Object> customProperties) {
+        this.customProperties = customProperties;
```
leleuj: Can we just check that the `customProperties` is not `null`?

mmoayyed: Sure. Done.

PR #1121, opened 2018-03-18
Callback save in session default value (2.x)

victornoel:
Hi,

I noticed a discrepancy in the way defaults where handled for the `saveInSession` parameter of `CallbackLogic`.

The way it was before this PR is that a NPE would be emitted if `null` was called as the parameter of `setSaveInSession` and thus it wasn't possible to set the default value like this.

I also updated pmd and findbugs dependencies because they are not working with latest versions of eclipse (I already made a PR that was merged for the master branch in the past).


leleuj:
The release of the version 2.3.1 is in progress...

PR #1127, opened 2018-04-11
Support for European e-identification (merging with 2.2.0)

luis100:
Integration with the [European eId](https://ec.europa.eu/digital-single-market/en/e-identification), specifically with [Portuguese implementation](https://www.autenticacao.gov.pt/) that use the  SAML2 protocol

Changes include:
- [x] Allow to restrict signing algorithms to rsa-sha1 and sha1 (bug)
- [x] Support configuring the ProviderName (previously hardcoded to 'pac4j-saml')
- [x] Support defining AuthnRequest extensions (used to request specific attributes on each authentication request)
- [x] Guarding against a NPE when AuthContext is defined AuthContextClassRef child is not (this element is not mandatory on the SAML specification, and does not appear in the use case)
- [x] Adding support for setting a SAML attribute as the principal Id: this is necessary as the eId specification defines that the SAML NameId is not used to convey the user identification, an attribute is used instead (to note that many possible user identifications may be on the attributes, as the fiscal number or the social security number)

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
@@ -112,7 +112,7 @@ public class SAML2Client extends IndirectClient<SAML2Credentials, SAML2Profile> 
             credentials.setClientName(getName());
             return credentials;
         });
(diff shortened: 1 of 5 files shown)


leleuj:
Thanks for this great contribution.

PR #1129, opened 2018-04-18
Add twitter include_email support

sbearcsiro:
Support the twitter include_email parameter documented here: https://developer.twitter.com/en/docs/accounts-and-users/manage-account-settings/api-reference/get-account-verify_credentials


leleuj:
Thanks

PR #1131, opened 2018-05-03
Allow JWTs to be generated/authenticated with expiration time

mmoayyed:
Title says all.


leleuj:
Thanks for the contribution. The build fails with findbugs errors.

pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
```diff
@@ -300,6 +306,10 @@ public void setRealmName(final String realmName) {
         this.realmName = realmName;
     }
 
+    public void setExpirationTime(final Date expirationTime) {
```
leleuj: The `getExpirationTime` getter is missing, isn't it?

leleuj:
I'd like to cut the RC2 today. Do you think you'll be able to make the requested changes on this PR today? Thanks

Issue #1074, opened 2017-12-27
Restore originally requested URL using all HTTP methods: POST, HEAD... and not only GET


S66D:
Hi, I am currently using pac4j with Jooby, and times there are some login issues when redirecting to the requested URL after a successful login, where I have found out that you will be redirected to the last requested URL from the browser even if its an css.map file requested by a HEAD method so is there a way to distinguish between GET and HEAD methods.

We could set LAST_REQUESTED_URL and LAST_GET_REQUESTED_URL if this is a standard use case.

leleuj:
The question has never been raised, but I think there is some improvement in the way the request is saved and restored (your use case, but what about POST requests also?)

Currently, you can override the `saveRequestedUrl` method of the `DefaultSecurityLogic` component.

S66D:
Thanks for your interest, and sorry for the late reply,
I'm not so sure about POST request, but what I know is if a POST request is being sent and the client is not authenticated the client should be redirected to the login page and after the client logged-in he should resend the request again by visiting the page or the location where he sent the POST request unless if `pac4j ` have some kind of cache for POST requests body, even so I think it should be server side business, also `pac4j ` could gave some interfaces on implementing cached POST requests.

best of luck.

S66D:
I have found out that it is even more reasonable and convenient if there was a condition for last requested url to be type of html mime by the requested `Accept` header

aruis:
I got the same issues when i use a cas server. And I think use 307 http status instead of 302 will be ok.
so I suggest `org.pac4j.core.context.TEMP_REDIRECT = 307`

onnoweb:
I would like to check in on this issue and specifically POST requests. When my POST request will be authenticated by an indirect method (specifically SAML) then the body is stripped and not restored after successful authentication. 

Are there plans to address this?

I can, and probably will need to, address this in my own code but it seems something that pc4j should do out of the box.

Thoughts?

leleuj:
It has been fixed in pac4j v4.

onnoweb:
Thanks for the quick response. pac4j v4 was not been released yet? I see an RC1?

leleuj:
Indeed, a RC2 is in progress.

mukulchakane:
Hi , pac4j 4.0 is released or it is still in RC??

mmoayyed:
Still in RC.
*- Misagh*

Issue #866, opened 2017-03-22
Prefer 303 over 302 for redirecting after a POST on the callback URL


victornoel:
Hi,

This is more for discussion than a bug per-se, but according to the HTTP specification, after POST, it would be wiser to use [303](https://en.wikipedia.org/wiki/HTTP_303) to redirect than [302](https://en.wikipedia.org/wiki/HTTP_302).

From the [302](https://en.wikipedia.org/wiki/HTTP_302) page on wikipedia:
> Many web browsers implemented this code in a manner that violated this standard, changing the request type of the new request to GET, regardless of the type employed in the original request (e.g. POST).[1] For this reason, HTTP/1.1 (RFC 2616) added the new status codes 303 and 307 to disambiguate between the two behaviours, with 303 mandating the change of request type to GET, and 307 preserving the request type as originally sent. Despite the greater clarity provided by this disambiguation, the 302 code is still employed in web frameworks to preserve compatibility with browsers that do not implement the HTTP/1.1 specification.

See also https://blog.jayway.com/2012/10/17/what-you-may-not-know-about-http-redirects/ for example that explains the same.

leleuj:
Yes, a 303 HTTP code should be used when redirecting after a POST. Yet, we must think about it carefully because sometimes it's easier to use a well-known code even if it's not exactly the appropriate one (remember the logout 204 vs 200 HTTP code discussion).

Marked as "discussion" for version 3.0.0.

victornoel:
I remember the previous discussion :)

I was thinking that if we restricted this behaviour to HTTP/1.1, then the browser would understand a 303 (and thus that it could be integrated in v2 as with the 204 HTTP code)

victornoel:
FTR mainly the problem is with things like Apache HttpClient, Jersey Client and so on. They are usually more strict by default with regard to the specifications and don't follow 302 after POST for example.

victornoel:
Should we do this?

leleuj:
Yes, like the 204 HTTP code, we are at the beginning of the 3.0 development, so let's give it a try... We'll see if we eventually revert it or not.

leleuj:
@victornoel I strongly refactored the way HTTP actions are handled in v4. And I submitted the following PR regarding this issue: https://github.com/pac4j/pac4j/pull/1223

PR #1134, opened 2018-05-09
use 307 status when redirectToOriginallyRequestedUrl

aruis:
use 307 status when redirectToOriginallyRequestedUrl


aruis:
At first, I thought that using 307 status code instead of 302 would solve problems like https://github.com/pac4j/pac4j/issues/1074. But when I actually tested with cas, I did not succeed. Since the application server had already returned 302 before returning 307, the last time 307 was not executed as expected.
So this PR should have no practical significance.
thank you.

leleuj:
The concern has already been raised via this issue: https://github.com/pac4j/pac4j/issues/866

But things are not that easy. Despite 307 makes more sense, its limited adoption compared to 302 may generate more problems than it solves.

PR #1135, opened 2018-05-11
Generalizes OIDC session expiration when token expires

juanjovazquez:
Currently, this behavior is only available for the Azure OIDC client. This PR generalizes this making it configurable through two new secondary attributes: `ExpireSessionWithToken` and `TokenExpirationAdvance`.

Azure OIDC client users must take into account that this behavior is now optional in contrast with before that it wasn't.


leleuj:
It looks good to me, although I'd appreciate seeing a few tests to confirm the behavior.

leleuj:
Excellent! Thanks

PR #1136, opened 2018-05-11
Added a configurable way of removing setNameQualifier on Issuer. Usef…

EngOpenPlatform:
Added a configurable way of removing setNameQualifier on Issuer. Useful when using ADFS 3.0 that does not accept NameQualifier when using urn:oasis:names:tc:SAML:2.0:nameid-format:entity.

In SAML2ClientConfiguration you can use setUseNameQualifier to disable the NameQualifier from SAML Request. The flag default is true, in this way it will continue to work as before is Name Qualifier is required.


leleuj:
Great! Can you add some documentation in the SAML page: https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/saml.md? Thx

Issue #1138, opened 2018-05-16
Suboptimal default behavior choosing ClientAuthenticationMethod


haumacher:
If no `ClientAuthenticationMethod` is configured (`null`), the very first method that is suggested by the `OIDCProviderMetadata` is choosen. This is pretty clumsy, since only `client_secret_post` and `client_secret_basic` are supported. This causes the code to fail few lines later with a `TechnicalException("Unsupported client authentication method: " + chosenMethod)`:

https://github.com/pac4j/pac4j/blob/ad0072a4a40ae983980dfcb261b667534beb8790/pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/authenticator/OidcAuthenticator.java#L57-L74

Better would be to iterate the provider-supported methods an choose the first one that is supported by pac4j.

leleuj:
Yes, you're right. Would you mind submitting a PR for that? Thanks

haumacher:
Yes, I'll prepare one.

PR #1139, opened 2018-05-17
Better default for ClientAuthenticationMethod if none is configured

haumacher:
When no ClientAuthenticationMethod is configured, use the first
supported method found in the provider metadata (instead of the very
first one that is potentially not supported by the implementation).
Additionally, do not log a warning, if no ClientAuthenticationMethod is
configured and one is chosen from the provider metadata.

See #1138.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/authenticator/OidcAuthenticator.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/authenticator/OidcAuthenticator.java
@@ -61,20 +61,20 @@ public class OidcAuthenticator implements Authenticator<OidcCredentials> {
         final ClientAuthenticationMethod chosenMethod;
         if (CommonHelper.isNotEmpty(metadataMethods)) {
             if (preferredMethod != null) {
-            	if (metadataMethods.contains(preferredMethod)) {
-            		chosenMethod = preferredMethod;
-            	} else {
-            		throw new TechnicalException(
-	    				"Preferred authentication method (" + preferredMethod + ") not supported " + 
-	    				"by provider according to provider metadata (" + metadataMethods + ").");
-            	}
+                if (metadataMethods.contains(preferredMethod)) {
+                    chosenMethod = preferredMethod;
+                } else {
+                    throw new TechnicalException(
+                        "Preferred authentication method (" + preferredMethod + ") not supported " + 
+                        "by provider according to provider metadata (" + metadataMethods + ").");
+                }
             } else {
                 chosenMethod = firstSupportedMethod(metadataMethods);
             }
         } else {
-        	chosenMethod = preferredMethod != null ? preferredMethod : ClientAuthenticationMethod.getDefault();
-        	logger.info("Provider metadata does not provide Token endpoint authentication methods. Using: {}",
-        			chosenMethod);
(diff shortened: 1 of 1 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/authenticator/OidcAuthenticator.java
```diff
@@ -50,17 +58,40 @@ public OidcAuthenticator(final OidcConfiguration configuration, final OidcClient
...
+            logger.warn("Configured authentication method ({}) is not supported.", configurationMethod);
+        } 
+        
         if (CommonHelper.isNotEmpty(metadataMethods)) {
-            if (metadataMethods.contains(configurationMethod)) {
-                chosenMethod = configurationMethod;
+            if (preferredMethod != null && metadataMethods.contains(preferredMethod)) {
+                chosenMethod = preferredMethod;
             } else {
-                chosenMethod = metadataMethods.get(0);
```
leleuj: I found your code a bit complicated: isn't it this line only we should change to take the first supported authentication methods?

haumacher: There are some more problems: 
* When only changing this line, I still get a warning, if I do not configure a client authentication method, see the code following this line. This is not acceptable default behavior.
* In the case `!CommonHelper.isNotEmpty(metadataMethods)` (if the provider metadata does not provide any information of provider-supported methods), the code did choose `ClientAuthenticationMethod.getDefault()`, even if I have explicitly configured a method to use. This is quite absurd, because I'm not able to explicitly choose a method in case of missing metadata information from a provider. 
* The problem, that the explicitly configured method is not supported by the implementation should be handled separately resulting in an explicity warning message.
* The fact that the configured method could be `null` should be made explicit. In my opinion, a contains-check `metadataMethods.contains(configurationMethod)` should not be done with the argument being `null`, where `null` has a special meaning "I want to select the best matching provider-supported method".

leleuj: You're right. But I'd still prefer to simplify the source code with the explicit four use cases:metadata yes/no, configurationMethod: null/not null.

```java
if (metadata) {
  if (configMethod != null) {
    if (! metadata.contains(configMethod)) throw Exception
  } else {
    configMethod = metadata.search(client_secret_basic or client_secret_post)
    if (configMethod == null) throw Exception
  }
} else (no metadata) {
  if (configMethod != null) {
    if (configMethod != client_secret_basic and configMethod != client_secret_post) throw Exception
  } else {
    configMethod = client_secret_basic
  }
}
```

haumacher: In case 
```
if (metadata) {
  if (configMethod != null) {
    if (metadata.contains(configMethod)) {
```
The configured method may still not within the supported methods of the pac4j library. This is the reason, why I separated the lookup of the preferred method from the rest of the logic.

I'll try to re-formulate with your suggested test-sequence.

leleuj:
Using methods with meaningful names makes things much easier. Thanks

PR #1141, opened 2018-05-24
Add case-insensitive header name comparison in HeaderExtractor

petarov:
Headers are sent with lower case names by Google Chrome since version 60 [1].
This change seems to be a conformity to the rfc7320 spec, where header fields
are described as 'case-insensitive'. [2]

Unfortunately, this renders client authentication via DirectBasicAuthClient unusable.

This patch extends the header look up in HeaderExtractor by implementing a fall back
mechanism to look for a lowercase named header name.

[1] https://bugs.chromium.org/p/chromium/issues/detail?id=749086
[2] https://tools.ietf.org/html/rfc7230#section-3.2


leleuj:
Excellent! Thanks

petarov:
np, cheers! 🍺

PR #1147, opened 2018-06-14
Only skip anonymous profile if there is a 'better' profile

### LINKED ISSUE #1144: AnonymousProfile should not be skipped when flattened into a single profile
In ProfileHelper:

```java
public static <U extends CommonProfile> Optional<U> flatIntoOneProfile(Collection<U> profiles) {
        return profiles.stream().filter((p) -> {
            return p != null && !(p instanceof AnonymousProfile);
        }).findFirst();
    }
```

Is there a reason as to why a function that is supposed to be a generic flattener method takes on the opinion that anonymous profiles do not matter and should not be included? 

This condition should not be part of this method. If there is anything in the codebase that can't or won't operate or support AnonymousProfile(s), then those *things* need to implement this condition, and leave the method here to be as generic as possible. It doesn't make sense for one to authenticate the request using an anonymous profile, and then not be able to retrieve it from the profile manager.

leleuj:
Closes #1144

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -3,6 +3,10 @@ layout: doc
 title: Release notes&#58;
 ---
 
+**v3.0.1**:
+
+- The `ProfileHelper.flatIntoOneProfile` method returns an `AnonymousProfile` (instead of empty) if it's the only profile
+
 **v3.0.0**:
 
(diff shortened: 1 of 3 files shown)


mmoayyed:
Merge away! Thank you.

PR #1152, opened 2018-06-26
CLEANUP: Fixes typos, object construction using reflection, Collection use, missing annotations, etc

mmoayyed:


leleuj:
Excellent! Thanks

PR #1157, opened 2018-07-04
Refactoring GenderConverter with regular expressions。

zawn:
#749 this pull requst refactored the GenderConverter, This refactoring weakens the functionality of the GenderConverter ，Although the pr #749 has successfully merged the functionality of the GenderIntegerConverter, but the implementation has special code for VkProfileDefinition。we can't continue to use new GenderConverter("男", "女") or new GenderConverter("2", "1") to create an GenderConverter instance。

Currently I use regular expressions to better implement this feature and be more flexible.
 And we need to use on WeChat （new GenderConverter("男", "女") ）

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/converter/GenderConverter.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/converter/GenderConverter.java
@@ -1,41 +1,41 @@
 package org.pac4j.core.profile.converter;
 
+import java.util.regex.Pattern;
+
 import org.pac4j.core.profile.Gender;
 
 /**
  * This class converts a String to a Gender.
- * 
+ *
  * @author Jerome Leleu
  * @since 1.0.0
  */
 public final class GenderConverter extends AbstractAttributeConverter<Gender> {
 
+    private final Pattern maleText;
+    private final Pattern femaleText;
+
     public GenderConverter() {
         super(Gender.class);
+        this.maleText = Pattern.compile("(^m$)|(^male$)");
+        this.femaleText = Pattern.compile("(^f$)|(^female$)");
+    }
+
+    public GenderConverter(final String maleText, final String femaleText) {
+        super(Gender.class);
+        this.maleText = Pattern.compile(maleText);
+        this.femaleText = Pattern.compile(femaleText);
     }
 
     @Override
     protected Gender internalConvert(final Object attribute) {
-        if (attribute instanceof String) {
-            final String s = ((String) attribute).toLowerCase();
-            if ("m".equals(s) || "male".equals(s)) {
-                return Gender.MALE;
-            } else if ("f".equals(s) || "female".equals(s)) {
-                return Gender.FEMALE;
-            } else {
-                return Gender.UNSPECIFIED;
-            }
-            // for Vk:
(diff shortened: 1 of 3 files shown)


pac4j-core/src/main/java/org/pac4j/core/profile/converter/GenderConverter.java
```diff
@@ -1,41 +1,41 @@
...
  * @since 1.0.0
  */
 public final class GenderConverter extends AbstractAttributeConverter<Gender> {
 
+    private final Pattern maleText;
+    private final Pattern femaleText;
+
     public GenderConverter() {
         super(Gender.class);
+        this.maleText = Pattern.compile("(^m$)|(^male$)");
```
leleuj: Why not add here the "2" pattern for the male gender? Something like: `(^m$)|(^male$)|(^2$)`

pac4j-core/src/main/java/org/pac4j/core/profile/converter/GenderConverter.java
```diff
@@ -1,41 +1,41 @@
...
+        this.femaleText = Pattern.compile("(^f$)|(^female$)");
```
leleuj: Why not add here the "1" pattern for the female gender? Something like: `(^f$)|(^female$)|(^1$)`

pac4j-oauth/src/main/java/org/pac4j/oauth/profile/vk/VkProfileDefinition.java
```diff
@@ -60,7 +61,7 @@ public VkProfileDefinition() {
             CAN_WRITE_PRIVATE_MESSAGE})
             .forEach(a -> primary(a, Converters.BOOLEAN));
         primary(BIRTH_DATE, new DateConverter("dd.MM.yyyy"));
-        primary(SEX, Converters.GENDER);
+        primary(SEX, new GenderConverter("2", "1"));
```
leleuj: And keep the current `Converters.GENDER`?

zawn: Because f or female undoubtedly represents Gender.FEMALE, m or male undoubtedly represents Gender.MALE
But 1 and 2 may be different in different situations. For example, wechat uses 1 for male and 2 for female.
https://open.weixin.qq.com/cgi-bin/showdocument?action=dir_list&t=resource/res_list&verify=1&id=open1419316518&token=&lang=zh_CN
image
image

If we hardcode 1 and 2, we can't define wechat.

pac4j-core/src/main/java/org/pac4j/core/profile/converter/GenderConverter.java
zawn: 1 and 2 may be different in different situations. For example, wechat uses 1 for male and 2 for female.

pac4j-core/src/main/java/org/pac4j/core/profile/converter/GenderConverter.java
zawn: 1 and 2 may be different in different situations. For example, wechat uses 1 for male and 2 for female.

PR #1168, opened 2018-08-10
Problem: Saml2Credential should be serializable 

mmoayyed:
- Ensures that all classes that are marked as serializable are assigned an ID
- Modified the SAML2Credential class to not use native SAML2 objects that are not serializable; instead provides wrapper classes to hold the same data-set accessed by pac4j today.
- Add test cases for above.


mmoayyed:
Do not merge yet. Need to run some sanity tests against an IDP.

mmoayyed:
Tested. All works as expected.

leleuj:
I tested it and it works properly. Merging

leleuj:
Do you plan to merge: https://github.com/pac4j/pac4j/pull/1171 before the 3.1.0 release?

mmoayyed:
Yes, I think we should.

leleuj:
I just cut the release (3.1.0), it will show up in the Maven central repo in a few hours...

PR #1172, opened 2018-08-16
Fix typo in configuration setter for OAuth20Client.

aindlq:
I was hit hard by this in ini configs for shiro.


leleuj:
Good catch! Merging

Issue #1056, opened 2017-12-14
GenericOAuth20ProfileDefinition Can not set Id


cabbage89:
https://github.com/apereo/cas/blob/master/support/cas-server-support-pac4j/src/main/java/org/apereo/cas/support/pac4j/config/support/authentication/Pac4jAuthenticationEventExecutionPlanConfiguration.java#L304

The client should need to set the profile id

PR #1173, opened 2018-08-23
Allow GenericOAuth20Client to specify profileId

cabbage89:
https://github.com/apereo/cas/blob/5.3.x/support/cas-server-support-pac4j-core-clients/src/main/java/org/apereo/cas/support/pac4j/authentication/DelegatedClientFactory.java#L394

Because GenericOAuth20Client can't specify profileid, causing cas-oauth error, so we should allow the specified profileid

The Issues is similar to #1056


cabbage89:
You are welcome, this is what I should

PR #1188, opened 2018-10-02
Updates for OpenSAML support

scalding:
Two major changes:

1. updated to OpenSAML 3.4. There were some package changes that occurred
2. rework the way the OpenSAML library is bootstrapped. Use of the Java service API to find a configuration service, and falling back to default configuration if none is found. This allows for finer control over the library that some developers might need.


pac4j-saml/pom.xml
```diff
@@ -13,7 +13,7 @@
     <name>pac4j for SAML protocol</name>
 
     <properties>
-        <opensaml.version>3.3.1</opensaml.version>
+        <opensaml.version>3.4.0-SNAPSHOT</opensaml.version>
```
scalding: This will need to be changed when the final version is released

pac4j-saml/pom.xml
```diff
@@ -215,11 +225,18 @@
             <id>shib-release</id>
             <url>https://build.shibboleth.net/nexus/content/groups/public</url>
             <snapshots>
-                <enabled>false</enabled>
+                <enabled>true</enabled>
```
scalding: This will need to be changed when the final release of opensaml 3.4 is available

pac4j-saml/pom.xml
```diff
@@ -215,11 +225,18 @@
...
             </snapshots>
             <releases>
                 <enabled>true</enabled>
             </releases>
         </repository>
+        <repository>
```
scalding: this repository should be removed when the final release of opensaml 3.4 is available

mmoayyed:
LGTM. I earmarked this for 3.3.0.

pac4j-saml/pom.xml
```diff
@@ -191,6 +191,16 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
leleuj: I don't think this is necessary.

pac4j-saml/src/main/java/org/pac4j/saml/util/Configuration.java
```diff
@@ -46,55 +55,97 @@ private Configuration() {
...
-        try {
-            parserPool.initialize();
-        } catch (final ComponentInitializationException e) {
-            throw new RuntimeException("Exception initializing parserPool", e);
+        for (ConfigurationManager configurationManager : configurationManagers) {
+            configurationManager.configure();
         }
+    }
 
+    public static class DefaultConfigurationManager implements ConfigurationManager {
```
leleuj: Unless there is a specific reason, I'd prefer to have this class as a separate one.

documentation/docs/clients/saml.md
```diff
@@ -274,3 +274,81 @@ SimpleSAMLphp exposes his IdP metadata on `http://idp-domain/simplesamlphp/saml2
 	</md:EntityDescriptor>
 </md:EntitiesDescriptor>
 ```
+
+### Custom OpenSAML bootstrap
```
leleuj: Could you also add a line about this new feature in the "Release notes" doc: https://github.com/pac4j/pac4j/blob/master/documentation/docs/release-notes.md?

pac4j-saml/pom.xml
scalding: build failed locally without this dependency. There's a dependency on version 4.4 in opensaml 3.4 while 4.3 is being asserted by one of the deps (having a hard time tracking down which one).

leleuj: Copy that

documentation/docs/clients/saml.md
scalding: I've added a note in the release notes

pac4j-saml/pom.xml
scalding: found it; it's xmlsectool. the version I have here is compatible with xmlsectool. Oh, the joys of maven dependency resolution. I've created a property and moved this dep out of the testing block

scalding:
opensaml 3.4 was released. I think this is ready for final review

PR #1192, opened 2018-10-09
Allow saml2 authenticator to virtually map/rename attributes

mmoayyed:
Expose a new configuration setting to map SAML2 credential (identity provider) attributes virtually to new names before building the final SAML2 profile. An example might be to transform `urn:oid:2.16.840.1.113730.3.1.241` to `displayName`, etc.


leleuj:
LGTM. You could also have created a `MapAttributesProfileCreator` to map attributes, but it doesn't make much difference...

Issue #1200, opened 2018-11-06
classcast exception in CommonProfile : cannot cast jsonArray to String 


andre-doherty:
Hello,

There might be some issue between CommonProfile and CommonProfileDefinition.

The username field (when parsed from an SAML2Response) is mapped into a JSONArray.
Indeed in the CommonProfileDefinition there is no rule regarding that attribute 

As a result the getUsername in CommonProfile fails with a ClassCastException.

Shouldn't there be a rule for username as well to keep consistency between the 2 classes ?

André

leleuj:
In the `CommonProfile`,  the username is expected to be a `String`: https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/profile/CommonProfile.java#L75 which indeed may conflict with an attribute in the `SAML2Profile` which is a list.

I'm not sure to see the link the `JSONArray`:  can you paste the full stack trace? Thanks

andre-doherty:
yes here it is : 

`
java.lang.ClassCastException: net.minidev.json.JSONArray cannot be cast to java.lang.String
	at org.pac4j.core.profile.CommonProfile.getUsername(CommonProfile.java:62) ~[pac4j-core-3.0.2.jar:na]
	at org.pac4j.core.profile.Pac4JPrincipal.<init>(Pac4JPrincipal.java:24) ~[pac4j-core-3.0.2.jar:na]
	at org.pac4j.core.profile.CommonProfile.asPrincipal(CommonProfile.java:116) ~[pac4j-core-3.0.2.jar:na]
	at java.util.Optional.map(Optional.java:215) ~[na:1.8.0_112]
	at org.pac4j.j2e.util.Pac4JHttpServletRequestWrapper.getPrincipal(Pac4JHttpServletRequestWrapper.java:39) ~[j2e-pac4j-4.0.0.jar:na]
	at org.pac4j.j2e.util.Pac4JHttpServletRequestWrapper.getUserPrincipal(Pac4JHttpServletRequestWrapper.java:44) ~[j2e-pac4j-4.0.0.jar:na]
	at javax.servlet.http.HttpServletRequestWrapper.getUserPrincipal(HttpServletRequestWrapper.java:225) ~[com.ibm.websphere.javaee.servlet.3.1_1.0.22.jar:na]
[...]

PR #1203, opened 2018-11-07
3.4.x : fix mapping of attribute username to String

andre-doherty:
please find the pr regarding the issue #1200


pac4j-core/src/main/java/org/pac4j/core/profile/definition/CommonProfileDefinition.java
```diff
@@ -33,6 +34,7 @@ public CommonProfileDefinition() {
         primary(PICTURE_URL, Converters.URL);
         primary(PROFILE_URL, Converters.URL);
         primary(LOCATION, Converters.STRING);
+        primary(USERNAME, Converters.STRING);
```
leleuj: You can use the `Pac4jConstants.USERNAME` instead of creating a new one...

andre-doherty: Fixed

PR #1207, opened 2018-11-24
Feature/#1206

### LINKED ISSUE #1206: Support for Resource Owner Password Credential Grant for OIDC authentication mechanism
The Resource Owner Password Credential Grant flow is depicted by the following diagram.
```
    +----------+
     | Resource |
     |  Owner   |
     |          |
     +----------+
          v
          |    Resource Owner
         (A) Password Credentials
          |
          v
     +---------+                                  +---------------+
     |         |>--(B)---- Resource Owner ------->|               |
     |         |         Password Credentials     | Authorization |
     | Client  |                                  |     Server    |
     |         |<--(C)---- Access Token ---------<|               |
     |         |    (w/ Optional Refresh Token)   |               |
     +---------+                                  +---------------+
```

Currently, there is no way to authenticate your client using this flow. The idea is to get an access token from the authorization server using user credentials and using it to access a resource. 

The proposed method is to use the `HeaderClient` and create a specific `UserInfoOidcAuthenticator` which checks the access token on the userinfo endpoint and this creates an `OidcProfile`

rockydcoder:
This PR resolves #1206


rockydcoder:
@leleuj , can you review this please?

pac4j-oidc/pom.xml
```diff
@@ -62,6 +69,16 @@
             <artifactId>logback-classic</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
leleuj: We try to limit the dependencies pulled by pac4j even for testing. Let's remove this one, as I see it used only for statuses!

pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/authenticator/UserInfoOidcAuthenticator.java
```diff
@@ -0,0 +1,85 @@
...
+import com.nimbusds.oauth2.sdk.ParseException;
+import com.nimbusds.oauth2.sdk.http.HTTPRequest;
+import com.nimbusds.oauth2.sdk.http.HTTPResponse;
+import com.nimbusds.oauth2.sdk.token.BearerAccessToken;
+import com.nimbusds.openid.connect.sdk.UserInfoErrorResponse;
+import com.nimbusds.openid.connect.sdk.UserInfoRequest;
+import com.nimbusds.openid.connect.sdk.UserInfoResponse;
+import com.nimbusds.openid.connect.sdk.UserInfoSuccessResponse;
+
+public class UserInfoOidcAuthenticator implements Authenticator<TokenCredentials> {
```
leleuj: Javadoc required + `@author` and `@since 3.5.0` annotations on every class.

pac4j-oidc/pom.xml
rockydcoder: Also, it is not just the statuses that are being used from the module. `WebServer` extends `NanoHTTPD` which is being used in the test class. So, I think the dependency is required.

leleuj: OK. You're right. In that case, let's put the `nanohttpd.version` property in the main `pom.xml` file...

leleuj:
@rockydcoder For non-yet-existing Java files, just compute the URL as it will be, you won't be able to verify it before committing.

rockydcoder:
I am done with the changes

rockydcoder:
On retrospection, I think it would be a good idea to have a feature in pac4j to refresh the token if it has expired, since the refresh token is also sent by the OIDC provider when it authenticates a user. What do you think?

leleuj:
Feels good to me. Merging...

leleuj:
@rockydcoder About the expiration management in pac4j, currently, pac4j profiles have an `isExpired` method which is checked by the `ProfileManager`. For the OIDC protocol, the expiration date of the ID Token is checked.

PR #1208, opened 2018-11-24
add config building of oauth2

javasoze:
currently, there is not a way to have `PropertiesConfigFactory` to build an Oauth2Client via a configuration.

This PR aims to add the ability to specify parameters to build an Oauth2Client.

### DIFF
--- a/pac4j-config/src/main/java/org/pac4j/config/builder/OAuthBuilder.java
+++ b/pac4j-config/src/main/java/org/pac4j/config/builder/OAuthBuilder.java
@@ -121,4 +121,15 @@ public class OAuthBuilder extends AbstractBuilder {
             clients.add(twitterClient);
         }
     }
+
+    public void tryCreateOauth2Client(final List<Client> clients) {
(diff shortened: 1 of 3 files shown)


pac4j-config/src/main/java/org/pac4j/config/builder/OAuthBuilder.java
```diff
@@ -121,4 +121,15 @@ public void tryCreateTwitterClient(final List<Client> clients) {
             clients.add(twitterClient);
         }
     }
+
+    public void tryCreateOauth2Client(final List<Client> clients) {
+        final String key = getProperty(OAUTH2_KEY);
+        final String secret = getProperty(OAUTH2_SECRET);
+        if (isNotBlank(key) && isNotBlank(secret)) {
+            final OAuth20Client oAuth20Client = new OAuth20Client();
```
leleuj: Did you test this change? I don't think it works.

You should use the `GenericOAuth20Client` and set more properties: `authUrl`, `tokenUrl`, ...

PR #1211, opened 2018-12-05
Additional profile attributes for HiOrg-Server OAuth client

MartinBoehmer:
This adds some more profile attributes and minor improvements on the HiOrg-Server client and its documentation.

PR #1213, opened 2018-12-07
Additional profile attributes for HiOrg-Server OAuth client (Forward port)

MartinBoehmer:
Forward port of #1211.

This adds some more profile attributes and minor improvements on the HiOrg-Server client and its documentation.


leleuj:
Thanks

PR #1219, opened 2018-12-18
Upgrade nimbusds dependencies for JWT and OIDC

papegaaij:
This PR upgrades both NimbusDS dependencies to the latest version without changing the Pac4j API. It also adds a commons-lang3 dependency to pac4j-config, because that module depends on that lib, which is no longer provided via nimbus-jose-jwt.

### DIFF
--- a/pac4j-config/pom.xml
+++ b/pac4j-config/pom.xml
@@ -13,6 +13,7 @@
     <name>pac4j configuration</name>
 
     <properties>
+        <commonslang3.version>3.8.1</commonslang3.version>
         <hikaricp.version>2.7.6</hikaricp.version>
     </properties>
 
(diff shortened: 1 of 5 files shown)


leleuj:
@papegaaij As you can see, the build fails. This is why I'm reluctant to this change. It has a runtime breaking change (the size of the keys using the RSA algorithm).

papegaaij:
I suspect this is caused by the JDK used. I use OpenJDK, and looking at the output, it seems travis uses the Oracle JDK. This JDK comes with less strong encryption settings by default. You probably need the JCE unlimited strength jurisdiction policy files to enable large RSA keys.

As this failure is caused by the upgrade of nimbus-jose-jwt, I can try the PR with the jwt library at 6.0.2. The change in checks for weak keys was introduced in 6.1, see https://bitbucket.org/connect2id/nimbus-jose-jwt/src/a109fa2c60d16c091823fdcf92b9e552472b396c/CHANGELOG.txt?at=master&fileviewer=file-view-default

leleuj:
Excellent! It feels like a good compromise. Please update the PR. I re-opened it.

leleuj:
OK. Perfect. It builds now.

papegaaij:
Thanks for merging

leleuj:
Sure. Thanks for your contribution

PR #1239, opened 2019-01-21
Add direct support for RFC 6750 bearer authentication tokens.

minfrin:
This patch adds formal support for RFC 6750 bearer authentication tokens, carried in the Authorization header.

https://tools.ietf.org/html/rfc6750


minfrin:
Raised in error.

Issue #1228, opened 2019-01-08
Google+ API Being Deprecated


leleuj:
Discussion on the mailing list: https://groups.google.com/forum/?fromgroups#!topic/pac4j-users/xlFNWXEMEog

JohnBergant:
@leleuj  I have a branch I've been working on locally, would you mind if I run with this?

leleuj:
Sure. Go ahead. Notice we have a master branch (v4.0.0-SNAPSHOT) and a 3.5.x branch (v3.5.0-SNAPSHOT). You need to target at least the master branch, but you may want to target both...

ottboy4:
@JohnBergant - Do you have any updates on the progress of this? Thanks!

NgSekLong:
I have make an PR which attempt to solve this issue (see #1249), see if that PR is suitable, thanks.

PR #1249, opened 2019-02-06
[master] Fix Google Plus API deprecated issue

NgSekLong:
Related Issue: #1228 
Related Google Group discussion: https://groups.google.com/forum/?fromgroups#!topic/pac4j-users/xlFNWXEMEog

## What is changed
What this PR does, is changing the url from `https://www.googleapis.com/plus/v1/people/me` to `https://www.googleapis.com/oauth2/v3/userinfo`, with minor changes to make the test case and properties compatible again.

## What will break from this PR
This PR will break the following:
- `birthday` is removed
- `emails` is removed (Note: `email` without an `s` will still be retained)

Other then that, everything else should still be working as before.

------------------------------------
P.s. I am aware that someone else is also researching the fix, if they come up with a better fix / my fix is not compatible. I will be glad to delete / change this PR, cheers!

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -14,9 +14,11 @@ title: Release notes&#58;
 - By default, the CSRF check applies on the PUT, PATCH and DELETE requests in addition to the POST requests
 - "csrf,securityheaders" is the default authorizers definition
 - Renamed the `SAMLMessageStorage*` classes as `SAMLMessageStore*` (based on `Store`)
+- For `Google2Client`, change profile url from `https://www.googleapis.com/plus/v1/people/me` to `https://www.googleapis.com/oauth2/v3/userinfo`. This change is to prepare for the shutdown of Google plus API. This change will remove the `birthday` and `emails` attribute for `Google2Client`.
 
 **v3.6.0**:
 - Multiple authn context class refs can be set in the SAML protocol support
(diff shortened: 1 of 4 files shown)


leleuj:
OK. Thanks. Fairly easy. It looks good to me. You need to add something in the release notes here: https://github.com/pac4j/pac4j/blob/master/documentation/docs/release-notes.md

And, as pac4j v4 is not planned before a few months, you need to open a second PR targeting the 3.6.x branch. This one must keep the `getBirthday` method and return `null` and the `getEmails` method and return an empty list.

NgSekLong:
Ok, will do the above today if possible, thanks.

NgSekLong:
> You need to add something in the release notes

Done in this PR here.

> you need to open a second PR targeting the 3.6.x branch

It is submitted here #1250 

> This one(i.e. 3.6.x) must keep the `getBirthday` method 

Done accordingly

>  ... and (for 3.6.x) the `getEmails` method return an empty list.

For better backward compatibility experience, in my opinion it is better to resolve the `getEmails` from the `getEmail` function, so at least they can still get an email when using `getEmails`. See if the team agree with this. I can change it back to empty list if deemed appropriate. 


Thanks.

leleuj:
Excellent! Thanks for your contribution. Indeed, you were right about backward compatibility.

PR #1254, opened 2019-02-11
Improve metadata generation of signature & digest methods supported

scoldwell:


leleuj:
It might be a silly question, but why don't use the signature algorithms and stuffs defined in the `SAML2Configuration` class?

scoldwell:
I added that in SAML2ServiceProviderMetadataResolver. My reasoning for adding it to the resolver and not modifying the generator class itself was that maybe someone might want to use the generator class without the configuration dependency. I can always go a different direction if that's preferred.

leleuj:
OK. We'll see what @mmoayyed thinks...

mmoayyed:
Sure. Thank you for the changes. I'll have a look early next week.

mmoayyed:
This looks good to me; and I like the separation of config as well. I would ask that you might want to think about a few test cases to confirm some of this behavior. Otherwise, this is good to merge. Thanks again.

leleuj:
@mmoayyed thanks for your feedback
@scoldwell thanks for your contribution
Merging

scoldwell:
@mmoayyed I was planning to add test cases for it, but I realized with the runtime support checks that the tests may fail on different platforms.  I'll look into allowing the runtime checks to be bypassed for test purposes.

mmoayyed:
Gotcha though that wouldn't be a good idea. We'd want those checks to pass at least once preferrably for the CI we run here, to make sure we have at least one functional reference.

Issue #1258, opened 2019-02-13
IndirectClient (re)generating state token for ajax requests


brunojcm:
I have an application here that performs a lot of AJAX requests in parallel for a few reasons, and what is happening is that I get state check errors every now and then, which are pretty hard to debug.

I took a look at `org.pac4j.core.client.IndirectClient#getRedirectAction`, it has:

```
        if (ajaxRequestResolver.isAjax(context)) {
            logger.info("AJAX request detected -> returning the appropriate action");
            RedirectAction action = redirectActionBuilder.redirect(context);
            cleanRequestedUrl(context);
            return ajaxRequestResolver.buildAjaxResponse(action.getLocation(), context);
        }
```

What is very weird for me is that it calls `redirectActionBuilder.redirect(context)` even when it is an AJAX method, probably just to calculate the location, but it doesn't actually redirect. 

So far, so good, but it causes an issue: if `redirectActionBuilder.redirect(context)` is called, a new `State` is put in the session store. If I had already started my authentication redirection by a previous request (and this is just an AJAX call being performed by an iframe, for example, which I believe is the case here), by the time my SSO solution returns the result for the first process we have another state token in the session store (put by the AJAX request that happened in parallel), so my auth fails with a "State parameter is different from the one sent in authentication request. Session expired or possible threat of cross-site request forgery" message.

I tried to extend `OidcRedirectActionBuilder` so we don't overwrite the state for AJAX requests, but at that point, there is no way to check if the request is an AJAX or not (unless I do some ugly stuff), so that's why I'm creating this issue.

I believe that, if it is an AJAX, we should not call `redirectActionBuilder.redirect(context)` at all. Probably separating the logic to create the location from the actual redirect, but I'm not so sure about that.

Can you help, please?

brunojcm:
Started a PR with just a failing test case for now:

https://github.com/pac4j/pac4j/pull/1259

Please, let me know what you think.

leleuj:
The redirection URL is computed to be added as a header to the 401 HTTP response. I don't think it is used in most cases.
So I'm in line that it should be better not to compute it, although it should be possible.

So the `redirectActionBuilder.redirect(context)` should certainly happen in the `AjaxRequestResolver.buildAjaxResponse` method if it is necessary. Controlled by some `addRedirectionUrlAsHeader` property which is `false` by default.
We can do that in v4 (master branch) as it's a breaking change.

[...]

PR #1259, opened 2019-02-15
issue #1258 - ajax requests happening before auth is complete are overwriting state in session store

brunojcm:
creating PR firstly with the test only.

Not sure what's the best design to fix this, but my idea is to inject the `AjaxRequestResolver` into the `OidcRedirectionActionBuilder` and check in `org.pac4j.oidc.redirect.OidcRedirectionActionBuilder#redirect` if the request is ajax, not calling `addStateAndNonceParameters(context, params)` is case it is.

Please, let me know what you think.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/http/ajax/DefaultAjaxRequestResolver.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/http/ajax/DefaultAjaxRequestResolver.java
@@ -4,6 +4,7 @@ import org.pac4j.core.context.HttpConstants;
 import org.pac4j.core.context.Pac4jConstants;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.exception.http.*;
+import org.pac4j.core.redirect.RedirectionActionBuilder;
 import org.pac4j.core.util.CommonHelper;
 
 /**
@@ -13,7 +14,9 @@ import org.pac4j.core.util.CommonHelper;
  * @since 1.8.0
  */
 public class DefaultAjaxRequestResolver implements AjaxRequestResolver, HttpConstants, Pac4jConstants {
-
+    
+    private boolean addRedirectionUrlAsHeader = false;
+    
     @Override
     public boolean isAjax(final WebContext context) {
         final boolean xmlHttpRequest = AJAX_HEADER_VALUE.equalsIgnoreCase(context.getRequestHeader(AJAX_HEADER_NAME));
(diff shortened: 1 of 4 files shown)


leleuj:
See my comment: https://github.com/pac4j/pac4j/issues/1258#issuecomment-463946548

pac4j-core/src/main/java/org/pac4j/core/http/ajax/DefaultAjaxRequestResolver.java
```diff
@@ -13,21 +14,36 @@
...
         final boolean xmlHttpRequest = AJAX_HEADER_VALUE.equalsIgnoreCase(context.getRequestHeader(AJAX_HEADER_NAME));
         final boolean hasDynamicAjaxParameter = Boolean.TRUE.toString().equalsIgnoreCase(context.getRequestHeader(IS_AJAX_REQUEST));
         final boolean hasDynamicAjaxHeader = Boolean.TRUE.toString().equalsIgnoreCase(context.getRequestParameter(IS_AJAX_REQUEST));
         return xmlHttpRequest || hasDynamicAjaxParameter || hasDynamicAjaxHeader;
     }
-
+    
     @Override
-    public HttpAction buildAjaxResponse(final RedirectionAction action, final WebContext context) {
+    public HttpAction buildAjaxResponse(WebContext context, RedirectionActionBuilder redirectionActionBuilder) {
```
leleuj: I find this code a bit complicated. I'd prefer to have:

```java
    public HttpAction buildAjaxResponse(final WebContext context, final RedirectionActionBuilder redirectionActionBuilder) {
        String url = null;
        if (addRedirectionUrlAsHeader) {
            final RedirectionAction action = redirectionActionBuilder.redirect(context);
            if (action instanceof FoundAction) {
                url = ((FoundAction) action).getLocation();
            }
        }

        if ( CommonHelper.isBlank(context.getRequestParameter(FACES_PARTIAL_AJAX_PARAMETER))) {
            if (CommonHelper.isNotBlank(url)) {
                context.setResponseHeader(HttpConstants.LOCATION_HEADER, url);
            }
            throw UnauthorizedAction.INSTANCE;
        }

        final StringBuilder buffer = new StringBuilder();
        buffer.append("<?xml version='1.0' encoding='UTF-8'?>");
        buffer.append("<partial-response>");
        buffer.append("<redirect url=\"" + url.replaceAll("&", "&amp;") + "\"></redirect>");
        buffer.append("</partial-response>");

        return new OkAction(buffer.toString());
    }
```

brunojcm: I tried to keep the old method to avoid hard compatibility issues, but if you don't think it's worth it, that's all right! I did a commit with your version, just adding a `if` on the partial builder if url is null. Please, have a look. :)

leleuj: I got your idea, but as it's a major version, we can really change the source code.

leleuj:
Thanks for your contribution. Merging...

PR #1280, opened 2019-03-12
add Pac4j HTTPPostSimpleSignEncoder implementation

vincent314:
Implementing HTTP Post **Simple-Sign** protocol, inspired from `Pac4jHTTPPostEncoder` but extending directly openSAML class `HTTPPostSimpleSignEncoder`.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/transport/Pac4jHTTPPostSimpleSignEncoder.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/transport/Pac4jHTTPPostSimpleSignEncoder.java
@@ -1,0 +1,67 @@
+package org.pac4j.saml.transport;
+
+import org.apache.velocity.VelocityContext;
+import org.opensaml.messaging.context.MessageContext;
+import org.opensaml.messaging.encoder.MessageEncodingException;
+import org.opensaml.saml.common.SAMLObject;
+import org.opensaml.saml.common.binding.BindingException;
+import org.opensaml.saml.common.binding.SAMLBindingSupport;
+import org.opensaml.saml.saml2.binding.encoding.impl.HTTPPostSimpleSignEncoder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.OutputStreamWriter;
+import java.net.URI;
+
+/**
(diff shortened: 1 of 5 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/transport/Pac4jHTTPPostSimpleSignEncoder.java
```diff
@@ -0,0 +1,68 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.OutputStreamWriter;
+import java.net.URI;
+
+/**
+ * Pac4j implementation for HTTP Post Simple-Sign extending openSAML {@link HTTPPostSimpleSignEncoder}.
+ *
+ * @author Misagh Moayyed
```
leleuj: You're not Misagh: can you properly set the author and version?

pac4j-saml/src/main/java/org/pac4j/saml/transport/Pac4jHTTPPostSimpleSignEncoder.java
```diff
@@ -0,0 +1,68 @@
...
+            final OutputStreamWriter out = responseAdapter.getOutputStreamWriter();
+            this.getVelocityEngine().mergeTemplate(this.getVelocityTemplateId(), "UTF-8", velocityContext, out);
+            out.flush();
+        } catch (Exception e) {
+            throw new MessageEncodingException("Error creating output document", e);
+        }
+    }
+
+    @Override
+    protected void doInitialize() { }
```
leleuj: Why is the `doInitialize` method empty? Why not use the super parent one (which has several checks)?

leleuj:
Overall, it looks good.

Yet, there are a few things to fix:
- there are a few comments to address before the merge
- you need to submit a similar PR for the master
- the documentation must be updated, regarding the release notes: https://github.com/pac4j/pac4j/blob/master/documentation/docs/release-notes.md and the SAML support itself: https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/saml.md
- the build must be successful (remove the useless imports)

Thanks.

PR #1297, opened 2019-03-29
lower "AJAX request" log to debug level

brunojcm:
My app does a lot of AJAX requests for each client, so basically the log is flooded with this message :)

I think it's more like a debug than info, what do you think?


leleuj:
Yes, good point. Can you submit the same PR for the 3.7.x branch? Thanks

PR #1299, opened 2019-05-07
Add a link to akka-http-pac4j in README

vidma:
@leleuj 

added a demo project too: https://github.com/vidma/akka-http-pac4j-demo  tested with LDAP & SAML - works fine.


vidma:
to my knowledge, there are no docs. so I created a demo project from scratch. it wasn't as hard as I thought :)

most likely it doesn't work in akka-http Java, as it's all scala code and it uses `scaladsl` for directives like `withAuthentication` inside the routes file (which is like `routes + controller` in the same file).

leleuj:
@vidam Thanks for the pointer. I just updated the README and website...

Issue #1301, opened 2019-05-13
CsrfTokenGeneratorAuthorizer does not handle cookie maxAge


gaetan224:
CsrfTokenGeneratorAuthorizer does not handle cookie maxAge

PR #1302, opened 2019-05-13
add_maxAge_csrf_token_cookie

gaetan224:
proposed pr for #1301


leleuj:
You'll need to submit the same PR for the 3.7.x branch which is used by the current vertx-pac4j version...

PR #1304, opened 2019-05-16
Support response mode in any order

minfrin:
In https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#RegistryContents there
is a registry defined of possible response type combinations, however
https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Terminology in the
same spec adds the proviso that order does not matter, and pac4j until now does not support
this.

Switch to the com.nimbusds.oauth2.sdk.ResponseType implementation for response type
handling, which supports any order.


minfrin:
Please backport to v3.7.x if possible.

pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
```diff
@@ -282,11 +299,16 @@ public void setResourceRetriever(final ResourceRetriever resourceRetriever) {
...
+        return responseType.toString();
     }
 
     public void setResponseType(final String responseType) {
-        this.responseType = responseType;
+        responseTypeException = null;
+        try {
+            this.responseType = ResponseType.parse(responseType);
+        } catch (ParseException e) {
+            responseTypeException = new TechnicalException("Unrecognised responseType: " + responseType, e);
```
leleuj: Let's make things more simple: directly throw the `TechnicalException` here.

leleuj:
For the backport, I'd prefer you submit a new PR. Thanks

pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
minfrin: This changes the behaviour of the code to throw exceptions in the setters rather than in the init method like now, however if you're happy the change in behaviour is not a problem I'll update it.

leleuj: yes, but in any case, the exception will be thrown when the object is initialized so the behavior is changed. So I'd prefer to throw it as early as possible. Given the fact that this exception should never be thrown.

leleuj:
It's merged.

minfrin:
Thanks for this.

Syncing branches with forks in order to do the backport for you seems to be a difficult task, the advice I've found is to blow away the fork and start again.

Let me do that quickly and submit a backport PR.

minfrin:
Backport is here: https://github.com/pac4j/pac4j/pull/1306

PR #1307, opened 2019-05-17
Fix a NullPointerException when configuration.secret is omitted.

minfrin:
The OpenID Connect specification defines “none” as follows:

https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

but the Discovery spec makes no mention of “none” under
https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata
token_endpoint_auth_methods_supported and so some IDPs - AWS Cognito
explicitly - leave “none" out, and pac4j follows the secret path and crashes.

Add support for "none", and use it when the secret is not provided.


leleuj:
Thanks for your contribution.

PR #1308, opened 2019-05-17
feat: be able to configure a unique name on client signing certificates

jgribonvald:
Like already discussed here is the PR to customize certificate name generated for one client.
Test change are provided

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
@@ -1,5 +1,31 @@
 package org.pac4j.saml.config;
 
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.math.BigInteger;
+import java.net.InetAddress;
+import java.net.MalformedURLException;
+import java.nio.charset.StandardCharsets;
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.KeyStore;
+import java.security.PrivateKey;
+import java.security.Signature;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.ArrayList;
+import java.util.Calendar;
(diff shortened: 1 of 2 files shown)


leleuj:
As this change targets the master branch = the new major v4, I think we can introduce a breaking change by setting the `certificateNameToAppend` to the `clientName` by default.

jgribonvald:
@leleuj : It's not adressed specifically to master branch and so v4, I don't know your release and contribution policy. So it would be great to get it on all latest branch. So I can provide a PR to 3.6 (version used in CAS), but it will be merged to other branch ? I would need a such feature in the next CAS 6.1.x version So let me know what is the best.

Else you suggest that instead of adding a new property that we use the clientname appended to certificates files ? that's true it could be better but I have to normalyze the clientName String (replacing spaces, and other special chars). So should I rewrite my PR on this way ?

leleuj:
pac4j v4 is not released yet and won't be until a few months. So you certainly want to target the branch 3.7.x (a 3.7.0 release is planned before end of May).

Yes, you need to normalize the client name (I would say just keep a-z, A-Z and numbers should be enough).

jgribonvald:
@leleuj If I'm not wrong the clientName is not provided to the Configuration. So in any case I need to add a property expect if there is a property that can be used as an ID ?

leleuj:
Yes, indeed, this makes my comment completely irrelevant.

leleuj:
One thing we could do nonetheless to improve the PR is to mutualize the logic related to the name computation in the `getSigningXXX` methods...

jgribonvald:
Don't worry, it's monday and I didn't remember why I've made my change on this way and not on yours so... but know I understand well !
The question is more if we need to get and use the clientName ? If yes my property should be renamed ! But also I need to add a method to clean the String added.

jgribonvald:
> One thing we could do nonetheless to improve the PR is to mutualize the logic related to the name computation in the `getSigningXXX` methods...

I will improve that so

leleuj:
You can keep the property name as is. It's an option we will rarely use, but it makes sense when we have multiple IdPs and we want to retrieve the generated certificate files...

leleuj:
You should update this one like the one targeting the v3.7.0.

jgribonvald:
Done !

PR #1311, opened 2019-06-03
Disable type-checking for CAS Profile definitions

mmoayyed:
## Problem

The CAS profile definition, as part of the hierarchy, attempts to install a number of attribute converters to auto-convert attributes such as `gender`, `locale`, etc. These attributes are only converted if they match the converter's required attribute type. This assumption does not hold true for CAS, as all attribute released in validation payloads are of type `String` and will never be classified with their own dedicated type. With such type restrictions, (specially more so when CAS is running as an OIDC provider), certain attributes can never be released to clients because their value will never match the required type.

## Patch

- Allow gender matching to be case insensitive
- For CAS profile definitions, disable type checking that requires certain attributes such as `gender` to match a certain class

This should also be backported to 3.7.x.

### DIFF
--- a/pac4j-cas/src/main/java/org/pac4j/cas/profile/CasProfileDefinition.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/profile/CasProfileDefinition.java
@@ -2,7 +2,9 @@ package org.pac4j.cas.profile;
 
 import org.jasig.cas.client.authentication.AttributePrincipal;
 import org.pac4j.cas.client.CasProxyReceptor;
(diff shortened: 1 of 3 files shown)


leleuj:
Indeed, it makes sense.

mmoayyed:
PS This now ported over to the 3.7.x branch as well.

leleuj:
Thanks

leleuj:
I just realized we should have done that on a 3.8.x branch. Anyway, it will be available on the 3.7.1-SNAPSHOT version...

PR #1314, opened 2019-06-06
Backport to v3.8.0: Add a Saml2MetadataFilter implementation

minfrin:
that can be used to return SAML metadata to a caller.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/metadata/Saml2MetadataFilter.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/metadata/Saml2MetadataFilter.java
@@ -1,0 +1,79 @@
+package org.pac4j.saml.metadata;
+
+import java.io.IOException;
+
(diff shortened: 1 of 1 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/metadata/Saml2MetadataFilter.java
```diff
@@ -0,0 +1,79 @@
...
+ * /API/SAML2/metadata = saml2MetadataFilter
+ * 
+ * @author Graham Leggett
+ * @since 3.8.0
+ */
+public class Saml2MetadataFilter implements Filter {
+
+    private Config config;
+
+    private String client;
```
leleuj: Change the property name as well: `clientName`...

minfrin: It's been done.

PR #1321, opened 2019-06-12
Fix handling of WantAssertionsSigned and do not allow unsigned assertions

papegaaij:
This PR contains both commits, as discussed via email. The second patch depends on the first, therefore 2 PRs didn't make sense (the patch wouldn't apply).

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/crypto/ExplicitSignatureTrustEngineProvider.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/crypto/ExplicitSignatureTrustEngineProvider.java
@@ -1,23 +1,14 @@
 package org.pac4j.saml.crypto;
 
 import net.shibboleth.utilities.java.support.component.ComponentInitializationException;
-import net.shibboleth.utilities.java.support.resolver.CriteriaSet;
-
 import org.opensaml.saml.metadata.resolver.MetadataResolver;
 import org.opensaml.saml.metadata.resolver.impl.PredicateRoleDescriptorResolver;
 import org.opensaml.saml.security.impl.MetadataCredentialResolver;
-import org.opensaml.security.SecurityException;
-import org.opensaml.security.credential.Credential;
-import org.opensaml.security.credential.CredentialResolver;
-import org.opensaml.security.trust.TrustedCredentialTrustEngine;
 import org.opensaml.xmlsec.config.impl.DefaultSecurityConfigurationBootstrap;
 import org.opensaml.xmlsec.keyinfo.KeyInfoCredentialResolver;
-import org.opensaml.xmlsec.signature.Signature;
 import org.opensaml.xmlsec.signature.support.SignatureTrustEngine;
 import org.opensaml.xmlsec.signature.support.impl.ExplicitKeySignatureTrustEngine;
 import org.pac4j.saml.exceptions.SAMLException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Provider returning well configured {@link SignatureTrustEngine} instances.
@@ -26,20 +17,11 @@ import org.slf4j.LoggerFactory;
  * @since 1.8.0
  */
 public class ExplicitSignatureTrustEngineProvider implements SAML2SignatureTrustEngineProvider {
-    
-    private static final Logger log = LoggerFactory.getLogger(ExplicitSignatureTrustEngineProvider.class);
-    
+
     private final MetadataResolver metadataResolver;
-    
-    private final boolean signatureValidationDisabled;
 
     public ExplicitSignatureTrustEngineProvider(final MetadataResolver metadataResolver) {
-        this(metadataResolver, false);
-    }
-    
-    public ExplicitSignatureTrustEngineProvider(final MetadataResolver metadataResolver, final boolean signatureValidationDisabled) {
         this.metadataResolver = metadataResolver;
-        this.signatureValidationDisabled = signatureValidationDisabled;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


mmoayyed:
Could you please update the PR description to include the full use case and rationale, etc? "Via email" isn't exactly helpful for others who might stumble upon this patch. Thank you.

papegaaij:
The mail discussion was quite lengthy, but I've included the most important parts.

During the implementation of SSO with the Dutch service DigiD, I noticed an 
omission in the validation of signatures in SAML responses. When at the SP-
side the option WantAssertionsSigned is disabled (or left out, because false 
is the default), pac4j accepts SAML responses without any signature at all. In 
my opinion this is not a correct implementation of the specification. It allows 
an attacker to remove the signature and modify the content of the response and 
its assertions.

The cause of this problem is in how 
SAML2AuthnResponseValidator.validateAssertionSignature and 
AbstractSAML2ResponseValidator.validateSignatureIfItExists interact. The first 
method checks if the assertion has a signature, and if it does, also if it's 
valid. It then checks if assertions need to be signed, but mixes this with the 
signing of the message itself. These methods fail to raise an error when the 
message itself is not signed and signing of the assertion is not required.

My suggestion for a fix is to remove the check for 
peerContext.isAuthenticated() from validateAssertionSignature. This will make 
the signature on the assertion mandatory when requested. In addition, a new 
validation method is needed that checks if at least the message or all the 
assertions in it are signed.

Below is an overview of the possible combinations and expected vs actual outcome.

Msg-Signed: the message is signed
Assert-Signed: the assertion(s) is/are signed
WAS: WantAssertionSigned is enabled

Msg-Signed, Assert-Signed, WAS -> valid
Msg-Signed, Assert-Signed, No-WAS -> valid
Msg-Signed, No-Assert-Signed, WAS -> ERROR-1
Msg-Signed, No-Assert-Signed, No-WAS -> valid
No-Msg-Signed, Assert-Signed, WAS -> valid
No-Msg-Signed, Assert-Signed, No-WAS -> valid
No-Msg-Signed, No-Assert-Signed, WAS -> ERROR-2
No-Msg-Signed, No-Assert-Signed, No-WAS -> ERROR-3

ERROR-1:
This should be considered invalid because the SP requires the assertion-
elements to be explicitly signed. This message does conform to the web-profile 
requirements because the assertions are implicitly signed via inheritance. 
Pac4j currently (in my opinion) erroneously considers this message valid.

ERROR-2:
This should be considered invalid for 2 reasons. First, the SP requires the 
assertion-elements to be explicitly signed and they are not. Second, this 
message does not conform to the web-profile because the assertions are not 
signed (neither directly nor via inheritance). This case is handled correctly 
by pac4j.

ERROR-3:
This should be considered invalid because it does not conform to the web-
profile requirement that assertions must always be signed. Pac4j currently (in 
my opinion) erroneously considers this message valid.

Basically it boils down to the interpretation of section 2.4.4 in the metadata 
specification: "WantAssertionsSigned: Optional attribute that indicates a 
requirement for the <saml:Assertion> elements received by this service 
provider to be signed. If omitted, the value is assumed to be false. This 
requirement is in addition to any requirement for signing derived from the use 
of a particular profile/binding combination."

IMHO this states that the assertion elements itself need to be signed, which 
might be needed by the SP when it wants to pass the assertion to other 
systems. It also states that this is *in addition to* any requirement for 
signing derived from the use of a particular profile/binding combination. The 
web-profile requires assertions to be signed when using the POST binding, 
although it allows signing of the message.

leleuj:
LGTM. The build failure must be fixed.

@mmoayyed are you in line with this change?

After Misagh's feedback, the same PR will need to be opened on master.

leleuj:
I forgot to mention that the documentation must be updated: https://github.com/pac4j/pac4j/blob/master/documentation/docs/release-notes.md and https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/saml.md

papegaaij:
I've got a very busy schedule today and tomorrow, but I can update the PR on Wednesday to fix the build and update the documentation.

leleuj:
No problem. Take your time.

leleuj:
Can you sync with master? Thanks

mmoayyed:
I haven't had a chance to review this full yet, but I will do so before the weekend.

mmoayyed:
I guess the very first requirement is, for every scenario that is outlined we'd likely prefer to see a test case to ensure the behavior is kept and verified as we move forward. That would be a good start.

papegaaij:
The PR contains updated testcases. The testcases were asserting invalid handling of the WantAssertionsSigned flag, this is fixed in this PR. I've tried adding new testcases, but that's very hard in the current code base, because there is no code to construct an actual valid (or invalid) response. Adding a test harness for the response handling is a good thing but beyond the scope of this PR IMHO.

I do want to note that this PR fixes an actual security issue, because with WantAssertionsSigned disabled, pac4j implementations are now vulnerable to signature attacks. When WantAssertionsSigned is disabled, you can bypass signature validation by simply removing the signature. This will allow anyone to login as anyone.

pac4j-saml/src/main/java/org/pac4j/saml/crypto/ExplicitSignatureTrustEngineProvider.java
```diff
@@ -17,11 +26,20 @@
...
+    
     private final MetadataResolver metadataResolver;
+    
+    private final boolean signatureValidationDisabled;
 
     public ExplicitSignatureTrustEngineProvider(final MetadataResolver metadataResolver) {
+        this(metadataResolver, false);
+    }
+    
+    public ExplicitSignatureTrustEngineProvider(final MetadataResolver metadataResolver, final boolean signatureValidationDisabled) {
```
mmoayyed: Please inject the engine from the outside, rather than a boolean parameter to build the engine internally.

papegaaij:
@mmoayyed I'm not sure what testcases you are missing in this PR. If you look at the 8 scenario's from the description, this PR affects 2 (ERROR-1 and ERROR-3). These are reflected by the tests `testAuthenticatedResponseAndAssertionWithoutSignatureThrowsException` (ERROR-1) and `testResponseWithoutSignatureThrowsException` (ERROR-3). As I said before, pac4j would benefit from more tests in this area, but that would require code to generate responses and assertions with and without signatures. I don't have the time to work on that at the moment.

mmoayyed:
Ah, gotcha. Thanks for clarifying. I mistakenly took this PR as trying to cover all scenarios. I am happy to spend a bit of time adding test cases here to make sure we are covering everything. The question is, should we collaborate on the same PR, or would @leleuj prefer to see them outside here? Either way is fine with me.

Thanks again.

leleuj:
Your way. But if you want my opinion, you can do that on the same PR...

leleuj:
Thanks for this contribution as well.

PR #1325, opened 2019-06-21
Build on jdk11

papegaaij:
This PR fixes the Pac4j build on JDK11. No functional changes were made.

2 SAML related tests were fixed in a 'quick and dirty' way, but that seemed the best to me in the current code base. These tests would benefit from something like xmlunit, but that would require many more changes.


leleuj:
Can you prepare the same PR for master?

PR #1332, opened 2019-06-27
Artifact binding

papegaaij:
This PR adds support for the artifact binding for the SAML Reponse (not for the AuthnRequest). The artifact is resolved via a SOAP call to the artifact resolution service.

The implementation is complete, but it is lacking tests at the moment. Unfortunately, I see no way of testing the the majority of the classes without setting up an actual artifact resolution service. The most important classes are `SAML2ArtifactBindingDecoder` (which 'decodes' the artifact reference into the actual response) and `DefaultSOAPPipelineFactory` (which defines all checks on the communication and what information needs to be collected). These classes require a real `HttpSOAPClient` to test, which in turn requires an artifact resolution service. This would be hard to mock. Do you have any ideas on this?

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
@@ -66,6 +66,8 @@ public class SAML2MetadataGenerator implements SAMLMetadataGenerator {
 
     protected String assertionConsumerServiceUrl;
 
+    protected String responseBindingType = SAMLConstants.SAML2_POST_BINDING_URI;
+
     protected String singleLogoutServiceUrl;
 
     protected boolean authnRequestSigned = false;
@@ -190,7 +192,7 @@ public class SAML2MetadataGenerator implements SAMLMetadataGenerator {
 
         int index = 0;
         spDescriptor.getAssertionConsumerServices()
-            .add(getAssertionConsumerService(SAMLConstants.SAML2_POST_BINDING_URI, index++, this.defaultACSIndex == index));
+            .add(getAssertionConsumerService(responseBindingType, index++, this.defaultACSIndex == index));
         spDescriptor.getSingleLogoutServices().add(getSingleLogoutService(SAMLConstants.SAML2_POST_BINDING_URI));
         spDescriptor.getSingleLogoutServices().add(getSingleLogoutService(SAMLConstants.SAML2_POST_SIMPLE_SIGN_BINDING_URI));
         spDescriptor.getSingleLogoutServices().add(getSingleLogoutService(SAMLConstants.SAML2_REDIRECT_BINDING_URI));
@@ -326,6 +328,10 @@ public class SAML2MetadataGenerator implements SAMLMetadataGenerator {
 
     public final void setAssertionConsumerServiceUrl(final String assertionConsumerServiceUrl) {
         this.assertionConsumerServiceUrl = assertionConsumerServiceUrl;
+    }
+
+    public void setResponseBindingType(String responseBindingType) {
+        this.responseBindingType = responseBindingType;
     }
 
     public final void setSingleLogoutServiceUrl(final String singleLogoutServiceUrl) {
--- a/pac4j-saml/src/main/java/org/pac4j/saml/sso/artifact/SAML2ArtifactBindingDecoder.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/sso/artifact/SAML2ArtifactBindingDecoder.java
@@ -1,0 +1,88 @@
+package org.pac4j.saml.sso.artifact;
+
+import org.opensaml.messaging.context.InOutOperationContext;
+import org.opensaml.messaging.decoder.MessageDecodingException;
+import org.opensaml.saml.common.SAMLObject;
+import org.opensaml.saml.common.binding.impl.DefaultEndpointResolver;
+import org.opensaml.saml.common.messaging.context.SAMLPeerEntityContext;
+import org.opensaml.saml.common.xml.SAMLConstants;
+import org.opensaml.saml.metadata.resolver.impl.PredicateRoleDescriptorResolver;
+import org.opensaml.saml.saml2.binding.decoding.impl.HTTPArtifactDecoder;
+import org.opensaml.saml.saml2.metadata.ArtifactResolutionService;
+import org.opensaml.saml.saml2.metadata.IDPSSODescriptor;
+import org.opensaml.security.SecurityException;
+import org.opensaml.soap.client.http.PipelineFactoryHttpSOAPClient;
+import org.opensaml.soap.common.SOAPException;
+import org.pac4j.core.context.J2EContext;
+import org.pac4j.core.context.WebContext;
+import org.pac4j.saml.context.SAML2MessageContext;
+import org.pac4j.saml.metadata.SAML2MetadataResolver;
+import org.pac4j.saml.transport.AbstractPac4jDecoder;
+
+public class SAML2ArtifactBindingDecoder extends AbstractPac4jDecoder {
+    private final SAML2MetadataResolver idpMetadataResolver;
+
+    private final SAML2MetadataResolver spMetadataResolver;
+
+    private final SOAPPipelineProvider soapPipelineProvider;
+
+    public SAML2ArtifactBindingDecoder(final WebContext context, final SAML2MetadataResolver idpMetadataResolver,
+            final SAML2MetadataResolver spMetadataResolver, final SOAPPipelineProvider soapPipelineProvider) {
+        super(context);
+        this.idpMetadataResolver = idpMetadataResolver;
+        this.spMetadataResolver = spMetadataResolver;
+        this.soapPipelineProvider = soapPipelineProvider;
+    }
+
+    @Override
+    public String getBindingURI(SAML2MessageContext messageContext) {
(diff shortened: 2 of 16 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
```diff
@@ -190,7 +192,7 @@ protected final SPSSODescriptor buildSPSSODescriptor() {
 
         int index = 0;
         spDescriptor.getAssertionConsumerServices()
-            .add(getAssertionConsumerService(SAMLConstants.SAML2_POST_BINDING_URI, index++, this.defaultACSIndex == index));
+            .add(getAssertionConsumerService(responseBindingType, index++, this.defaultACSIndex == index));
```
leleuj: I'm wondering if we shouldn't add both bindings (POST and artifact)?

pac4j-saml/src/main/java/org/pac4j/saml/sso/artifact/SAML2ArtifactBindingDecoder.java
```diff
@@ -0,0 +1,96 @@
...
+                        throws SOAPException, SecurityException {
+                    super.send(endpoint, operationContext);
+                    transferContext(operationContext, messageContext);
+                }
+            };
+            soapClient.setPipelineFactory(soapPipelineProvider.getPipelineFactory());
+            soapClient.setHttpClient(soapPipelineProvider.getHttpClientBuilder().buildClient());
+
+            HTTPArtifactDecoder artifactDecoder = new HTTPArtifactDecoder();
+            artifactDecoder.setHttpServletRequest(((J2EContext) context).getRequest());
```
leleuj: This is a serious issue: this won't work in Play, Vert.x and many other frameworks. Can we stay at a more abstract level (web context)?

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
papegaaij: That should be possible, but it would require detecting the binding used or registering different services (urls). You can detect the artifact binding by the `SAMLart` parameter (query or post).

pac4j-saml/src/main/java/org/pac4j/saml/sso/artifact/SAML2ArtifactBindingDecoder.java
papegaaij: Unfortunately the `HTTPArtifactDecoder` requires the servlet request. It is only used to read parameters (`RelayState` and `SAMLart`), but the class is written in such a way that the request is mandatory. There is no way to modify the class (ie by overriding certain methods) to make it work without the request. It should be possible to implement most of the `HttpServletRequest` interface by wrapping the `WebContext`. I do want to use `HTTPArtifactDecoder`, because it performs most of the complicated stuff.

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
leleuj: I may be missing the point, but why not add the two bindings by default?

pac4j-saml/src/main/java/org/pac4j/saml/sso/artifact/SAML2ArtifactBindingDecoder.java
leleuj: Sure, I understand, but we can't put that in pac4j, this would break the pac4j core promise.

I took a look at the `HTTPArtifactDecoder`, it seems that the request is only used to get the `SAMLart` parameter. Apart from the fact that the code does not allow easy overriding, it does not seem to be used more than that.
I know it's painful, but can you copy/paste this class somehow and use the web context?

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
papegaaij: In the metadata that would be possible, but then you'd have to detect the type of the binding used by the IdP on an incoming request. At the moment `SAML2Client` uses a different `SAML2MessageReceiver` when configured for the artifact binding (see `initSAMLProfileHandler`). I guess it should be possible to write a `SAML2MessageReceiver` that wraps both receivers and detects the binding being used.

When you provide both bindings, it's up to the IdP to make the choice. This might cause issues as well, because the artifact binding might require some additional configuration. If a client is not configured correctly to use the artifact binding, it will break when upgrading pac4j and then there would be no way to disable it.

Another problem might be ill-configured IdPs, that do not understand the metadata, but I can't say if that's a valid concern or not.

Rethinking this: maybe it's a good idea to make the `responseBindingType` a list and call it `responseBindingTypes`.

pac4j-saml/src/main/java/org/pac4j/saml/sso/artifact/SAML2ArtifactBindingDecoder.java
papegaaij: In that case, I'd rather go for wrapping the web context in a dummy `HttpServletRequest`. IMHO `HTTPArtifactDecoder` contains way too much code to simply copy paste it. `WebContext` has enough information to implement at least the basics of `HttpServletRequest`. As you said, only `getParameter(String)` is used, but many of the other methods map just fine to `WebContext`. The ones missing can throw `UnsupportedOperationException`.

This way we can continue to use `HTTPArtifactDecoder` without having to copy paste the code (and remember to do that again and again on every upgrade of OpenSAML) and keep the contract of pac4j core.

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
leleuj: OK. So let's keep one binding.

pac4j-saml/src/main/java/org/pac4j/saml/sso/artifact/SAML2ArtifactBindingDecoder.java
leleuj: I understand your concerns, but how can you do wrap the web context in a `HttpServletRequest` for a framework which has none?

papegaaij: `HttpServletRequest` is just an interface which you can implement. The only problem is that it has a gazillion methods, many of which will not be supported by `WebContext`, but that's fine for this use case. I actually started that way when I wrote this code, but thought it would be fine to do it like this. I wasn't aware that pac4j promises not to depend on a servlet container like this. I can implement this tomorrow.

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
papegaaij: Yes, as an IdP is only going to use one binding anyway, I guess it doesn't make much sense to support both simultaneously.

pac4j-saml/src/main/java/org/pac4j/saml/sso/artifact/SAML2ArtifactBindingDecoder.java
papegaaij: I've implemented an `HttpServletRequest` and `HttpSession` on top of `WebContext`. Many of the methods do not map, but the most important ones do. At least it is more than enough for `HTTPArtifactDecoder`, which actually uses `getContentType()` as well:
```java
log.debug("Beginning to decode message from HttpServletRequest");
log.debug("HttpServletRequest indicated Content-Type: {}",
        getHttpServletRequest().getContentType());
```

leleuj: OK. But I still don't see how it can work in a non-JEE environment with an import on `HttpServletRequest`? It will return a `ClassNotFoundException`...

papegaaij: Well, you do need the servlet-api classes on the classpath. You only need the classes, you don't need the whole container or implementation.

leleuj: I understand, but this won't be acceptable. Can you revert your latest change? I'll merge it as is.

It will work for J2E only at first, I'll change the `HTTPArtifactDecoder` implementation later on.

papegaaij: I've removed the last commit from the PR

leleuj: OK. Thanks

leleuj:
@papegaaij I just added a new `Pac4jHTTPArtifactDecoder` (agnostic from the JEE servlet request) on the 3.8.x branch: can you test with an artifact binding scenario and confirm it works? Thanks

papegaaij:
Hi @leleuj. Thanks for merging this. I can confirm that your implementation works fine in the setup we are using.

leleuj:
OK. Good. I ported it forward.

PR #1335, opened 2019-07-09
Sign metadata when configured to do so

papegaaij:
Also open up the metadata generation API for customization.


leleuj:
LGTM Can you open the same PR for master? Thanks
CC @mmoayyed

PR #1337, opened 2019-07-10
Never sign AuthnRequests when using REDIRECT binding

papegaaij:
Signing is done via the Signature query parameter. This is explained in
section 3.4.4.1 DEFLATE Encoding of the binding specification. The
signature is never generated for AuthnRequests when using REDIRECT, and
if it were, it is now always removed as part of the DEFLATE encoding.

Note that the specification states that you should first generate an XMLSig, then remove it and then generate the Signature query parameter if the XMLSig was present in the first place. I found this rather inefficient (signature generation is quite expensive), so I opted for not generating the XMLSig at all.


leleuj:
LGTM
cc @mmoayyed

PR #1339, opened 2019-07-11
Remove only specific query parameters from HTTP-Redirect

jasonbw:
An Identity Provider Location URL can have query parameters that should
persist through an HTTP-Redirect encoding.

Testing done:
[x] Ran all tests in Pac4jHTTPRedirectDeflateDecoderTest

PR #1340, opened 2019-07-12
Remove only specific query parameters from HTTP-Redirect (#1339)

jasonbw:
An Identity Provider Location URL can have query parameters that should
persist through an HTTP-Redirect encoding.

(cherry picked from commit e70fd195f02c4676a4e4288b19024212c95b3c60)

Testing done:
[x] ran all Pac4jHTTPRedirectDeflateDecoderTest


leleuj:
Thanks

PR #1351, opened 2019-08-12
AbstractProfileService::convertAttributesToProfile sets profile id with username instead of id

marcadella:
Set profile id with id instead of username in legacy mode

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/service/AbstractProfileService.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/service/AbstractProfileService.java
@@ -235,11 +235,8 @@ public abstract class AbstractProfileService<U extends CommonProfile> extends Pr
                 profile.addAttribute(USERNAME, username);
             }
             final Object retrievedId = storageAttributes.get(getIdAttribute());
-            if (retrievedId != null) {
-                profile.setId(ProfileHelper.sanitizeIdentifier(profile, retrievedId));
-            } else {
-                throw new TechnicalException("Could not retrieve id"); //Should not happen
-            }
+            assertNotNull(getIdAttribute(), retrievedId);
+            profile.setId(ProfileHelper.sanitizeIdentifier(profile, retrievedId));
             if (isNotBlank(linkedId)) {
                 profile.setLinkedId(linkedId);
             }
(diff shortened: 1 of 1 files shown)


pac4j-core/src/main/java/org/pac4j/core/profile/service/AbstractProfileService.java
```diff
@@ -230,9 +230,15 @@ protected U convertAttributesToProfile(final List<Map<String, Object>> listStora
...
-                profile.setId(ProfileHelper.sanitizeIdentifier(profile, retrievedUsername));
+                profile.addAttribute(USERNAME, retrievedUsername);
             } else {
-                profile.setId(username);
+                profile.addAttribute(USERNAME, username);
+            }
+            final Object retrievedId = storageAttributes.get(getIdAttribute());
+            if (retrievedId != null) {
+                profile.setId(ProfileHelper.sanitizeIdentifier(profile, retrievedId));
+            } else {
```
mmoayyed: Else clause unnecessary.

marcadella: Well, most Technical exceptions are stuff that "should" not happen and are sanity check. But I can remove it.

leleuj:
Can you elaborate on this?

marcadella:
In the base code, `profile.setId(ProfileHelper.sanitizeIdentifier(profile, retrievedUsername));` sets the id with the username which is wrong in general: there is no reason why the id should equal the username. So I replaced this line with:
- `profile.setId(ProfileHelper.sanitizeIdentifier(profile, retrievedId));` to properly set the id
- `profile.addAttribute(USERNAME, retrievedUsername);` to set the username

leleuj:
@marcadella Yes, this is strange, but it's the way the "legacy" mode works. Do you encounter any specific issue?

marcadella:
Yes, on my setup id != username so the profile has an empty username and
the id is the username, which obviously is not right. I quick fixed it by
overriding the method but it would be nice if it was not needed.

ons. 14. aug. 2019 kl. 07:47 skrev LELEU Jérôme <notifications@github.com>:

leleuj:
I don't consider this a bug, it's the way the legacy mode works and I'm reluctant to change that.

Issue #1266, opened 2019-02-19
LinkedIn API upgrade


leleuj:
https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/migration-faq?context=linkedin/consumer/context

leleuj:
The LinkedIn v1 API still works, it's only deprecated. Close for now

miloradvojnovic:
The LinkedIn v1 API doesn't work anymore. Do you have plans for migration to the v2 API?

leleuj:
No plan yet, but we should do it.

PR #1359, opened 2019-08-27
LinkedIn v2 API support

vasvir:
LinkedIn v2 API support #1266


vasvir:
Ok the line was more than 140 chars and that triggers a checkstyle error. Abort and retry...

PR #1412, opened 2019-10-08
Respect the useNameQualifier flag when generating a SAML compliant

minfrin:
response during logout. Makes ADFS logout work.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutRequestBuilder.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutRequestBuilder.java
@@ -11,6 +11,7 @@ import org.opensaml.saml.saml2.core.LogoutRequest;
 import org.opensaml.saml.saml2.core.NameID;
 import org.opensaml.saml.saml2.core.SessionIndex;
 import org.opensaml.saml.saml2.metadata.SingleLogoutService;
+import org.pac4j.saml.config.SAML2Configuration;
 import org.pac4j.saml.context.SAML2MessageContext;
 import org.pac4j.saml.profile.SAML2Profile;
 import org.pac4j.saml.util.Configuration;
@@ -26,6 +27,8 @@ public class SAML2LogoutRequestBuilder {
 
     private String bindingType;
 
+    private boolean useNameQualifier;
+
     private int issueInstantSkewSeconds = 0;
 
     private final XMLObjectBuilderFactory builderFactory = Configuration.getBuilderFactory();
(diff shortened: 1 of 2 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutRequestBuilder.java
```diff
@@ -67,9 +71,12 @@ protected final LogoutRequest buildLogoutRequest(final SAML2MessageContext conte
         final NameID nameId = nameIdBuilder.buildObject();
         nameId.setValue(profile.getId());
         nameId.setFormat(profile.getSamlNameIdFormat());
-        nameId.setNameQualifier(profile.getSamlNameIdNameQualifier());
-        nameId.setSPNameQualifier(profile.getSamlNameIdSpNameQualifier());
-        nameId.setSPProvidedID(profile.getSamlNameIdSpProviderId());
+        if (this.useNameQualifier) {
+            System.out.println("************************** blah blah");
```
mmoayyed: Remove please.

pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutRequestBuilder.java
```diff
@@ -35,8 +38,9 @@
      *
      * @param bindingType the binding type
      */
-    public SAML2LogoutRequestBuilder(final String bindingType) {
-        this.bindingType = bindingType;
+    public SAML2LogoutRequestBuilder(final SAML2Configuration cfg) {
+        this.bindingType = cfg.getSpLogoutRequestBindingType();
+        this.useNameQualifier = cfg.isUseNameQualifier();
```
mmoayyed: Could you please elaborate on how this flag is used today?

pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutRequestBuilder.java
```diff
@@ -35,8 +38,9 @@
      *
      * @param bindingType the binding type
      */
-    public SAML2LogoutRequestBuilder(final String bindingType) {
-        this.bindingType = bindingType;
+    public SAML2LogoutRequestBuilder(final SAML2Configuration cfg) {
```
mmoayyed: It might be good to add the configuration object as a field and directly use that now, rather than individual fields that are basically duplicates.

pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutRequestBuilder.java
minfrin: The flag is used on login to work around a SAML spec compliance bug that causes ADFS logins to fail:

https://github.com/pac4j/pac4j/blob/d1cd6b89eae6d17e529a336451630319072940c3/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java#L164

This patch uses the same flag to fix the same SAML spec compliance bug that causes ADFS logouts to fail.

See the following thread: https://groups.google.com/forum/?hl=en#!topic/pac4j-users/A08HzzY0KBQ

pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutRequestBuilder.java
leleuj: @mmoayyed @minfrin Notice that this should go into a 3.9.x branch but I'll handle that after the merge...

leleuj:
@mmoayyed are you ok now for the merge?

PR #1442, opened 2019-11-07
Allow relative callback URI resolution at request-time for SAMLv2 client

davoustp:
The current implementation forces to configure an absolute callback URI when instantiating the SAMLv2 client.

This might be a problem when the pac4j-equipped application is multi-homed and can be reached using various URLs/domains, or when the application is moved from one domain to another ( requires reconfiguration of the SAMLv2 part of the app in the process).

This pull request streamlines these use cases by allowing to define a relative callback URI at configuration time (say: `/auth/callback`) and performs the callback URI resolution at request-processing time using the configured client's `UrlResolver`, so that the callback URI is resolved against the URI which was used initially to reach the app.

Note: you'll see that the `DefaultUrlResolver` was instantiated in the `IndirectClient` internal init without the flag allowing this resolution to occur, so I switched it on - but I don't have the full picture of the entire pac4j codebase, so I'm not sure if there is any side-effect that could affect other indirect clients. Let me know...

In case you want to move forward with this code change, I suspect another pull request on the 3.x branch is required - I'll be happy to do that as well.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
@@ -49,7 +49,7 @@ public abstract class IndirectClient<C extends Credentials> extends BaseClient<C
         // check configuration
         CommonHelper.assertNotBlank("callbackUrl", this.callbackUrl, "set it up either on this IndirectClient or on the global Config");
         if (this.urlResolver == null) {
-            this.urlResolver = new DefaultUrlResolver();
+            this.urlResolver = new DefaultUrlResolver(true);
         }
         if (this.callbackUrlResolver == null) {
             this.callbackUrlResolver = newDefaultCallbackUrlResolver();
--- a/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
@@ -90,7 +90,6 @@ public class SAML2AuthnRequestBuilder implements SAML2ObjectBuilder<AuthnRequest
         final SingleSignOnService ssoService = context.getIDPSingleSignOnService(this.bindingType);
         final String idx = this.assertionConsumerServiceIndex > 0 ? String.valueOf(assertionConsumerServiceIndex) : null;
         final AssertionConsumerService assertionConsumerService = context.getSPAssertionConsumerService(idx);
-        assertionConsumerService.setLocation(this.urlResolver.compute(assertionConsumerService.getLocation(), context.getWebContext()));
         return buildAuthnRequest(context, assertionConsumerService, ssoService);
     }
 
@@ -134,7 +133,9 @@ public class SAML2AuthnRequestBuilder implements SAML2ObjectBuilder<AuthnRequest
         if (assertionConsumerServiceIndex >= 0) {
             request.setAssertionConsumerServiceIndex(assertionConsumerServiceIndex);
         } else {
-            request.setAssertionConsumerServiceURL(assertionConsumerService.getLocation());
+            request.setAssertionConsumerServiceURL(
+                this.urlResolver.compute(assertionConsumerService.getLocation(), context.getWebContext())
+            );
         }
         request.setProtocolBinding(assertionConsumerService.getBinding());
 
--- a/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
@@ -23,6 +23,7 @@ import org.opensaml.saml.saml2.core.impl.NameIDPolicyBuilder;
 import org.opensaml.saml.saml2.core.impl.RequestedAuthnContextBuilder;
 import org.opensaml.saml.saml2.metadata.AssertionConsumerService;
 import org.opensaml.saml.saml2.metadata.SingleSignOnService;
+import org.pac4j.core.http.url.UrlResolver;
 import org.pac4j.saml.config.SAML2Configuration;
 import org.pac4j.saml.context.SAML2MessageContext;
 import org.pac4j.saml.profile.api.SAML2ObjectBuilder;
@@ -62,12 +63,14 @@ public class SAML2AuthnRequestBuilder implements SAML2ObjectBuilder<AuthnRequest
 
     private final XMLObjectBuilderFactory builderFactory = Configuration.getBuilderFactory();
 
+    private final UrlResolver urlResolver;
+
     /**
      * Instantiates a new Saml 2 authn request builder.
      *
      * @param cfg Client configuration.
      */
-    public SAML2AuthnRequestBuilder(final SAML2Configuration cfg) {
+    public SAML2AuthnRequestBuilder(final SAML2Configuration cfg, final UrlResolver urlResolver) {
         this.forceAuth = cfg.isForceAuth();
         this.comparisonType = getComparisonTypeEnumFromString(cfg.getComparisonType());
         this.bindingType = cfg.getAuthnRequestBindingType();
@@ -79,6 +82,7 @@ public class SAML2AuthnRequestBuilder implements SAML2ObjectBuilder<AuthnRequest
         this.providerName = cfg.getProviderName();
         this.extensions = cfg.getAuthnRequestExtensions();
         this.useNameQualifier = cfg.isUseNameQualifier();
+        this.urlResolver = urlResolver;
     }
 
     @Override
@@ -86,6 +90,7 @@ public class SAML2AuthnRequestBuilder implements SAML2ObjectBuilder<AuthnRequest
         final SingleSignOnService ssoService = context.getIDPSingleSignOnService(this.bindingType);
         final String idx = this.assertionConsumerServiceIndex > 0 ? String.valueOf(assertionConsumerServiceIndex) : null;
         final AssertionConsumerService assertionConsumerService = context.getSPAssertionConsumerService(idx);
+        assertionConsumerService.setLocation(this.urlResolver.compute(assertionConsumerService.getLocation(), context.getWebContext()));
         return buildAuthnRequest(context, assertionConsumerService, ssoService);
     }
 
--- a/pac4j-saml/src/test/java/org/pac4j/saml/client/AuthnRequestInflator.java
+++ b/pac4j-saml/src/test/java/org/pac4j/saml/client/AuthnRequestInflator.java
@@ -13,6 +13,13 @@ import java.util.zip.InflaterInputStream;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.utils.URLEncodedUtils;
 
+/**
+ * Utility class for decoding SAML authentication request payload.
+ * Originally present in class {@link RedirectSAML2ClientTests} and extracted for reuse in other tests. 
+ * 
+ * @author davoustp
+ * @since 4.0.0
+ */
 public class AuthnRequestInflator {
 
     /**
--- a/pac4j-saml/src/test/java/org/pac4j/saml/client/AuthnRequestInflator.java
+++ b/pac4j-saml/src/test/java/org/pac4j/saml/client/AuthnRequestInflator.java
@@ -15,6 +15,11 @@ import org.apache.http.client.utils.URLEncodedUtils;
 
 public class AuthnRequestInflator {
 
+    /**
+     * Extracts the AuthnRequest payload from the encoded URL location.
+     * @param location the URL location containing the encoded AuthnRequest payload
+     * @return the plain decoded AuthnRequest payload
+     */
     public static String getInflatedAuthnRequest(final String location) {
         final List<NameValuePair> pairs = URLEncodedUtils.parse(java.net.URI.create(location), "UTF-8");
         final Inflater inflater = new Inflater(true);
--- a/pac4j-saml/src/test/java/org/pac4j/saml/client/AuthnRequestInflator.java
+++ b/pac4j-saml/src/test/java/org/pac4j/saml/client/AuthnRequestInflator.java
@@ -1,0 +1,37 @@
+package org.pac4j.saml.client;
+
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.List;
+import java.util.zip.Inflater;
+import java.util.zip.InflaterInputStream;
+
+import org.apache.http.NameValuePair;
+import org.apache.http.client.utils.URLEncodedUtils;
+
+public class AuthnRequestInflator {
+
+    public static String getInflatedAuthnRequest(final String location) {
+        final List<NameValuePair> pairs = URLEncodedUtils.parse(java.net.URI.create(location), "UTF-8");
+        final Inflater inflater = new Inflater(true);
+        final byte[] decodedRequest = Base64.getDecoder().decode(pairs.get(0).getValue());
+        final ByteArrayInputStream is = new ByteArrayInputStream(decodedRequest);
+        final InflaterInputStream inputStream = new InflaterInputStream(is, inflater);
+        final BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8));
+        String line;
+        final StringBuilder bldr = new StringBuilder();
+        try {
+            while ((line = reader.readLine()) != null) {
+                bldr.append(line);
+            }
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
+        }
+        return bldr.toString();
+    }
+
+}
--- a/pac4j-saml/src/test/java/org/pac4j/saml/client/RedirectSAML2ClientTests.java
+++ b/pac4j-saml/src/test/java/org/pac4j/saml/client/RedirectSAML2ClientTests.java
@@ -13,7 +13,6 @@ import org.springframework.mock.web.MockHttpServletResponse;
 import java.util.Arrays;
 
 import static org.junit.Assert.*;
-import static org.pac4j.saml.client.AuthnRequestInflator.getInflatedAuthnRequest;
 
 /**
  * Redirection tests on the {@link SAML2Client}.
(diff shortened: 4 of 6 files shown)


pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
```diff
@@ -49,7 +49,7 @@ protected final void internalInit() {
         // check configuration
         CommonHelper.assertNotBlank("callbackUrl", this.callbackUrl, "set it up either on this IndirectClient or on the global Config");
         if (this.urlResolver == null) {
-            this.urlResolver = new DefaultUrlResolver();
+            this.urlResolver = new DefaultUrlResolver(true);
```
mmoayyed: I am not sure about this. This effectively changes the behavior of all IndirectClients who rely on defaults. Instead, why not do this for the SAML client only, or simply update the documentation to indicate how one would need to set the url-resolver if they need relative callback URLs?

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
```diff
@@ -79,13 +82,15 @@ public SAML2AuthnRequestBuilder(final SAML2Configuration cfg) {
...
         this.useNameQualifier = cfg.isUseNameQualifier();
+        this.urlResolver = urlResolver;
     }
 
     @Override
     public AuthnRequest build(final SAML2MessageContext context) {
         final SingleSignOnService ssoService = context.getIDPSingleSignOnService(this.bindingType);
         final String idx = this.assertionConsumerServiceIndex > 0 ? String.valueOf(assertionConsumerServiceIndex) : null;
         final AssertionConsumerService assertionConsumerService = context.getSPAssertionConsumerService(idx);
+        assertionConsumerService.setLocation(this.urlResolver.compute(assertionConsumerService.getLocation(), context.getWebContext()));
```
mmoayyed: This also doesn't look right to me. I am not a fan of overriding metadata behavior. Why not just change the ACS url based on what you need?

pac4j-saml/src/test/java/org/pac4j/saml/client/AuthnRequestInflator.java
```diff
@@ -0,0 +1,37 @@
...
+        final StringBuilder bldr = new StringBuilder();
+        try {
+            while ((line = reader.readLine()) != null) {
+                bldr.append(line);
+            }
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
+        }
+        return bldr.toString();
+    }
```
mmoayyed: You'll need:

1. Javadoc for the method
2. Streams must all be closed
3. This is effectively a utility class with only one static method and should be designed as such.

pac4j-saml/src/test/java/org/pac4j/saml/client/RedirectSAML2ClientTests.java
```diff
@@ -12,18 +10,10 @@
...
-import java.io.InputStreamReader;
-import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
-import java.util.Base64;
-import java.util.List;
-import java.util.zip.Inflater;
-import java.util.zip.InflaterInputStream;
 
 import static org.junit.Assert.*;
+import static org.pac4j.saml.client.AuthnRequestInflator.getInflatedAuthnRequest;
```
mmoayyed: Remove, and instead qualify with `AuthnRequestInflator.getInflatedAuthnRequest()` where needed.

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
leleuj: Yes, I don't like it at all. This smells thread issues...

Let's be clear on this: if you want to handle multiple hosts/callback URLs, you'll need multiple SAML2 clients. Otherwise I don't think it will work properly (thread issues, one metadata...)

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
davoustp: Yes, I hesitated before doing this change but had the same reasoning: why would a `UrlResolver` be instantiated without the resolution logic enabled?
There are actually three ways doing this:

1.  we leave it as it, and as you said, people will inject there one instance of `UrlResolver` with their appropriate settings - it's just one more things to do, things won't work out of the box
2. change the default `UrlResolver`in the `SamlClient`instead of the global `IndirectClient`, as mentionned by Misagh, making this breaking change more local
3. perform the global default in `IndirectClient` as pushed, which raises more questions about other clients and their behaviour

I tend to think that #2 is the best compromise, but if you want to keep with #1, that's all fine with me. Let me know.

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
davoustp: Doh, I actually pushed the wrong commit - this was a preliminary attempt to check feasibility. This change must actually be performed on the `AuthnRequest` which is a local request-scoped transient object, not the `AssertionConsumerService`,  indeed!
What's the best way to handle this? Do you want me to create a brand new pull request or should I rebase this pull request to the next commit I'll push with the changes?

pac4j-saml/src/test/java/org/pac4j/saml/client/AuthnRequestInflator.java
davoustp: I just moved this existing code to a separate class so it can be reused...
But OK, I'll do the changes.

pac4j-saml/src/test/java/org/pac4j/saml/client/RedirectSAML2ClientTests.java
davoustp: OK

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
leleuj: I'm more in favor of 3: if we have this logic and I don't think it really matches real-world use cases, let's use it, it will be sufficient for tests as least. Or option 4, let's remove it.

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
leleuj: Let's do that on this PR. You still have the issue of one metadata for N ACS URLs...

pac4j-saml/src/test/java/org/pac4j/saml/client/AuthnRequestInflator.java
leleuj: Yes, it's that now we see it ;-) I also like to have Javadoc on the class with the `@since` and `@author` annotations...

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
davoustp: OK - so I leave the change originally proposed in the PR (copied below as well for convenience), right ?
```
-            this.urlResolver = new DefaultUrlResolver();
+            this.urlResolver = new DefaultUrlResolver(true);
```

pac4j-saml/src/test/java/org/pac4j/saml/client/AuthnRequestInflator.java
davoustp: :-D
Class-level Javadoc pushed.

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
davoustp: I'm not sure I get this - but let's dig into it to make sure there is no hidden problem.

The code I committed since then performs the URL resolution only in the case when no `AssertionConsumerServiceIndex` is provided.
In other words, in case of an ACS metadata providing with multiple URLs *and* a `AssertionConsumerServiceIndex` is specified, then there is no URL location set nor resolved (even if configured).

In the other case, pac4j-saml relies on the configured callback URL - and this is where there is now URL resolution taking place.

Is there any missing case which I'm blind to?
Could you have a look at the latest committed code and let me know what you think?

leleuj:
@davoustp I've been thinking about your PR. Although it raises good questions, I think we're wrong in the approach.

Let's go back to the definitions:
- the `UrlResolver` is meant to turn a given URL into an appropriate URL for the current infrastructure. Most of the time, it does nothing as there are absolute URLs. But it could turn relative URLs into absolute ones by retrieving the host, scheme and port in the appropriate way
- the `CallbackUrlResolver` is meant to turn a URL into a callback one (for example, by adding the `client_name` parameter).

The `CallbackUrlResolver` relies on the `UrlResolver`.

What you propose works, but it's luck. In SAML, the callback URL put in the metadata is computed at the initialization with a `null` web context, so the `CallbackUrlResolver` is called which calls the `UrlResolver` but this latter does nothing  as the web context is `null`. Then, you propose to apply the `UrlResolver` to the computed callback url to compute the authn request and it works as finally, both resolvers will be applied in the reverse order as it should be.

The metadata will never be correct as the `UrlResolver` will be never applied. Which is not a problem as most of the time absolute URLs are used. I think it's the first issue to solve. Then we could think about using the `CallbackUrlResolver` to compute the callback URL in the authn request, which is somehow a trick as we'll never have more than one metadata file.

I will do some exploratory work on this.

davoustp:
@leleuj Thanks for taking the time to analyse this.

Agreed with the definitions - no misunderstanding here.

When you say:
> In SAML, the callback URL put in the metadata is computed at the initialization with a null web context, so the CallbackUrlResolver is called which calls the UrlResolver but this latter does nothing as the web context is null.
you are referring to the pac4j-saml implementation, not the SAML specs, right? I'm at loss at finding any such constraint in the OASIS SAML specs, but I'm no expert at this, so I may have missed something...

> The metadata will never be correct as the UrlResolver will be never applied. Which is not a problem as most of the time absolute URLs are used. I think it's the first issue to solve. 
I don't see which problem you are trying to solve here - but I'm probably shortsighted.
If the configured URL is not an absolute one, there is no way to infer the absolute URL at init time, since the web container is unaware of any upfront network component, and therefore oblivious to the domain which can be used to reach it. You'll need an HTTP request context to perform such resolution.

What is the problem of having the metadata containing a non-absolute URL? As far as I could tell, this is an in-memory only object, right? Or is there any other place or call flow that would use this metadata as is, and which could then be broken by having a non absolute URL?

leleuj:
I'm referring to the pac4j-saml implementation: https://github.com/pac4j/pac4j/blob/master/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java#L203

Indeed, it might not be possible to properly compute a relative URL at startup, this is something I must investigate.
The SP metadata is the data you will give to your IdP and I strongly doubt it will work if there is a relative URL in the metadata.

That said, I realize that this is not blocking for your PR. What you should change is to use the `CallbackUrlResolver` on the `callbackUrl` instead of the `UrlResolver` on the `location`.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
mmoayyed: No. You will need to revert the change back to original. Changing behavior globally for all clients/users is never ever acceptable. 

People will inject their one instance of UrlResolver with their appropriate settings. It's one more thing to do, and things do work out of the box except for those who have a different specialized environment than the 90% norm in which case we expect them to be capable enough and have the technical know-how to make a one-liner change in their configuration of the pac4j library.

leleuj: That said, it's a major version (v4) so we can change things globally. Does it make sense? I think it won't change the behavior for most people, but as we have this extra behavior available, I tend to agree...

mmoayyed: I don't necessarily agree. Yes, a major version allows one to theoretically change things, assuming there is an understanding of what major/minor means in semantic terms. However, just as a policy allows one to change things, it doesn't mean that we should. Breaking code or behavior, regardless of what some document says, is never a pleasant experience for upgrades and users.

Specially in this case, the change in question is a one-liner that can be changed by those interested with a specialized set-up. So it's not a huge effort or fix to expect. There is also no test case to ensure this change breaks existing behavior or covers new ground which makes it more difficult to verify. Finally, *invisible* behavioral changes are the worst when people upgrade. There is no warning that the behavior shall change and the only way to learn this is by shuffling through possible release notes. I am generally fine with breaking APIs for a new major release; Breaking behavior invisibly is not a good idea, again, for such as simple thing that anyone can simply adjust with one line.

mmoayyed: A compromise might be:

1. Activate this behavior conditionally, on-demand.
2. Warn users with a log message, indicating this behavior might at some point become the default.
3. Allow time for the message to sink in.
4. Then, make the change globally.

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
mmoayyed: Manipulating ACS urls dynamically is a violation of the specification and the deployment profile:

> The AssertionConsumerServiceURL value, if present, MUST match an endpoint location expressed in the SP’s metadata exactly, without requiring URL canonicalization/normalization.

pac4j-core/src/main/java/org/pac4j/core/client/IndirectClient.java
leleuj: Yes, indeed, but I'm not sure to see what can be: "conditionally, on-demand" in this case.

If we have a doubt, let's stick to the current behavior as it can still be enabled fairly easily.

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnRequestBuilder.java
leleuj: We should have started here in fact ;-)

Among all things, we won't go against the the spec for sure. Even if this trick was appealing.

davoustp:
Guys,
You are the experts and decision makers here, so please do not take any offense, but I'd like to understand where this constraint is coming from...

Could you elaborate on where it's written or hinted at in the specs that such request-time URL resolution mechanism would be a violation?
I've looked at all the content provided in the official spec package (the zip file http://docs.oasis-open.org/security/saml/v2.0/saml-2.0-os.zip provided at http://saml.xml.org/saml-specifications), and I cannot find any reference to the text you mentioned:
> The AssertionConsumerServiceURL value, if present, MUST match an endpoint location expressed in the SP’s metadata exactly, without requiring URL canonicalization/normalization.
Where is it located exactly?

Starting with the xsd files, it specifies that field `AssertionConsumerServiceURL` is of type `anyURI`, which allows relative URIs as per http://www.datypic.com/sc/xsd/t-xsd_anyURI.html.

Same applies to with the Location attribute in the metadata, see saml-metadata-2.0-os.pdf @ line 232:
> Location [Required]
> A required URI attribute that specifies the location of the endpoint. The allowable syntax of this URI depends on the protocol binding.
However, no further detail is provided in the saml-bindings-2.0-os.pdf file for the HTTP Redirect or POST bindings.

But this may be a loosely constrained type for the sake of schema validation ease, so I looked at the various pdf specification files more thoroughly. Here, it is said that the `AssertionConsumerServiceURL` must be validated at IdP side to ensure that the target Service Provider is indeed an authorized location - see saml-core-2.0-os.pdf @ line 2062:
> Specifies by value the location to which the <Response> message MUST be returned to the requester. The responder MUST ensure by some means that the value specified is in fact associated with the requester. [SAMLMeta] provides one possible mechanism; signing the enclosing <AuthnRequest> message is another.

This implies to me that the `AssertionConsumerServiceURL` is absolute when received by the IdP (no canonization or normalization occurs here), which is obvious for Redirect or POST bindings, but not that this URL is absolute in the Service Provider metadata.
This looks also clear to me that the validation process of the URL is open, two of them are hinted at but no constraint is explicitly stated ("...ensured by some means.."), as far as I can tell.
Indeed, during my few experiments with Keycloak and Azure AD, this validation is performed by explicitly declaring which target URLs are allowed, so hitting the IdP with a URL not matching any of the configured ones at IdP side is just plainly rejected, as it should be. But nothing related to the value specified in the `AssertionConsumerServiceURL` , which I specified as relative without any problem in both systems.

Later on, on the SOAP profile, it is also explicitly stated that field `AssertionConsumerServiceURL` can be relative... Weird, uh?

Please let me know your thinking, and don't hesitate to prove me wrong. :-)

Thx

leleuj:
I will let @mmoayyed provide the reference to the right part of the spec...

leleuj:
BTW, I took a closer look at injecting the `WebContext` for the metadata generation and the change is fairly huge, so I push that aside for now.

I have the feeling we won't reach an agreement on this, me and Misagh being reluctant. So my best proposal is to send a new pull request which will allow to easily use your specific `SAML2AuthnRequestBuilder`.

mmoayyed:
Please take a look at the SAMLProfiles (Implementation and Deployment) produced by the Kantara Initiative.

davoustp:
@mmoayyed 
Ok, so the constraint is expressed in the Kantara Initiative, and not in the original SAMv2 specifications - that explains why I could not find it. Thx.

@leleuj 
Thanks for the suggestion, Jérôme.
I went down into checking how much could be done to avoid hard-wiring / hard-configuring the absolute callback URL using the existing pac4j-saml module. Because the lib is so much locked down, I stopped when hacking around these by-design constraints led me to code that would diverge from the lib itself, would potentially introduce security or maintenance issues.

Instead, I went another route, which I describe here for the sake of completeness - and potentially helping others having the same problem to solve: since the pac4j-saml lib enforces one client instance per absolute callback URL, I implemented my own `ClientFinder` which materializes the SAML client with the proper absolute URL callback (using the current `WebContext`), and injected it into the `DefaultSecurityLogic` and `DefaultCallbackLogic`.
This works because the `ClientFinder` is involved in the very early stages of both `SecurityLogic` instances - therefore the security logic itself acts on the materialized clients themselves.

Regards

PR #1447, opened 2019-11-15
Expose certificate algorithm, expiration and key length in SAML2 configuration

mmoayyed:
We might also consider to switch key sizes to 3072 in the future.


leleuj:
Thanks

PR #1450, opened 2019-11-17
Support generation of ContactPerson elements for SAML2 SP metadata

mmoayyed:
Also allows one to customize the supported protocols.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataContactPerson.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataContactPerson.java
@@ -1,0 +1,61 @@
+package org.pac4j.saml.metadata;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class SAML2MetadataContactPerson {
+    private String givenName;
+    private String surname;
+    private List<String> emailAddresses = new ArrayList<>();
+    private List<String> telephoneNumbers = new ArrayList<>();
+    private String type = "technical";
+    private String companyName;
+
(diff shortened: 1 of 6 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataContactPerson.java
```diff
@@ -0,0 +1,61 @@
+package org.pac4j.saml.metadata;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class SAML2MetadataContactPerson {
```
leleuj: The class Javadoc is missing (+ `@author` + `@since`)

pac4j-saml/src/test/java/org/pac4j/saml/client/PostSAML2ClientTests.java
```diff
@@ -30,6 +32,16 @@ public void testCustomSpEntityIdForPostBinding() {
         final SAML2Client client = getClient();
         client.getConfiguration().setServiceProviderEntityId("http://localhost:8080/cb");
         client.getConfiguration().setUseNameQualifier(true);
+
+        final SAML2MetadataContactPerson person = new SAML2MetadataContactPerson();
```
leleuj: I'm not sure to understand if you test something here or if this information is now required (I though not).

mmoayyed: It's not required, but the block will make sure the changes are covered by the existing test when metadata is generated.

PR #1455, opened 2019-11-20
Handle not falling back to anonymous if credentials to a client are provided but the client does not authenticate

garpinc:
In my use case I attempt multiple clients including the anonymous client to see if one of them authenticates. At the end if none does authenticate but anonymous does then you can proceed with anonymous authentication only if no credentials were provided to any of the other clients that determined that there were credentials for it, but then did not authenticate,

You wouldn't want to for instance query a db for stuff you could see as a particular user and then get back results instead for anonymous user. the client may want to error out in that case.

In my usage I keep track of lastAnonymousAuthentication and  sCredentialsPresented which I determine by calling below method. then if (lastAnonymousAuthentication.get() != null &&
!isCredentialsPresented.get() I allow the anonymous credentials/

With this method in a request the map captures per client what credentials were passed so you can
action on that.

```
    private static void populateIsCredentialsPresented(Client<? extends
Credentials, ? extends UserProfile> client, J2EContext context,
AtomicBoolean isCredentialsPresented) {
        @SuppressWarnings("unchecked")
        Map<Client<? extends Credentials, ? extends UserProfile>,
Credentials> credentialsSuppliedMap = (Map<Client<? extends Credentials,
? extends UserProfile>, Credentials>)
context.getRequestAttribute(BaseClient.CREDENTIALS_SUPPLIED_MAP);
        if (credentialsSuppliedMap != null) {
            Credentials credentialsSupplied =
credentialsSuppliedMap.get(client);
            if (credentialsSupplied != null && !(credentialsSupplied
instanceof AnonymousCredentials)) {
                if (credentialsSupplied instanceof
UsernamePasswordCredentials) {
                    UsernamePasswordCredentials usernamePwdCred =
(UsernamePasswordCredentials) credentialsSupplied;
                    if ((usernamePwdCred.getUsername() == null ||
usernamePwdCred.getUsername().isEmpty()) &&
(usernamePwdCred.getPassword() == null ||
usernamePwdCred.getPassword().isEmpty())) {
                        return;
                    }
                }
                isCredentialsPresented.set(true);
            }
        }
    }
```

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
@@ -65,11 +65,11 @@ public abstract class BaseClient<C extends Credentials> extends InitializableObj
             optCredentials.ifPresent(credentials -> {
                 @SuppressWarnings("unchecked")
                 Optional<Map<Client<C>, C>> optCredentialsSuppliedMap = 
-                    context.getRequestAttribute(CREDENTIALS_SUPPLIED_MAP);
+                        context.getRequestAttribute(CREDENTIALS_SUPPLIED_MAP);
                 
                 Map<Client<C>, C> credentialsSuppliedMap;
                 if (optCredentialsSuppliedMap.isPresent()) {
-                	credentialsSuppliedMap = optCredentialsSuppliedMap.get();
+                    credentialsSuppliedMap = optCredentialsSuppliedMap.get();
                 } else {
                     credentialsSuppliedMap = new ConcurrentHashMap<>();
                     context.setRequestAttribute(CREDENTIALS_SUPPLIED_MAP, credentialsSuppliedMap);
--- a/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/client/BaseClient.java
@@ -38,6 +38,8 @@ import org.slf4j.LoggerFactory;
 public abstract class BaseClient<C extends Credentials> extends InitializableObject implements Client<C> {
 
     protected final Logger logger = LoggerFactory.getLogger(getClass());
+    
+    public static final String CREDENTIALS_SUPPLIED_MAP = "credentialsSuppliedMap";
 
     private String name;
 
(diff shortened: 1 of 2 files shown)


leleuj:
Direct clients are tried one after the other: if one authentication does not succeed whether it is because there is no credentials or the credentials are not correct, we try the next one. This is a very simple logic I want to keep.

Other logics require customization. If you want to stop the authentication process after some credentials has been passed but do not match, you need to throw an `UnauthorizedAction` from your specific `Authenticator`.

garpinc:
The problem with this is most authenticators are not ones I write so I can't change them and if I override them with a subclass each one requires specific custom logic as they all indicate failure in their own way. The framework should have some way to capture if the authentication succeeded for that client so upstream that can be determined and you may want also to know which credential was provided that did not succeed.



- Garry


-------- Original message --------

garpinc:
Yeah I've looked into it and I find the API lacking.

in org.pac4j.core.client.BaseClient.retrieveCredentials(WebContext) it calls

            final C credentials = this.credentialsExtractor.extract(context);

If credentials are null then nothing was presented and return goes  back to the caller without ever calling the authenticator.validate method hence there is no way to know from caller if credentials were presented or not. null is returned if there is a CredentialsException or if nothing is presented at all so there is no way to know the difference. The code change I made considers the fact that the public API has been like this for sometime hence people maybe relying on it being null in both cases. But I need to distinguish. So the map that's tied to request allows the api to remain the same but allows you to get the info if you need it.

garpinc:
I’ll have to investigate this. UnauthorizedAction was not available in 3.8.x

I assume you are no longer doing any backports?

________________________________

leleuj:
I understand.
You are the first one to ask for that so it's an edge use case. The solution is not well integrated with the API: there is no modification, it's just about keeping the credentials in the request.
I'm still reluctant to this change unless @mmoayyed is enthusiastic about it...

leleuj:
As a solution, I think you can iterate through all your clients and wrap each `CredentialsExtractor` in one `KeepCredentialsInRequestWrappedCredentialsExtractor`...

leleuj:
In pac4j v3.x, it's `HttpAction.unauthorized`...

garpinc:
> As a solution, I think you can iterate through all your clients and wrap each `CredentialsExtractor` in one `KeepCredentialsInRequestWrappedCredentialsExtractor`...

That would be a great solution if the client were passed to credentialsExtractor but it isn't hence there is no way to know which client was associated with with credential

leleuj:
You can do whatever you want as you create the wrapper and iterate over the clients. Something like:

```java
for (BaseClient client : clients) {
   client.setCredentialsExtractor(new KeepCredentialsInRequestWrappedCredentialsExtractor(this, client.getCredentialsExtractor());
}
```

garpinc:
@leleuj @mmoayyed  there was a problem with this I encountered so wanted to make sure you're still on board with what you proposed. client.getCredentialsExtractor() returns nothing until the client decides to populate it. Usually that's in the init method but not necessarily I guess. So I have to run init() on each client before I run client.getCredentialsExtractor() and before I set the extractor to the wrapper. 1) it's not intuitive and 2) maybe not foolproof depending on who wrote the client. Thoughts?

leleuj:
Yes, you need to init() the client to be sure to get a `CredentialsExtractor` and then wrap it. In any case, this is what will be done before pac4j uses the client in one of its filters, so you'll get the same result.

PR #1458, opened 2019-11-21
Handle a spring resource properly because a resource may not be a file

garpinc:
Handle case where resource is not file

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
@@ -15,7 +15,6 @@ import org.opensaml.saml.ext.saml2mdreqinit.RequestInitiator;
 import org.opensaml.saml.metadata.resolver.MetadataResolver;
 import org.opensaml.saml.metadata.resolver.impl.AbstractBatchMetadataResolver;
 import org.opensaml.saml.metadata.resolver.impl.DOMMetadataResolver;
-import org.opensaml.saml.metadata.resolver.impl.FilesystemMetadataResolver;
 import org.opensaml.saml.saml2.core.NameIDType;
 import org.opensaml.saml.saml2.metadata.AssertionConsumerService;
 import org.opensaml.saml.saml2.metadata.AttributeConsumingService;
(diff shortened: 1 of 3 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
```diff
@@ -119,7 +118,7 @@
     public MetadataResolver buildMetadataResolver(final Resource metadataResource) throws Exception {
         final AbstractBatchMetadataResolver resolver;
         if (metadataResource != null) {
-            resolver = new FilesystemMetadataResolver(metadataResource.getFile());
+            resolver = new SpringResourceMetadataResolver(metadataResource);
```
mmoayyed: What is the use case here? Could you explain the scenario that has you define and load SP metadata that is not backed by a file? What specific implementation of the `WritableResource` are you dealing with?

garpinc:
Well the whole idea of spring resource is that it an interface to a resource with different implementations. My particular implementation happens to be a type of object store in a semantic web database.. so when the service provider metadata is written it writes to that object store and generates an associated uri to point to it.



- Garry


-------- Original message --------

leleuj:
I'm hesitating on this: yes, the change is fairly minimal and would address all use cases, but do we really want to address such an unlikely use case?

@mmoayyed what do you think?

garpinc:
I don't think it's an unlikely use case that the file is not stored in a file system. In fact read does work but write currently doesn't. That's the whole reason we used resource in the first place.



- Garry


-------- Original message --------

aeismann:
E.g. in my case: All data that is created during execution has to be stored in mongodb in gridFS, because all filesystems in production are read-only. So a GridFS backed implementation of `WritableResource` could do the trick. But when `getFile` is explicitly used, this cannot not work.

mmoayyed:
The reason the Spring resource class is used is primarily because we didn't want to re-create our own resource class, and not because we wanted to allow for alternative abstractions of resources; that is simply a side-benefit, but nothing to focus on. Furthermore, this is a fairly uncommon use case; as in, it's a lot more likely and common for an identity provider to want to store service metadata somewhere in a database, etc, than it is for a service provider to store its own metadata (or the identity provider's metadata) in a database. This is, for pac4j, demonstrated by the number of issues, pull requests and user activity in this area on the mailing list.

That said, I do like and support the idea of taking advantage of this abstraction but I am not a fan of introducing a whole new metadata-resolver resource by default with background threads and timers to refresh stuff, etc. That's complexity that is unwarranted at this point and is rather unnecessary until further evidence pops up. What would make sense to me, is allowing folks to define the metadata resource abstraction on their own as a pluggable strategy so that one can plug their own implementation into it and deal with object databases or anything else that is supported by that writeable-resource. 

This pull request does a lot more and I think it needs to trim down first.

garpinc:
I submitted a PR that does it with injection instead.

PR #1459, opened 2019-11-22
Handle a spring resource properly because a resource may not be a file.  Inject metadataresolver

garpinc:
Handle a spring resource properly because a resource may not be a file.  Inject metadataresolver leaving the default be the FileMetadataResolver


pac4j-saml/src/main/java/org/pac4j/saml/metadata/MetadataResolverFactory.java
```diff
@@ -0,0 +1,10 @@
+package org.pac4j.saml.metadata;
+
+import org.opensaml.saml.metadata.resolver.impl.AbstractBatchMetadataResolver;
+import org.springframework.core.io.Resource;
+
+public interface MetadataResolverFactory {
+
+	AbstractBatchMetadataResolver getInstance(Resource metadataResource) throws Exception;
```
mmoayyed: Please use the parent interface, `MetadataResolver`, if possible as the return type.

pac4j-saml/src/main/java/org/pac4j/saml/metadata/MetadataResolverFactory.java
```diff
@@ -0,0 +1,10 @@
+package org.pac4j.saml.metadata;
+
+import org.opensaml.saml.metadata.resolver.impl.AbstractBatchMetadataResolver;
+import org.springframework.core.io.Resource;
+
+public interface MetadataResolverFactory {
```
mmoayyed: Please rename the component to be `SAML2MetadataResolverFactory` to keep a consistent naming convention.

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataGenerator.java
```diff
@@ -111,6 +110,8 @@
     protected List<String> signatureReferenceDigestMethods = null;
 
     private List<SAML2MetadataContactPerson> contactPersons = new ArrayList<>();
+    
+    private MetadataResolverFactory metadataResolverFactory = new FileMetadataResolverFactory();
```
mmoayyed: Also add a getter please.

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2ServiceProviderMetadataResolver.java
```diff
@@ -129,17 +132,27 @@ private void writeServiceProviderMetadataToResource(final String tempMetadata) t
         final WritableResource metadataResource = configuration.getServiceProviderMetadataResource();
         if (metadataResource != null && CommonHelper.isNotBlank(tempMetadata)) {
             if (metadataResource.exists() && !configuration.isForceServiceProviderMetadataGeneration()) {
-                logger.info("Metadata file already exists at {}.", metadataResource.getFile());
+                logger.info("Metadata file already exists at {}.", metadataResource.getFilename());
```
mmoayyed: Log the resource, and not the file name. Not all writeable-resource implementations might in fact have a filename, and I suspect may throw an exception.
Also please adjust the log message. The resource may not always be a file as the message says.

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2ServiceProviderMetadataResolver.java
```diff
@@ -129,17 +132,27 @@ private void writeServiceProviderMetadataToResource(final String tempMetadata) t
...
+                            if (!parent.exists() && !parent.mkdirs()) {
+                                logger.warn("Could not construct the directory structure for SP metadata: {}",
+                                       parent.getCanonicalPath());
+                            }   
+                        }
                     }
+                } catch (UnsupportedOperationException e) {
+                    // do nothing since likely a resource that doesn't have a filesystem
+                    logger.warn("no filesystem", e);
                 }
```
mmoayyed: This entire block seems like it should be moved to the file-based resolver.

garpinc: It is not 100% certain that particular resolver does or does not support directories. You can't know this ahead of time. Plus the current file-based resolver isn't pac4j code.

mmoayyed: Understood. Either way, it has to move out of there and go somewhere else. It was semi sensible when we were explicitly using the file-backed resolver; Now that it's abstracted out, the responsibility does not seem like it would belong here. This code is clearly dealing with the file system; so either the code has to move out and find a new home, or ideally, it would have to go into the file-backed resolver (or our extension of it that would own this bit)

garpinc: But the issue is that not only filesystems support directory structures as is evident in the abstraction. An object store may do the same. But I can attempt to move it and see if it makes sense

leleuj:
The new classes should be commented with `@author` and `@since`. I don't understand why the `SpringResourceMetadataResolver` is in this PR.

But above all and sorry for the hard work that has been done, I find the result pretty ugly. So much complexity added.

I'm -1 on this.

When a PR comes in, I see three options:

1) It really makes sense and we should have it in pac4j: we merge it and it is by default
2) Yes, why not? It's a customization that might be useful, let's create some extension point for it: it can be addressed by configuration
3) It seems to be a really rare use case: let's turn the code into something that could make the use case possible.

I think here it's option 3). So my proposal is to add a `createMetadataResolver` method in the `SAML2MetadataGenerator` which returns a new `FilesystemMetadataResolver(metadataResource.getFile())` + a few `null` checks and a `try`/`catch` and nothing more.
This way, you'll be able to override this method with a new class and use your own metadata resolver.

garpinc:
org.pac4j.saml.metadata.SAML2ServiceProviderMetadataResolver.buildMetadataGenerator() hardcodes the SAML2MetadataGenerator instance.. How do you propose utilizing the newly created instance?

leleuj:
So you'll also need to override the `SAML2ServiceProviderMetadataResolver`  class and the `buildMetadataGenerator` method...

garpinc:
@leleuj  turns out this is not enough
In org.pac4j.saml.metadata.SAML2ServiceProviderMetadataResolver.writeServiceProviderMetadataToResource(String) there is
final File parent = metadataResource.getFile().getParentFile();

on a resource that is not from a filesystem getFile() may through an exception or at least a null. I handled that the the previous PR. 

I can't override public final MetadataResolver resolve() since it's final which I would need to do to call prepareServiceProviderMetadata  which calls writeServiceProviderMetadataToResource(String) both methods of which are private

PR #1460, opened 2019-11-22
Allow saml metadata generator to create MDUI elements

mmoayyed:
The motivation for the change comes from the Kantara SAML2 deployment profile where MDUI compliance is a required item for SAML2 SPs. (There are a few other things that I might add, as we cover more ground in the spec).


leleuj:
I'm late on this, but LGTM.

PR #1461, opened 2019-11-22
Travis CI: Build pac4j against JDK 8 and 11 only

mmoayyed:
I don't think it's any longer necessary to build against JDK 9 and 10 as they are both EOLed. JDK 8 and 11 should for now be the canonical JDKs used for CI builds.


leleuj:
OK. It will save some time and power.

PR #1463, opened 2019-11-28
Support refresh grant in OidcClient

sorenvalentinjp:
This feature was discussed here https://groups.google.com/forum/?hl=en#!topic/pac4j-dev/r7As1SYp_To

First of all, this my first contribution to pac4j, so please bear with me if I made some obvious mistakes.

I have not yet found an elegant way to save the renewed profile into the session store, the only way I found it to work so far, was to save the profile from within the profilemanager, but this scope is missing the boolean configurations "saveInSession" and "multiProfile".

Ideally the new profile should be saved from within DefaultSecurityLogic, but I have yet to understand the intricacies of pac4j to figure how to do that. Maybe someone more experienced on the matter could point me in the right direction?

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
@@ -110,6 +110,8 @@ public class ProfileManager<U extends UserProfile> {
                         final Optional<U> newProfile = client.get().renewUserProfile(profile, context);
                         if (newProfile.isPresent()) {
                             profiles.put(key, newProfile.get());
+                            // TODO GET BOOLEANS FROM CONFIG OR MOVE THIS LOGIC ELSEWHERE
+                            save(true, newProfile.get(), false);
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
```diff
@@ -110,6 +110,8 @@ protected void removeOrRenewExpiredProfiles(final LinkedHashMap<String, U> profi
                         final Optional<U> newProfile = client.get().renewUserProfile(profile, context);
                         if (newProfile.isPresent()) {
                             profiles.put(key, newProfile.get());
+                            // TODO GET BOOLEANS FROM CONFIG OR MOVE THIS LOGIC ELSEWHERE
+                            save(true, newProfile.get(), false);
```
leleuj: You raise a good point here: it's a bug. We should have removed the profile when expired.

It's fixed now: https://github.com/pac4j/pac4j/commit/30236ef4d230520d93325df98c44407b9a1bd5c0

Can you remove this change?

sorenvalentinjp: Thank you for your swift response. That was exactly what was needed.

sorenvalentinjp: I tested it by running it in a play framework project, which uses the KeycloakOidcClient.

I set org.pac4j package logging level to trace, then i set the access tokens lifespan to 1 minute in my keycloak instance, and saw the profile getting refreshed over and over again.

I also verified that the profile is removed once both the access token and refresh token has expired.

leleuj: Excellent!

leleuj:
It's merged. Thanks for your contribution.

PR #1472, opened 2019-12-10
Make the exception more helpful in case the entity ID and role can't be found

Vlatombe:
```
Cannot find entity
org.pac4j.saml.metadata.SAML2ServiceProviderMetadataResolver@6d46256d or
role {urn:oasis:names:tc:SAML:2.0:metadata}SPSSODescriptor in metadata
provider
```


Vlatombe:
The exception I'm actually changing through this variable rename (which IMO makes the code clearer) is https://github.com/pac4j/pac4j/pull/1472/files#diff-b255b4c7964129ab7bcf1ea9ef6cd021R133

leleuj:
Indeed, the `SAML2MetadataResolver` is badly named.

PR #1502, opened 2020-01-31
Initial pass at boot autoconfig

### LINKED ISSUE #717: Pac4J + Spring Boot?
https://docs.google.com/document/d/1k-Cw8QiRHVA91nI-rww4Bh1ealU-qa0avBmUbDCofqQ/edit#heading=h.kdpv4vsgwf36

mmoayyed:
Closes #717

### DIFF
--- a/documentation/docs/springboot.md
+++ b/documentation/docs/springboot.md
@@ -41,3 +41,37 @@ The following auto-configuration classes are provided:
 | `pac4j.twitter.id`                                   |
 | `pac4j.twitter.secret`                               |
 
+### GitHub
+
+- `GitHubClientAutoConfiguration`
+
+| Property                                             |
+|------------------------------------------------------|
+| `pac4j.github.id`                                   |
+| `pac4j.github.secret`                               |
+
+### Google
+
+- `GoogleClientAutoConfiguration`
+
+| Property                                             |
+|------------------------------------------------------|
+| `pac4j.google.id`                                   |
+| `pac4j.google.secret`                               |
+
+### OAuth2
+
+- `OAuth20ClientAutoConfiguration`
+
+| Property                                             |
+|------------------------------------------------------|
+| `pac4j.oauth2.id`                                   |
+| `pac4j.oauth2.secret`                               |
+| `pac4j.oauth2.authUrl`                               |
+| `pac4j.oauth2.tokenUrl`                               |
+| `pac4j.oauth2.profileUrl`                             |
+| `pac4j.oauth2.profilePath`                            |
+| `pac4j.oauth2.profileId`                               |
+| `pac4j.oauth2.scope`                                   |
+| `pac4j.oauth2.withState`                               |
+| `pac4j.oauth2.clientAuthenticationMethod`              |
--- a/documentation/docs/springboot.md
+++ b/documentation/docs/springboot.md
@@ -75,3 +75,48 @@ The following auto-configuration classes are provided:
 | `pac4j.oauth2.scope`                                   |
 | `pac4j.oauth2.withState`                               |
 | `pac4j.oauth2.clientAuthenticationMethod`              |
+
+### OIDC
+
+- `OidcClientAutoConfiguration`
+
+| Property                                              |
+|-------------------------------------------------------|
+| `pac4j.oidc.type`                                     |
+| `pac4j.oidc.id`                                       |
+| `pac4j.oidc.secret`                                   |
+| `pac4j.oidc.scope`                                    |
+| `pac4j.oidc.discoveryUri`                             |
+| `pac4j.oidc.useNonce`                                 |
+| `pac4j.oidc.preferredJwsAlgorithm`                    |
+| `pac4j.oidc.maxClockSkew`                             |
+| `pac4j.oidc.clientAuthenticationMethod`               |
+| `pac4j.oidc.customParamKey`                           |
+| `pac4j.oidc.customParamValue`                         |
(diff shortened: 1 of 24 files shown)


leleuj:
@mmoayyed keep me posted when this will be good for merging...

mmoayyed:
I will probably finalize this by Monday. I would like to build support for oauth, oidc and saml2 clients and leave the rest for now to see if we can gather community interest a bit to add support for more via contributions.

leleuj:
Is this meant to eventually replace the `DelegatedClientFactory` in CAS?

mmoayyed:
Not quite, as that would be A LOT of work. But if the support here matures enough and gains interest from others to build and support other clients or enhance existing ones for Spring Boot, we could in the future shoot for a replacement.

leleuj:
I'm certainly missing the point, but I was not expected this solution. I thought you would have extracted the properties in a Spring Boot auto configuration class and have delegated the work to the `PropertiesConfigFactory`.
This factory already supports all the clients. I think it would be a lot easier.

mmoayyed:
> I thought you would have extracted the properties in a Spring Boot auto configuration class and have delegated the work to the PropertiesConfigFactory.

If you meant for this in a CAS context, that would certainly be ideal and perhaps an ultimate goal but I am not going to do that at the moment. I do like the idea and concept in general; At the moment, I do not see a lot of gains from doing so and I would prefer for the change to provide more than just refactoring and code-clean up.

My proposal is a bit different; let's forget CAS for a second: 

1. Spring Boot is a popular platform; many Java developers build apps on top of it.
2. Pac4j is a solid library for authn/authz; Spring Boot developers would want to use it in their own apps
3. Let's build support for SB directly in Pac4j to auto-configure clients based on properties using the SB model.
4. Let's wait and see who might be interested in using, contributing and maintaining this. (...which might be CAS dev, or a few other projects I can think of)
.
.
.
4004. Things are near-perfect, complete and used by the community
4005. Allow systems like CAS to use the native functionality with SB and Pac4j and clean up code there.

mmoayyed:
This PR, could for example, remain in WIP/Draft mode for quite some time to see if we run into anyone interested in this sort of functionality. We could then point them to this PR and gauge their interest. This PR is likely a good foundation for future development in this area, so long as there are folks interested to maintain, use and contribute to it. Let's find those before we spend more time on this.

leleuj:
I didn't mean it for CAS context. My point is to avoid work by using the `PropertiesConfigFactory`.

You have the `Pac4jConfigurationProperties`:

```java
public class Pac4jConfigurationProperties {
    public static final String PREFIX = "pac4j";

    private Saml2Properties saml2 = new Saml2Properties();

    private CasProperties cas = new CasProperties();
```

Why not turning it into something like this?

```java
public class Pac4jConfigurationProperties {

     private Map<String, String> properties;
```

Get all the properties, build the `PropertiesConfigFactory` and call `build` to get the `Config`.

We would not add `CasClient`, `FacebookClient`, ... as Spring beans but just the `Config` which has all clients inside it.

As I said, I'm certainly missing the point...

mmoayyed:
Thanks very much for the notes; let me explain better:

> Get all the properties, build the PropertiesConfigFactory and call build to get the Config.

Absolutely. This is quite easy to do and requires the least amount of work. However, some possible issues:

1. Collecting settings into a general-purpose map/container of some kind does not allow for type-safety at compile and/or runtime. 
2. There is no encapsulation, as one can in theory stuff anything into the map.
3. Using a general purpose map loses on the ability to use the Spring Metadata Processor/Generator.
4. Not having dedicates types for settings will, at some point, force developers to inject the map into other components to get direct access to a setting. Which means, they will have to refer to a setting by its String name (key) rather than a well-defined Getter, which makes maintenance very very difficult if we were to remove/rename settings. Such changes are usually invisible.
5. A general purpose map prevents Spring Boot to run validation events on names/values (i.e. prevent against unknown or badly formatted valued)

So indeed, my approach here requires a lot more work but it's argubly safer. That said, if you feel like we should continue with the factory as proposed, that would be just fine. We should know the alternatives, and then make a decision.

leleuj:
OK. This is very clear. Your approach is indeed safer. 

My feeling though is that what you propose is too ambitious and maybe unnecessary in most cases. I would go for the basic implementation only based on the `PropertiesConfigFactory`.

That said, I prefer to have a real target for a new implementation and I have always had the feeling that the CAS pac4j configuration should be fully delegated to pac4j if possible.

So for you, is the global `PropertiesConfigFactory` solution unacceptable to be used in CAS for pac4j configuration?

mmoayyed:
> So for you, is the global PropertiesConfigFactory solution unacceptable to be used in CAS for pac4j configuration?

Yes, I'd say so, for all the above reasons and not to mention that moving from "cas.authn.pac4j" as the configuration namespace to something like "pac4j.clients.xyz" is a major breaking change, and one that we wouldn't be able to successfully do without a major release...and perhaps with caution even then.

So for now, I can rework this pull request to be based on the `PropertiesConfigFactory` and we can then give this some time to brew and collect feedback. With time, things might improve/change for the better one way or another. Would this work?

leleuj:
Sure. Let's do that. Let's start with as basic solution based on the `PropertiesConfigFactory` and see how where it goes...

mmoayyed:
Sounds good. Will do.

mmoayyed:
Closing in favor of https://github.com/pac4j/pac4j/pull/1550

PR #1503, opened 2020-01-31
Current implementation does not support selected SP assertion consumer service different from the default one

fmartelli:
The issue is at the following line: provided assertion consumer service index is ignored.

https://github.com/pac4j/pac4j/blob/master/pac4j-saml/src/main/java/org/pac4j/saml/profile/impl/AbstractSAML2MessageReceiver.java#L59

Looking forward for a feedback.
Regards


fmartelli:
Some errors in my repository.
Let me close this and open again

leleuj:
Sure.

PR #1510, opened 2020-02-05
Permits to ask for a response mandatory signed

fmartelli:
Add and handle a configuration parameter to be used to ask for a response mandatory signed

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
@@ -143,6 +143,7 @@ public class SAML2Configuration extends InitializableObject {
     private List<String> signatureReferenceDigestMethods;
     private String signatureCanonicalizationAlgorithm;
     private boolean wantsAssertionsSigned = false;
+    private boolean wantsResponseSigned = false;
     private boolean allSignatureValidationDisabled = false;
 
     private String keyStoreAlias;
@@ -620,6 +621,14 @@ public class SAML2Configuration extends InitializableObject {
 
     public void setWantsAssertionsSigned(final boolean wantsAssertionsSigned) {
         this.wantsAssertionsSigned = wantsAssertionsSigned;
+    }
+
+    public boolean isWantsResponseSigned() {
+        return wantsResponseSigned;
+    }
+
+    public void setWantsResponseSigned(final boolean wantsResponseSigned) {
+        this.wantsResponseSigned = wantsResponseSigned;
     }
 
     public boolean isAuthnRequestSigned() {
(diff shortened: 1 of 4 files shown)


fmartelli:
Hi, build with jdk11 fails but it shouldn't be up to me ...
Please, let me know if I missed something.

Regards.

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnResponseValidator.java
```diff
@@ -195,6 +212,10 @@ protected void validateSamlProtocolResponse(final Response response, final SAML2
 
         validateSuccess(response.getStatus());
 
+        if (wantsResponseSigned && response.getSignature() == null) {
+            throw new SAMLSignatureValidationException("Signature not found");
```
mmoayyed: You should likely explain the root cause better here; something like:

```
Unable to find a signature on the SAML response returned. Pac4j is configured to enforce signatures on SAML2 responses from identity providers and the returned response {} does not contain a signature
```

Then log the response (by encoding it as XML first)

pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
```diff
@@ -143,6 +143,7 @@
     private List<String> signatureReferenceDigestMethods;
     private String signatureCanonicalizationAlgorithm;
     private boolean wantsAssertionsSigned = false;
+    private boolean wantsResponseSigned = false;
```
mmoayyed: Same as assertions, please rename to "wantsRespons**es**Signed"

pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnResponseValidator.java
```diff
@@ -212,8 +215,21 @@ protected void validateSamlProtocolResponse(final Response response, final SAML2
...
-            throw new SAMLSignatureValidationException("Signature not found");
+        if (wantsResponsesSigned && response.getSignature() == null) {
+            if (logger.isDebugEnabled()) {
+
+                try {
+                    logger.debug(
+                            "Unable to find a signature on the SAML response returned. Pac4j is configured to enforce "
+                            + "signatures on SAML2 responses from identity providers and the returned response\n{}\n"
+                            + "does not contain any signature",
+                            SerializeSupport.prettyPrintXML(XMLObjectSupport.marshall(response)));
```
mmoayyed: You should 

1- Use this https://github.com/pac4j/pac4j/blob/master/pac4j-saml/src/main/java/org/pac4j/saml/util/Configuration.java#L97

or 

2- Change that method to use your solution above instead.

PR #1511, opened 2020-02-06
Fixes possible serialization problems putting object into the session…

fmartelli:
Serialize XML objects before putting them into the session store

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
@@ -1,6 +1,5 @@
 package org.pac4j.saml.store;
 
-import java.io.StringReader;
 import org.opensaml.core.xml.XMLObject;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.util.CommonHelper;
@@ -9,7 +8,6 @@ import org.slf4j.LoggerFactory;
 
 import java.util.LinkedHashMap;
 import java.util.Optional;
-import org.opensaml.core.xml.util.XMLObjectSupport;
 import org.pac4j.saml.util.Configuration;
 
 /**
@@ -102,14 +100,7 @@ public class HttpSessionStore implements SAMLMessageStore {
         messages.clear();
         updateSession(messages);
 
-        try {
-            XMLObject message = XMLObjectSupport.unmarshallFromReader(
-                    Configuration.getParserPool(), new StringReader(o));
-            return Optional.of(message);
-        } catch (Exception e) {
-            log.error("Error unmarshalling message from input stream", e);
-            return Optional.empty();
-        }
+        return Configuration.deserializeSamlObject(o);
     }
 
     /**
--- a/pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
@@ -1,5 +1,6 @@
 package org.pac4j.saml.store;
 
+import java.io.StringReader;
 import org.opensaml.core.xml.XMLObject;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.util.CommonHelper;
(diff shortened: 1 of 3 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
```diff
@@ -34,7 +37,7 @@
     /**
      * Internal store for messages, corresponding to the object in session.
      */
-    private LinkedHashMap<String, XMLObject> internalMessages;
+    private LinkedHashMap<String, String> internalMessages;
```
mmoayyed: I realize this has always been here, but is there a reason LinkedHashMap is required? Could we just use the usual HashMap instead?

pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
```diff
@@ -86,17 +89,27 @@ public void set(final String messageID, final XMLObject message) {
...
     @Override
     public Optional<XMLObject> get(final String messageID) {
-        final LinkedHashMap<String, XMLObject> messages = getMessages();
-        final XMLObject o = messages.get(messageID);
+        final LinkedHashMap<String, String> messages = getMessages();
+        final String o = messages.get(messageID);
         if (o == null) {
-            log.debug("Message {} not found in session {}", messageID, context.getSessionStore().getOrCreateSessionId(context));
+            log.debug("Message {} not found in session {}",
+                    messageID, context.getSessionStore().getOrCreateSessionId(context));
```
mmoayyed: Not sure you need to log `context.getSessionStore().getOrCreateSessionId(context)`, as that might actually create a session id? @leleuj would know best.

pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
```diff
@@ -86,17 +89,27 @@ public void set(final String messageID, final XMLObject message) {
...
-        return Optional.of(o);
+
+        try {
+            XMLObject message = XMLObjectSupport.unmarshallFromReader(
+                    Configuration.getParserPool(), new StringReader(o));
+            return Optional.of(message);
+        } catch (Exception e) {
+            log.error("Error unmarshalling message from input stream", e);
+            return Optional.empty();
+        }
```
mmoayyed: You should likely move these lines into a method next to one in Configuration that does the serialization.

pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
```diff
@@ -86,17 +89,27 @@ public void set(final String messageID, final XMLObject message) {
...
-        log.debug("Message {} found in session {}, clearing", messageID, context.getSessionStore().getOrCreateSessionId(context));
+        log.debug("Message {} found in session {}, clearing",
+                messageID, context.getSessionStore().getOrCreateSessionId(context));
         messages.clear();
         updateSession(messages);
-        return Optional.of(o);
+
+        try {
+            XMLObject message = XMLObjectSupport.unmarshallFromReader(
+                    Configuration.getParserPool(), new StringReader(o));
```
mmoayyed: Is the StringReader expected to be closed by the caller(pac4j)?

pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
fmartelli: Good question!

pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
fmartelli: I left the log statement as it was.
Please let me know if you prefer to change that line

leleuj: Indeed, the `getOrCreateSessionId` method could create a session if it does not exist.
Given it's in the webflow of the SAML login process, it's unlikely though.
I think we can keep these for now.

pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
leleuj: I guess that there might be some performance idea behind that: messages are saved and should be retrieved in order...

fmartelli: thanks

pac4j-saml/src/main/java/org/pac4j/saml/store/HttpSessionStore.java
fmartelli: thanks

PR #1514, opened 2020-02-10
Update mongodb.md

dmaciaszek:
Everywhere is `<artifactId>pac4j-mongo</artifactId>` instead of `<artifactId>pac4j-mongodb</artifactId>`


leleuj:
Good catch! Thanks

PR #1521, opened 2020-02-20
Profile service read password attribute when validate

### LINKED ISSUE #1520: AbstractProfileService ignore password attribute when read for validate
In AbstractProfileService validate function, list of attributes to read use default PASSWORD attribute instead of password attribute set by user.
https://github.com/pac4j/pac4j/blob/991d46112d1d7294c6429a7166a69725616b62f5/pac4j-core/src/main/java/org/pac4j/core/profile/service/AbstractProfileService.java#L280-L290

ducthienbui97:
fix #1520


leleuj:
Good catch! Thanks for your contribution

PR #1532, opened 2020-02-26
Remove unused dependency to cryptacular

daniel-beck:
This dependency appears unused: It was unused when it was added in https://github.com/pac4j/pac4j/commit/6178f28136b55e4fe78ef749621276db993bf4d1, and seems to have remained unused since, despite being updated several times. `mvn clean verify` passes with this change. No relevant looking mentions on users or developers mailing list either.

I discovered this when I investigated a dependency scanner's complaints about an outdated transitive dependency, and thought I can submit what I used it to confirm this is unused.

I'm not subscribing to a new mailing list of a library I'm not even using to discuss this first, sorry. Please close this PR if that is unacceptable.


daniel-beck:
As removing the dependency on `cryptacular` broke something in a dependent component, this might not be a good idea after all. Sorry for the noise.

PR #1576, opened 2020-04-24
Default profile id, if undefined, for OAuth clients

mmoayyed:
Also fixes a number of typos, log parameters, and other minor issues.


leleuj:
Yes, it makes sense to have a default value for the `profileId`.

leleuj:
As the `profileId` is never `null`, I made this additional change: https://github.com/pac4j/pac4j/commit/09033edd05cf38ae0cc2759f8ca2f629b7b843a4

PR #1585, opened 2020-04-30
Avoid casting to LinkedHashMap for profile manager

mmoayyed:
Removes the expectation that profiles stored in session/request must be managed via a LinkedHashMap. Instead, any `Map` should be OK and it should be up to the caller to decide what `Map` type should be used.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/ProfileManager.java
@@ -115,14 +115,10 @@ public class ProfileManager<U extends UserProfile> {
     protected LinkedHashMap<String, U> retrieveAll(final boolean readFromSession) {
         final LinkedHashMap<String, U> profiles = new LinkedHashMap<>();
         this.context.getRequestAttribute(Pac4jConstants.USER_PROFILES)
-            .ifPresent(requestAttribute -> {
-                profiles.putAll((LinkedHashMap<String, U>) requestAttribute);
-            });
+            .ifPresent(requestAttribute -> profiles.putAll((Map<String, U>) requestAttribute));
         if (readFromSession) {
             this.sessionStore.get(this.context, Pac4jConstants.USER_PROFILES)
-                .ifPresent(sessionAttribute -> {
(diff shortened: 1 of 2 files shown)


leleuj:
Can you elaborate a bit on why we need that change? You directly put profiles in the request/session for further usage?

mmoayyed:
Yes, indeed. Although, not that I am doing that directly but for example, there are A LOT of tests in CAS where the session is populated with a profile first and then some OAUTH/OIDC component is invoked to verify operations. 

That aside, I am not sure why the cast has to be explicitly a LinkedHashMap. I think Map should be enough.

PS Just a side note, the behavior of the profile-manager is changed between 4.0 and 4.0.1 it seems where the manager can no longer recognize a single profile (which seems it was deprecated behavior. This should be called out in the release notes as it's a difficult-to-catch change and non-obvious on the first try for a patch release.

mmoayyed:
Example:

```java
        val request = new MockHttpServletRequest();
        val response = new MockHttpServletResponse();

        val profile = new CommonProfile();
        profile.setClientName(Authenticators.CAS_OAUTH_CLIENT_BASIC_AUTHN);
        profile.setId(RequestValidatorTestUtils.SUPPORTING_CLIENT_ID);
        val session = request.getSession(true);
        assertNotNull(session);
        session.setAttribute(Pac4jConstants.USER_PROFILES, <profile-should-go-into-any-kind-of-map>);

        request.setParameter(OAuth20Constants.GRANT_TYPE, OAuth20GrantTypes.AUTHORIZATION_CODE.getType());
        request.setParameter(OAuth20Constants.REDIRECT_URI, RegisteredServiceTestUtils.CONST_TEST_URL);
        request.setParameter(OAuth20Constants.CODE, SUPPORTING_SERVICE_TICKET);

        assertTrue(validator.validate(new JEEContext(request, response)));
```

leleuj:
The `LinkedHashMap` was meant to keep the order of the added profiles, but I don't think we really use it. Maybe we'll need to change that in the future.

About the deprecated behavior, it's indeed in the release notes, but it was changed in version 3.x in fact (multi-profile) so I should have changed it in version 4.0.0, but I forgot about it.

I understand your rationale. Merging...

PR #1590, opened 2020-05-06
additional changes to SAML2 metadata/keystore generators

mmoayyed:
Testing recent keystore/metadata generation changes, as part of: https://github.com/apache/syncope/pull/181 and making small adjustments to allow for customizations.


leleuj:
Is this ready for review?

mmoayyed:
Almost ready :) Will be doing some more tests tomorrow and then should be done (and ready for 4.0.1 too)

leleuj:
No problem. Just wanted to know if it goes into v4.0.1 (or v4.0.2).

mmoayyed:
This should be ready. Tested with Syncope+CAS and all works as expected.

mmoayyed:
Ready to merge with a passing build. Once merged, I'll push the change into a new CAS snapshot and finalize changes with Apache Syncope.

pac4j-core/src/main/java/org/pac4j/core/client/config/BaseClientConfiguration.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Base parent class for all client configurations
+ * to hold common fields or to be used as a common parent
+ * for type checking and casts in customizations.
+ *
+ * @author Misagh Moayyed
+ * @since 4.0.1
+ */
+public abstract class BaseClientConfiguration extends InitializableObject {
```
leleuj, line 13: I don't see any fields yet in this common class. Missing something?

mmoayyed: I didn't find a common fields for configurations, so I left it blank. My intention is to be able to use instanceof on configuration classes that are, per contract, expected to be of type `BaseClientConfiguration`, rather than a very generic `InitializableObject`. This is more of nice-to-have.

leleuj: I think the `logoutHandler` might fall into that category. Anyway for now. Merging.

PR #1626, opened 2020-06-18
Implemented capability to use custom State validation for CSRF mitigation

martinhansen:
Added a `StateValidator` interface, that we set in the `OidcConfiguration`, and the `OidcExtractor` will then use to validate the state of a callback request.

In a perfect world, i would have implemented it as a CSRF mitigation strategy instead. An interface that had both `generateState` and `validateState` methods, but i have chosen this simplistic extension to have backwards compatibility, in the least intrusive manner.

I can implement the other version if you'd like, but with some more if-cases in the `OidcRedirectionActionBuilder.addStateAndNonceParameters()` methods, without breaking compatibility. The code in general would seem more muddy in my opinion though, with that approach.

I have also thought about implementing a `enableCSRFMitigation(stateGenerator, stateValidator)` method in the `OidcConfiguration` class. A method that sets the `withState` property, and both `stateGenerator` and `stateValidator`, also you call if that's something we would like.

Example usage (with a HMAC stateless validator)
https://gist.github.com/martinhansen/3c5d64d0dbe1a610fdb9ec9113e04053

dev group thread:
https://groups.google.com/forum/?fromgroups#!topic/pac4j-dev/bIwjRNBe-eg

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
@@ -25,6 +25,8 @@ import com.nimbusds.oauth2.sdk.ResponseType;
 import com.nimbusds.oauth2.sdk.auth.ClientAuthenticationMethod;
 import com.nimbusds.openid.connect.sdk.OIDCResponseTypeValue;
 import com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata;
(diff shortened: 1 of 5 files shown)


leleuj:
Great!

Indeed, it raises several questions: should we merge the generation and validation? We could also turn that into something more generic (not depending from the OIDC `State`) at the `pac4j-core` level.

Let's merge it "as is" for now.

PR #1630, opened 2020-06-22
PKCE support, better handling of expired tokens and some minor improvements

papegaaij:
This PR implements PKCE support, as discussed on the Google groups. It also improves the handling of expired access tokens. The new StateValidator interface is replaces by ValueRetriever and used for the PKCE implementation. As this PR implements several enhancements, it's best to review the commits one by one.


papegaaij:
I did some minor cleanups and fixed the last remaining tests to make the build pass.

leleuj:
@papegaaij is the PR ready for review?

leleuj:
It looks good to me. Did you test it with and without PKCE and state?

papegaaij:
@leleuj Our IdP supports PKCE and the current implementation does not allow disabling PKCE when the IdP supports it (null means autodetect). I can add a boolean flag to disable PKCE entirely if needed, but personally I would never disable PKCE when the IdP supports it, because it is highly recommended to use it. I did run quite a few tests with our test suite, with different configurations and different flows. That's how I found the NPE and noticed the state parameter with 'withState' set to false.

leleuj:
Yes, please add a flag enabled by default but which allows to disable PKCE. Thx

papegaaij:
I've implemented it the other way around, rather then enabling PKCE, you can disable it. I think this makes more sense. Especially when the IdP does not support PKCE it would not have been clear what the behavior would have been if pkce was enabled.

papegaaij:
Btw, I did test without PKCE, but didn't notice it :) My stub IdP does not support PKCE, so I can confirm this works as expected, the parameters were not added.

The nice think about PKCE is that the client only needs to add certain parameters. You can even add these parameters when the IdP does not support PKCE, they will simply be ignored. All validation is done on the side of the IdP. In fact, when using PKCE, both the nonce and state add very little value. PKCE is much stronger.

leleuj:
Excellent! Will merge. I'll make some tests today and release tomorrow or the day after tomorrow.

PR #1634, opened 2020-06-26
pac4j-Kerberos bug fix for DirectKerberosClient example documentation

FredrikBakken:
Minor bug fix to the import of the pac4j-kerberos example documentation.


leleuj:
Thanks

PR #1654, opened 2020-07-20
Allow a class with a byte array member to be serialized

### LINKED ISSUE #340: Initialize CAS logout by default
Currently, the CAS logout handler is by default the No op one. We should use the https://github.com/pac4j/pac4j/blob/master/pac4j-cas/src/main/java/org/pac4j/cas/logout/CasSingleSignOutHandler.java if no logout handler is defined and the context is a `J2EContext`.

daddykotex:
I did not add tests because, currently, the test suites does not have a test for every single trusted package so I assume you don't want that.

See: https://github.com/pac4j/play-pac4j/issues/340


leleuj:
Thanks for your contribution

PR #1660, opened 2020-07-29
Test improvement

eas5:
This is a test refactoring.

**Problem:**
The Exception Handling test smell occurs when a test method explicitly a passing or failing of a test method is dependent on the production method throwing an exception.

**Solution:**
Use JUnit's exception handling to automatically pass/fail the test instead of writing custom exception handling code or throwing an exception.

**Result:**
_Before:_
```
@Test
public void testUnsafeAddAttribute() {
    final CommonProfile userProfile = new CommonProfile();
    try {
        userProfile.getAttributes().put(KEY, VALUE);
    } catch (final UnsupportedOperationException e) {
        fail();
    }
}
```
_After:_
```
@Test
public void testUnsafeAddAttribute() throws UnsupportedOperationException {
    final CommonProfile userProfile = new CommonProfile();
    userProfile.getAttributes().put(KEY, VALUE);
}
```

### DIFF
--- a/pac4j-core/src/test/java/org/pac4j/core/profile/CommonProfileTests.java
+++ b/pac4j-core/src/test/java/org/pac4j/core/profile/CommonProfileTests.java
@@ -92,12 +92,13 @@ public final class CommonProfileTests implements TestsConstants {
     }
 
     @Test
-    public void testUnsafeAddAttribute() {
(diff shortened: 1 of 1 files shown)


leleuj:
Thanks for your contribution

PR #1662, opened 2020-08-01
API changes for ldaptive v2.

dfish3r:
Note that ldaptive v2 requires Java 11 so you may not be able to make use of this anytime soon. But I didn't want to throw this away.

Notable changes:
 - try/finally paradigm no longer needed for operations
 - useSsl property no longer exists; use the LDAPS:// URL scheme
 - providers no longer exist

Let me know if you have any questions.

### DIFF
--- a/pac4j-config/src/main/java/org/pac4j/config/builder/LdapAuthenticatorBuilder.java
+++ b/pac4j-config/src/main/java/org/pac4j/config/builder/LdapAuthenticatorBuilder.java
@@ -1,6 +1,6 @@
 package org.pac4j.config.builder;
 
-import org.ldaptive.pool.PooledConnectionFactoryManager;
+import org.ldaptive.ConnectionFactoryManager;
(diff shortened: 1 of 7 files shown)


mmoayyed:
This is great! Jérôme, when do you think we might be able to switch to JDK 11? or perhaps could we try to consider multi-release JARs somehow?

leleuj:
@dfish3r thanks for the update

Things are complicated now between the nimbus-jose-jwt library which only supports Java 7 and OpenSAML or Ldaptive which require Java 11 ;-)

My current roadmap is: v4.1 in T4 2020 and v5 in T1 - T2 2021.
Version 5 should certainly be based on JDK 11.
Maybe I'll also skip v4.1.

leleuj:
I just released pac4j v4.1, started the development of pac4j v5 based on JDK 11 and merged this PR. Thanks @dfish3r

PR #1681, opened 2020-08-28
Support custom params for callback url resolvers

mmoayyed:
## Problem

At run time and after the client initialization phase, consider the following snippet:

```java
client.getConfiguration().addCustomParam("name", "value");
```

Using this setup, if a client builds a redirect action later to calculate the callback url, the custom parameter is not recognized and added to the url. While custom parameters typically are set up during the initialization phase, they can also be dynamically updated/managed to dictate certain behavior to the identity provider. The present functionality is rather limiting, because in certain scenarios, custom parameters are unknown until much later in the execution sequence.

## Proposal

This pull request allows callback url resolvers to be aware of changes applied to custom parameters. 

Concretely,

- Moves the custom params collection to the parent `BaseConfiguration` class, allowing all support clients to take advantage of custom params.
- Supporting clients are now asked to implement `ConfigurableClient` to indicate that they expose a configuration object. 
- A parent `BaseCallbackUrlResolver.java` is now added that allows its children to compute the final url based on given custom params.

For this to properly work, all callback url resolvers are expected to initialize using a reference to the config object they expose. 

For example,

```java
        var configurableClient = ConfigurableClient.class.cast(client);

        switch (props.getCallbackUrlType()) {
            case PATH_PARAMETER:
                client.setCallbackUrlResolver(new PathParameterCallbackUrlResolver(configurableClient.getConfiguration()));
                break;
            case NONE:
                client.setCallbackUrlResolver(new NoParameterCallbackUrlResolver(configurableClient.getConfiguration()));
                break;
            case QUERY_PARAMETER:
            default:
                client.setCallbackUrlResolver(new QueryParameterCallbackUrlResolver(configurableClient.getConfiguration()));
        }
```

cc: @astohn

### DIFF
--- a/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/client/CasClient.java
@@ -4,6 +4,7 @@ import org.pac4j.cas.authorization.DefaultCasAuthorizationGenerator;
 import org.pac4j.cas.config.CasConfiguration;
 import org.pac4j.cas.credentials.authenticator.CasAuthenticator;
 import org.pac4j.cas.credentials.extractor.TicketAndLogoutRequestExtractor;
+import org.pac4j.core.client.ConfigurableClient;
 import org.pac4j.core.http.callback.CallbackUrlResolver;
 import org.pac4j.core.logout.handler.LogoutHandler;
 import org.pac4j.cas.redirect.CasRedirectionActionBuilder;
@@ -32,7 +33,7 @@ import org.pac4j.core.util.CommonHelper;
  * @author Jerome Leleu
  * @since 1.4.0
  */
-public class CasClient extends IndirectClient<TokenCredentials> {
+public class CasClient extends IndirectClient<TokenCredentials> implements ConfigurableClient<CasConfiguration> {
 
     private CasConfiguration configuration = new CasConfiguration();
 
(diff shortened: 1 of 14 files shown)


mmoayyed:
@leleuj have you had a chance to review this?

leleuj:
I understand your concern: currently, the configuration is highly static and we could need to dynamically update it.

Your PR is good and theoretically, it makes sense.
Though practically I'm a little reluctant on this. I fear people will use it to change the configuration for a new request and this will fail because the config is global for all new concurrent requests.
I remember fixing a bug in CAS related to authentication delegation for the same reason.

Where do you want to use this change?

mmoayyed:
This is related to an issue in CAS and delegated authentication, where the external provider is a CAS server itself.

The flow is:

1. Receive authn request from application
2. Store the context/service/etc in the registry with ticket (session) identifier T (request parameter)
3. Redirect to external CAS server for authn with ticket (session) identifier as T (request parameter)
4. Receive service ticket from external CAS server to validate
5. Calculate callback (service URL) for service ticket validation.
6. Validation fails since the callback service URL does not match the service URL in #3 (since the callback URL resolver does not consider T)

In this solution, T is kept as a request parameter for CAS configuration object assigned to the client, so that its callback URL resolver can calculate the URL correctly.

I understand the issues with concurrency. What's a better way to solve this issue?

leleuj:
Indeed, the current `CasClient` is too static with one pre-defined callback URL.

This is why I created the `DirectCasClient` (https://github.com/pac4j/pac4j/blob/master/pac4j-cas/src/main/java/org/pac4j/cas/client/direct/DirectCasClient.java). It uses the current URL as the service URL (after removing the ticket parameter) for CAS validation.

If you receive the T parameter in step 4, it can work I think.

mmoayyed:
Thank you Jérôme. I will provide an update shortly.

mmoayyed:
Haven't found the time to pursue this further, but it's still on the todo list.

mmoayyed:
Diagnosed and resolved issue. Closing for now.

PR #1718, opened 2020-10-28
NullPointerException due to refresh response not returning an ID-Token

BenjaminBSI:
As discussed on the Mailing list, one can run into a NullPointer when refreshing, as the response doesn't need to send an ID-Token .

According to the specification [here](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse), the response isn't required to return an ID-Token and I think pac4j should reflect this.


leleuj:
Thanks for your contribution.

PR #1735, opened 2020-11-16
GitHub actions

HanSolo:
Added a github action to run mvn verify on each push on Ubuntu 18.04 and latest MacOS using JDK 11

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+---
+name: Java CI
+
+on: [push]
+
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


mmoayyed:
Could you modify the existing Travis CI build to do the same? Pac4j is based on Travis CI. It would be ideal to not have to maintain two separate builds.

leleuj:
Exactly, we cannot maintain two build systems. You must also replicate the logic already used in Travis: https://github.com/pac4j/pac4j/tree/master/travis
And we should eventually do the same for all pac4j implementations...

leleuj:
Closing due to inactivity.

PR #1778, opened 2021-01-11
pac4j-saml: Define dedicated class for SAML2 profile definitions

mmoayyed:
This pull request mainly refactors the saml2 profile definition construct into a dedicated class. The intention is to make this component more obvious and discoverable for extensions, in javadocs and from an API point of view vs what exists now as a lambda expression. It should make it easier to extend this class for custom functionality, while keeping certain sane default in the future, as we add support for certain attribute profiles (such as the eduPerson schema)


leleuj:
There is a PMD issue in the build...

mmoayyed:
Sorry about that; Cleaned up the PMD issue.

PR #1802, opened 2021-02-04
relax saml2 validation rules to skip exceptions

mmoayyed:
This pull request,

- Relaxes the `SAML2LogoutValidator` to not throw back exceptions with an extra option, in scenarios where the logout response is successfully validated. By default, the general expectation is that such exceptions/actions need to be caught and adapted. This PR provides an option to remove the catch requirement, but does keep the default behavior.

- This is purely for extensions of pac4j SAML. On the CAS side, we will need to do some minor work to actually catch such exceptions and adapt them accordingly.

- This PR also fixes a few issues where JDK9+ APIs were used to build pac4j v4.

I will also post a separate pull request for master/v5 mostly cherry-picking from here.


leleuj:
LGTM.

Once again, the Travis build is broken. It's not related to this PR.
Since a few weeks/months, Travis does not work very well. I certainly should move to Github actions...

Merging anyway. Please submit the same PR for the master.

PR #1804, opened 2021-02-05
port saml changes over to master branch

mmoayyed:
Porting #1802 to master.


leleuj:
I'm working on migrating from Travis to Github actions. You have 1 PMD violation ;-)

mmoayyed:
Darn! Thanks, I'll fix it. I suppose I can connect IDEA to use the PMD plugin to catch these, right?

leleuj:
Good question. I just run the build to perform the PMD check, but I suppose yes...

mmoayyed:
Should be fixed now, I hope.

By the way kudos on GH actions. Side suggestion: let's make sure deploy depends on build. Otherwise broken stuff might get published.

leleuj:
Yes, I realized my mistake on the missing dependency between my two jobs.

It works so much better than Travis... for now...

PR #1808, opened 2021-02-06
Additional SAML enhancements

mmoayyed:
This pull request contains the following set of changes, all of which generally improve or enhance the pac4j-saml functionality:

- Signing operations for metadata can now be done using the existing default method, or via XMLSec. Using XMLSec generally tends to work better with strict validators specially in govt-related projects.
- Additional checks to validate requests/responses.
- Ability to specify the SLO url in the configuration and metadata.
- Options to determine how to URLs should be compared when doing endpoint verifications.
- Logout validation can be given an expected destination so as to not just rely on the SLO endpoint (since the SLO endpoint usually is based on the callback URL, unless one specified a new URL which is a new option in this PR)
- Requested authentication context class refs are now checked again in SAML responses
- Test cases to verify critical areas.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
@@ -1,5 +1,8 @@
 package org.pac4j.saml.config;
 
+import net.shibboleth.utilities.java.support.net.URIComparator;
+import net.shibboleth.utilities.java.support.net.impl.BasicURLComparator;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.client.HttpClient;
 import org.opensaml.core.xml.schema.XSAny;
 import org.opensaml.saml.common.xml.SAMLConstants;
@@ -24,6 +27,7 @@ import org.pac4j.saml.metadata.SAML2IdentityProviderMetadataResolver;
 import org.pac4j.saml.metadata.SAML2MetadataContactPerson;
 import org.pac4j.saml.metadata.SAML2MetadataGenerator;
 import org.pac4j.saml.metadata.SAML2MetadataResolver;
+import org.pac4j.saml.metadata.SAML2MetadataSigner;
 import org.pac4j.saml.metadata.SAML2MetadataUIInfo;
 import org.pac4j.saml.metadata.SAML2ServiceProviderRequestedAttribute;
 import org.pac4j.saml.metadata.keystore.SAML2FileSystemKeystoreGenerator;
@@ -71,6 +75,10 @@ public class SAML2Configuration extends BaseClientConfiguration {
     private static final Logger LOGGER = LoggerFactory.getLogger(SAML2Configuration.class);
 
     private final List<SAML2ServiceProviderRequestedAttribute> requestedServiceProviderAttributes = new ArrayList<>();
+
+    private SAML2MetadataSigner metadataSigner;
+
+    private String singleSignOutServiceUrl;
 
     private String callbackUrl;
 
@@ -159,6 +167,8 @@ public class SAML2Configuration extends BaseClientConfiguration {
     private String attributeAsId;
 
     private Map<String, String> mappedAttributes = new LinkedHashMap<>();
+
+    private URIComparator uriComparator = new BasicURLComparator();
 
     private LogoutHandler logoutHandler;
 
@@ -582,6 +592,14 @@ public class SAML2Configuration extends BaseClientConfiguration {
         this.authnContextClassRefs = authnContextClassRefs;
     }
 
+    public URIComparator getUriComparator() {
+        return uriComparator;
+    }
+
+    public void setUriComparator(final URIComparator uriComparator) {
+        this.uriComparator = uriComparator;
+    }
+
     public String getNameIdPolicyFormat() {
         return nameIdPolicyFormat;
     }
@@ -716,6 +734,14 @@ public class SAML2Configuration extends BaseClientConfiguration {
         this.authnRequestExtensions = authnRequestExtensions;
     }
 
+    public SAML2MetadataSigner getMetadataSigner() {
+        return metadataSigner;
+    }
+
+    public void setMetadataSigner(final SAML2MetadataSigner metadataSigner) {
+        this.metadataSigner = metadataSigner;
+    }
+
     public String getAttributeAsId() {
         return attributeAsId;
     }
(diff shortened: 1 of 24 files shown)


pac4j-saml/src/main/java/org/pac4j/saml/sso/impl/SAML2AuthnResponseValidator.java
```diff
@@ -621,7 +617,23 @@ protected void validateAuthenticationStatements(final List<AuthnStatement> authn
...
-            // TODO implement authnContext validation
+            authnClassRefs.add(statement.getAuthnContext().getAuthnContextClassRef().getURI());
+        }
+
+        validateAuthnContextClassRefs(authnClassRefs);
+    }
+
+    protected void validateAuthnContextClassRefs(final List<String> providedAuthnContextClassRefs) {
+        if (!saml2Configuration.getAuthnContextClassRefs().isEmpty()) {
+            logger.debug("Required authentication context class refs are {}", saml2Configuration.getAuthnContextClassRefs());
```
aimiebell, line 654: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names should start with an upper case letter (NM_CLASS_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-should-start-with-an-upper-case-letter-nm-class-naming-convention).
Class names should be nouns, in mixed case with the first letter of each internal word capitalized. Try to keep your class names simple and descriptive. Use whole words-avoid acronyms and abbreviations (unless the abbreviation is much more widely used than the long form, such as URL or HTML).

mmoayyed: Thank you for the notes. In this case, I think we can live with the name.

aimiebell, line 654: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names should start with an upper case letter (NM_CLASS_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-should-start-with-an-upper-case-letter-nm-class-naming-convention).
Class names should be nouns, in mixed case with the first letter of each internal word capitalized. Try to keep your class names simple and descriptive. Use whole words-avoid acronyms and abbreviations (unless the abbreviation is much more widely used than the long form, such as URL or HTML).

mmoayyed:
I trust things look good here? Shall we merge?

leleuj:
I haven't had time to take a look. Go ahead. I'll review it later on.

mmoayyed:
Thanks much. I am vigorously reviewing and testing pac4j v5 with CAS. Just a warning: might also be posting a few other changes as I complete the testing! 😄

mmoayyed:
One thing that I did find immediately is, that I should update the releases notes. Will do in a second!

mmoayyed:
Updated release notes.

leleuj:
No problem. I planned a RC2 at the end of February. And the GA for March.

mmoayyed:
That is perfect! Thanks.

pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
```diff
@@ -860,6 +902,13 @@ public SAML2MetadataGenerator toMetadataGenerator() {
         }
     }
 
+    protected void determineSingleSignOutServiceUrl(final BaseSAML2MetadataGenerator generator) {
+        final String url = StringUtils.defaultIfBlank(this.singleSignOutServiceUrl, callbackUrl);
+        final String logoutUrl = CommonHelper.addParameter(url, Pac4jConstants.LOGOUT_ENDPOINT_PARAMETER, "true");
```
leleuj, line 907: What could be the use case here? Why would you receive SAML logout requests on another endpoint than the callback URL which knows how to handle that?

leleuj: The comment is outdated also.

pac4j-saml/src/main/java/org/pac4j/saml/metadata/SAML2MetadataSigner.java
```diff
@@ -0,0 +1,17 @@
...
+
+import org.opensaml.saml.saml2.metadata.EntityDescriptor;
+
+import java.io.File;
+
+/**
+ * This is {@link SAML2MetadataSigner}.
+ *
+ * @author Misagh Moayyed
+ * @since 6.4.0
```
leleuj, line 11: I guess you meant `5.0.0` ;-)

pac4j-saml/src/main/java/org/pac4j/saml/metadata/XMLSecSAML2MetadataSigner.java
```diff
@@ -0,0 +1,50 @@
...
+import org.pac4j.saml.config.SAML2Configuration;
+import org.pac4j.saml.exceptions.SAMLException;
+
+import java.io.File;
+
+/**
+ * This is {@link XMLSecSAML2MetadataSigner}.
+ *
+ * @author Misagh Moayyed
+ * @since 6.4.0
```
leleuj, line 14: Not the CAS version, but the pac4j one!

pac4j-saml/src/test/java/org/pac4j/saml/util/ExcludingParametersURIComparator.java
```diff
@@ -0,0 +1,26 @@
...
+import net.shibboleth.utilities.java.support.net.URIException;
+import org.apache.http.client.utils.URIBuilder;
+
+import javax.annotation.Nullable;
+
+/**
+ * This is {@link ExcludingParametersURIComparator}.
+ *
+ * @author Misagh Moayyed
+ * @since 6.4.0
```
leleuj, line 13: Mismatch with the CAS version.

leleuj (commented):
One question. And some comments about the version. I'll fix this later issue by myself.

pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
mmoayyed: First, thanks for the spring cleaning!

Often times, yes the endpoint that does the callback is responsible for handling both authn responses and logout responses. There are scenarios where you do not want them to be the same: idp posts an invalid logout response back, or one would want to handle the logout response rather differently via a dedicated url (that redirects to places, etc). So this allows one to separate the normal authn response handling and also potentially removes the need foe logoutendpoint parameter (which can become problematic as certain idps remove parameters when they post a logout response, which then causes destination values to not match)

leleuj: OK

PR #1810, opened 2021-02-09
Refactoring, annotations and JDK 11 migration

mmoayyed:
This is a pretty large pull request, but it's mostly cosmetic. It presents the following:

1. Adds missing annotations for `@Override` and `@FunctionalInterface`
2. Replaces lambdas with method references
3. Fixed log parameters and placeholders in areas where there is a mismatch
4. Adds `toString()` methods to HttpAction classes
5. Removes Travis CI config
6. Removes unused `TemporaryRedirectAction`
7. Uses type inference and `var`
8. Clean up `pac4j-saml` some more to allow more protected methods and break down large components.


leleuj:
Thanks for the big cleaning

PR #1811, opened 2021-02-09
allow samesite policy to be set on csrf cookie

hdeadman:
I was interested in being able to set samesite policy on the CSRF token (from CAS). This is adding support for that.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/context/JEEContext.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/context/JEEContext.java
@@ -6,6 +6,10 @@ import org.pac4j.core.util.CommonHelper;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.*;
 
 /**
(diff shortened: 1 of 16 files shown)


pac4j-core/src/main/java/org/pac4j/core/context/JEEContext.java
```diff
@@ -165,15 +165,7 @@ public String getFullRequestURL() {
 
     @Override
     public void addResponseCookie(final Cookie cookie) {
```
leleuj, line 167: I guess this could be moved as a default implementation of the `addResponseCookie` method in the `WebContext` interface.

leleuj (changes requested):
LGTM. One comment though.

hdeadman:
I just saw your comment and will adjust, but I made an update b/c I noticed I wasn't using the comment version or expires attributes of the cookie. I removed some of those because they don't seem to be used much and I based Expires on the max-age, modeled after this: 
https://github.com/spring-projects/spring-framework/blob/c9b27af64f41cc4fd06ee41a0faaca867e3cba4e/spring-web/src/main/java/org/springframework/http/ResponseCookie.java#L155

The method for setting the cookie is sort of a combination of CAS and Spring cookie handling.

hdeadman:
I moved it to JEEContext rather than WebContext, because it needs the HTTP response object to really do anything useful, but I broke it up so that the method to create the String could be moved to WebContext, but it would be a static method ? Is there a utility class it could move to?

hdeadman:
I copied some logic from CAS where it sets the Secure attribute on the cookie if the same site policy is "none". 

https://github.com/apereo/cas/blob/3f2311323d60e279ad2834a506cd7eeaf8a5daca/core/cas-server-core-cookie-api/src/main/java/org/apereo/cas/web/support/gen/CookieRetrievingCookieGenerator.java#L202

I don't know the rationale for that, but because the default same site policy in this PR is "none", the effect is that the cookie is set to "Secure" by default. That is probably not desirable and there is no way to turn off secure without changing the same site policy. Should the default be lax or should the logic copied from CAS be changed? 

I can make further updates or you are welcome to merge it and change it as you see fit. 

I added a couple tests. 

Thanks.

leleuj:
I think the `createCookieHeader` static method should be moved in the `ContextHelper` class (which should be renamed `WebContextHelper` to make things obvious).

leleuj:
About the default values, we can keep things like that. It is defined in the spec. See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

leleuj (approved):

hdeadman:
I see you merged, thanks. Regarding the default samesite policy of None, since most browsers seem to be defaulting to Lax if nothing is set, would Lax be a better default since that is the effective policy that users of pac4j are getting now?

leleuj:
For this major version 5, we can break things, but as you said, if most browsers default to Lax, we should certainly do the same.

PR #1812, opened 2021-02-11
Default samesite policy to Lax

hdeadman:
This changes the default SameSite policy to lax. Chrome sets policy to Lax by default if no policy is set, but in Firefox the default is configurable via a property and I believe the `network.cookie.sameSite.laxByDefault` is false by default. IE doesn't support SameSite at all. I don't feel strongly about what the default should be but I think this document says the "standard" was recently changed to use Lax if a policy isn't specified. The standard isn't really being followed by all browsers though. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#samesitenone_requires_secure

```
Standards related to the Cookie SameSite attribute recently changed such that:

The cookie-sending behavior if SameSite is not specified is SameSite=Lax. Previously the default was that cookies were sent for all requests.

Cookies with SameSite=None must now also specify the Secure attribute (they require a secure context/HTTPS).

```

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/context/WebContextHelper.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/context/WebContextHelper.java
@@ -135,6 +135,8 @@ public final class WebContextHelper implements HttpConstants {
 
     /**
      * Custom method for adding cookie because the servlet-api version doesn't support SameSite attributes.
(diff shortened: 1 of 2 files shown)


leleuj:
Excellent! Thanks for the follow up. I was about to do it myself...

PR #1826, opened 2021-02-22
Allow forced/passive authn to be defined per request

mmoayyed:
This pull request modifies the authentication request builders to support forced/passive authentication requests on demand. This is handled for SAML2, CAS and OIDC protocols, and is generally controlled via distinct http request attributes that get translated to the appropriate flags for each protocol.


leleuj:
Don't forget the release notes ;-)

mmoayyed:
Thanks for the reminder! Updated release notes and docs.

PR #1836, opened 2021-03-01
Support SAML2Configuration/OidcConfiguration overrides via http request attributes

mmoayyed:
This pull request extends the previous approach proposed by #1826 to allow overriding the SAML2 configuration via specific/dedicated http attributes. This is done by introducing a wrapping layer, `SAML2ConfigurationContext`, on top of the `SAML2Configuration` that looks for the http attribute and otherwise, falls back onto the default. The same technique is applied to overriding the `OidcConfiguration`.

Note that this pull request,

- Only provides a solution for SAML2 and OIDC. Other clients may be worked out in the future.
- Does not cover or suggest that every single setting in `SAML2Configuration` can be overridden per request. Only those fields/settings that are most practical or sensible to override on a per-request basis are included here, and others can be worked in the future on demand. Same goes for `OidcConfiguration`.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/matching/matcher/csrf/CsrfTokenGeneratorMatcher.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/matching/matcher/csrf/CsrfTokenGeneratorMatcher.java
@@ -1,6 +1,5 @@
 package org.pac4j.core.matching.matcher.csrf;
 
-import org.apache.shiro.util.StringUtils;
 import org.pac4j.core.context.Cookie;
 import org.pac4j.core.context.session.SessionStore;
 import org.pac4j.core.util.Pac4jConstants;
@@ -40,12 +39,12 @@ public class CsrfTokenGeneratorMatcher implements Matcher {
         final var token = csrfTokenGenerator.get(context, sessionStore);
         context.setRequestAttribute(Pac4jConstants.CSRF_TOKEN, token);
         final var cookie = new Cookie(Pac4jConstants.CSRF_TOKEN, token);
-        if (StringUtils.hasText(domain)) {
+        if (CommonHelper.isNotBlank(domain)) {
             cookie.setDomain(domain);
         } else {
             cookie.setDomain(context.getServerName());
         }
-        if (StringUtils.hasText(path)) {
+        if (CommonHelper.isNotBlank(path)) {
             cookie.setPath(path);
         }
         if (httpOnly != null) {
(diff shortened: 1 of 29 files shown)


leleuj:
I was about to release the 5.0.0-RC2. Will wait for this one to be merged...

pac4j-core/src/main/java/org/pac4j/core/matching/matcher/csrf/CsrfTokenGeneratorMatcher.java
```diff
@@ -39,12 +40,12 @@ public boolean matches(final WebContext context, final SessionStore sessionStore
         final var token = csrfTokenGenerator.get(context, sessionStore);
         context.setRequestAttribute(Pac4jConstants.CSRF_TOKEN, token);
         final var cookie = new Cookie(Pac4jConstants.CSRF_TOKEN, token);
-        if (domain != null) {
+        if (StringUtils.hasText(domain)) {
```
leleuj, line 43: Shiro is an optional dependency, we can't use its `StringUtils` method. Use the pac4j `CommonHelper`.

mmoayyed:
That would be great, thanks! I am applying the same technique (though to a lesser extent) to the OIDC config. Should be ready in an hour or so.

leleuj (changes requested):
I like the approach. The  `CommonHelper` class must be used instead of the `StringUtils` class from Shiro.

leleuj:
For once, I was almost too quick in my code review ;-)

mmoayyed:
This should now be ready for further review and/or merge.

leleuj (approved):

leleuj:
It looks good, except I have one question: where is the `commons-io` dependency used in the OIDC support?

mmoayyed:
Good question; I think it can be removed and I just did. I was at some point trying to take advantage of BooleanUtils but now it's not needed of course in OIDC.

pom.xml
```diff
@@ -297,6 +297,13 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <version>3.1.2</version>
+				<dependencies>
```
leleuj, line 300: I'm surprised by this addition: what is it for?

mmoayyed: From the CI log:

```
[INFO] --- maven-checkstyle-plugin:3.1.2:check (checkstyle) @ pac4j ---
Warning:  Old version of checkstyle detected. Consider updating to >= v8.30
Warning:  For more information see: maven.apache.org/plugins/maven-checkstyle-plugin/examples/upgrading-checkstyle.html
[INFO] You have 0 Checkstyle violations.
[INFO] 
```

mmoayyed:
PS Once RC2 is released, I'll update CAS and then will begin to build on to of this PR to allow for overrides per service.

leleuj:
I started the release...

leleuj:
It's done. The artifacts will be available in the Maven central repository very soon.

mmoayyed:
Thank you, kind sir. This calls for a celebratory tweet!

leleuj:
;-)

PR #1850, opened 2021-03-18
Fix certificate expiration calculation

julianladisch:
Properly handle leap years (366 days).

Properly handle months with less than 31 days (30, 29, 28 days).

Use a single start date - don't fetch two start dates using System.currentTimeMillis() and DateTime.now(), they may differ.


leleuj:
Thanks for the contribution.

PR #1871, opened 2021-04-07
Add an option to tolerate missing SAML response `Destination` attribute

jmesny:
The Response `Destination` attribute is optional according to the SAML protocol specification.
(Paragraph 3.2.2, line 1554: https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf)
For security reasons it is expected to be provided but this behavior can be turned off through the `SAML2Configuration`.

### DIFF
--- a/documentation/docs/clients/saml.md
+++ b/documentation/docs/clients/saml.md
@@ -133,6 +133,13 @@ By SAML specification, the authentication request must not contain a NameQualifi
 cfg.setUseNameQualifier(true);
 ```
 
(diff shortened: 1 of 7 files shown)


mmoayyed:
LGTM. Let's wait for @leleuj

leleuj:
Good. Can you update the doc as well here: https://github.com/pac4j/pac4j/blob/master/documentation/docs/release-notes.md and there: https://github.com/pac4j/pac4j/blob/master/documentation/docs/clients/saml.md? Thanks

leleuj:
Thanks for your contribution

PR #1872, opened 2021-04-08
Fix `OidcConfigurationContext` that should read scope from `OidcConfiguration`

jmesny:
Since version 5.0.0 (PR #1836), the scope value defined on `OidcConfiguration` is not considered anymore when computing the redirect URL.


leleuj:
Good catch!

PR #1879, opened 2021-04-15
Parse JWT claims as `JSONObject` into individual attributes

mmoayyed:
Given the following JWT:

```json
{
  "iss": "https://example",
  "aud": "https://test",
  "jti": "TEST",
  "iat": 1618481574,
  "exp": 1618482174,
  "sub": "user",
  "userData": {
    "id": 1436000,
    "gender": "M"
  }
}
```

Parse the JWT and build the final user profile such that the `userData` claim is expanded into individual profile attributes for `id` and `gender`. Keep the `userData` attribute as current behavior and do not remove it from the final profile by default.

After the authentication, the final profile would be:

```
{
  "iss": "https://example",
  "aud": "https://test",
  "jti": "TEST",
  "iat": 1618481574,
  "exp": 1618482174,
  "sub": "user",
  "id" : "1436000",
  "gender" : "M",
  "userData": {              // this can be removed using setKeepNestedAttributes(false);
    "id": 1436000,
    "gender": "M"
  }
}

### DIFF
--- a/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
+++ b/pac4j-jwt/src/main/java/org/pac4j/jwt/credentials/authenticator/JwtAuthenticator.java
@@ -1,5 +1,34 @@
 package org.pac4j.jwt.credentials.authenticator;
 
+import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jwt.EncryptedJWT;
+import com.nimbusds.jwt.JWT;
+import com.nimbusds.jwt.JWTParser;
+import com.nimbusds.jwt.PlainJWT;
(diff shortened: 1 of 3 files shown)


leleuj:
Just to be sure: this is not a breaking change, currently, the `id` and `gender` attributes would be ignored, right?

I planned to release the v5.0.1 today, but I guess this one can go with it and I can do the release after that. Is it good for you?

mmoayyed:
Yes, this is not a breaking change. The extra attributes would just be ignored, or be processed as normal attribute/claim just like any other.

It would be great if we could include this in 5.0.1, but not a showstopper if it's too much work. 

Thanks!

leleuj:
It's not too much work. Will merge and start the release.

PR #1882, opened 2021-04-19
SAML2: improve on idp metadata resolution options

mmoayyed:
This patch:

1. Allows the API to force the resolution and download of SAML2 idp metadata. (The only to do that is now to deconstruct the SAML2 client and reinitialize it)
2. Allow "last-modified" attributes for both file-system and URL-based metadata resources.
3. Support proxies for idp metadata URL resources

There is also some internal cleanup where the `SAML2IdentityProviderMetadataResolver` is broken down into smaller pieces/methods for easier review and extensions.

This patch should remain compatible with previous versions, both for behavior and API.

If possible, it would be great to backport this into 4.x and I can certainly either cherry-pick or post a PR if a 4.x patch can be released at some point.


leleuj:
LGTM. Would you mind dropping a line in the release notes? Thanks

leleuj:
Of course, you can cherry pick/open a new PR for this change on 4.x and I can cut a 4.4.0 release.

mmoayyed:
Thanks much. Updated release notes. I'll post another PR for 4.x when this is merged.

PR #1885, opened 2021-04-21
Backporting patches: SAML2 (forced) metadata resolution, nested JWTs

mmoayyed:


leleuj:
Maybe it should have been in a v4.5.0, but anyway, let's merge that in a v4.4.1. Do you need the v4.4.1 release now?

mmoayyed:
Thanks much. I don't need it just yet, (will do some internal testing with the SNAPSHOT) but it would be great to have it out before Apr 30th, so we could include it in the next CAS 6.3.x release.

leleuj:
OK. I'll do the release next week. Ping me on Wednesday/Thursday if you don't see the release coming.

mmoayyed:
Sure thing, thanks again!

leleuj:
I finally created a version 4.5.0. It will be available soon in the Maven central repository.

mmoayyed:
That's great. Thanks!

PR #1936, opened 2021-07-01
Expose HttpServletRequest's request url for web context

mmoayyed:
Avoids casting down to JEEContext.

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -6,6 +6,7 @@ title: Release notes&#58;
 **v5.1.2**:
 
 - Support SAML2 `Scoping` in authentication requests.
+- `WebContext` is now able to provide the request url directly.
 
 **v5.1.1**:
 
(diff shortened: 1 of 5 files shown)


leleuj:
OK. I understand.
There is a problem though: this is a breaking change and this cannot be done for a minor version.

Hopefully, we can workaround that problem: can you propose a default implementation of `getRequestURL` in the `WebContext` interface which just removes the query string from the `getFullRequestURL` method if it exist (like you do in `MockWebContext`)?
Thanks

mmoayyed:
Sure. Will do.

mmoayyed:
This is done.

PS I will also be posting a new PR shortly to support SAML2 Scoping in authentication requests.

leleuj:
Two last points:
- with the default implementation of `getRequestURL` in the `WebContext`, do you still need an implementation in `JEEContext`?
- can you drop a line in the release notes?

mmoayyed:
I updated the release notes.

I think it's safest for the `JEEContext` to override the default implementation, since it delegates to the *real* request, and real implementations might want to handle this scenario differently. I wouldn't want to interfere with that.

leleuj:
OK. Thanks

PR #1937, opened 2021-07-01
Support SAML2 scoping in authn requests

mmoayyed:
Scoping allows a pac4j SAML2 service provider to specify a list of identity providers in an authnRequest to an identity provider. This is an indication to the identity provider that the service will only deal with the identity providers specified.

Authn requests may look like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<saml2p:AuthnRequest xmlns:saml2p="urn:oasis:names:tc:SAML:2.0:protocol" 
  AssertionConsumerServiceIndex="1" ForceAuthn="false" 
  ID="_ebf5682cbf7c48cf908c4d9127057b09920a3b2" IsPassive="false" IssueInstant="2021-07-01T11:31:27.426Z" 
   ProviderName="pac4j-saml" Version="2.0">
<saml2:Issuer xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion" 
   Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity">entity-id</saml2:Issuer>
<saml2p:Scoping>
<saml2p:IDPList>
   <saml2p:IDPEntry Name="My Identity Provider" ProviderID="idp-entity-id"/></saml2p:IDPList>
</saml2p:Scoping>
</saml2p:AuthnRequest>
```

PS Support for this change will be added to CAS once pac4j is released.

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -3,6 +3,10 @@ layout: doc
 title: Release notes&#58;
 ---
 
+**v5.1.2**:
+
+- Support SAML2 `Scoping` in authentication requests.
+
 **v5.1.1**:
 
 - Removed the ORCID OAuth client which no longer works. Use the `OidcClient` instead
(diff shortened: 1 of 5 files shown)


leleuj:
I will be very soon on holidays, bu I'll cut a v5.1.2 release before that...

PR #1926, opened 2021-06-17
Extraction of attributevalues with complex types

fbusselgln:
Hello everyone,

while using Pac4j for SAML Delegate Authentication in CAS, I have encountered a problem when extracting complex type attributes from SAML2 assertions. After checking with the Pac4j codebase, I have noticed that there is no differentiation at all and the value is just being extracted by calling dom.getTextContent(), which stuffs all attribute values from complex types into one long string.

Example:

```
    <saml:AttributeStatement>
      <saml:Attribute Name="Foo">
        <saml:AttributeValue xsi:type="xs:string">Bar</saml:AttributeValue>
      </saml:Attribute>
      <saml:Attribute Name="Organization"
        NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:basic">
        <saml:AttributeValue>
          <OrganizationType
            xmlns:xsd="http://www.w3.org/2001/XMLSchema"
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xmlns:q1="urn:osp:foo:schemas:identity"
            xsi:type="q1:OrganizationType"
            id="ab417e71-3fe8-4981-d629-08d90ef68525">
            <q1:OrganizationName q1:attributeId="urn:osp:foo:attribute:organizationName">Example Corp</q1:OrganizationName>
            <q1:EmailAddress q1:attributeId="urn:osp:foo:attribute:emailAddress">employee@example.com</q1:EmailAddress>
            <q1:Address q1:attributeId="urn:osp:foo:attribute:address">
              <q1:Street q1:attributeId="urn:osp:foo:attribute:street">Example St</q1:Street>
              <q1:StreetNumber q1:attributeId="urn:osp:foo:attribute:streetNumber">123</q1:StreetNumber>
              <q1:City q1:attributeId="urn:osp:foo:attribute:city">ExampleCity</q1:City>
              <q1:ZipCode q1:attributeId="urn:osp:foo:attribute:zipCode">123456</q1:ZipCode>
              <q1:Country q1:attributeId="urn:osp:foo:attribute:country">ExampleCountry</q1:Country>
            </q1:Address>
          </OrganizationType>
        </saml:AttributeValue>
      </saml:Attribute>
    </saml:AttributeStatement>
```

While the attribute Foo is extracted without a problem, the attribute Organization with it's complex value type is extracted as one large text with line delimiters. After extraction with the current code base, printing the results looks like this:

```
Foo=Bar
Organization=
          
            Example Corp
            employee@example.com
[...]

PR #1938, opened 2021-07-02
Fix SAML2 response attributes resolution when using ADFS as IdP

jmesny:
Pac4j 5.1.1 (PR #1926) introduced a change on how attributes are extracted from SAML responses.
This causes a regression when dealing with ADFS IdPs that omit the `xsi:type` attribute on the `AttributeValue` elements.
The SAML specification indicates this attribute is optional ([2.7.3.1.1, line 1239](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf)), so Pac4j can't rely on it to resolve response attributes.

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -7,6 +7,7 @@ title: Release notes&#58;
 
 - Support SAML2 `Scoping` in authentication requests.
 - `WebContext` is now able to provide the request url directly.
+- Fix SAML2 response attributes resolution when using ADFS as IdP.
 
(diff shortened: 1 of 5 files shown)


leleuj:
Excellent! Thanks
I plan to cut the v5.1.2 release next Monday.

fbusselgln:
Sorry for the trouble and thanks for the excellent fix!

leleuj:
The version 5.1.2 has been released.

jmesny:
Thanks! 🙏

PR #1939, opened 2021-07-02
Allow claims mapping for OIDC user profile

mmoayyed:
When fetching the OIDC user profile, allow the OIDC configuration to specify arbitrary claim mapping rules.

PS This is similar to how SAML2 allows mapping of attributes.


leleuj:
I plan to cut the v5.1.2 release next Monday.

PR #1941, opened 2021-07-05
Allow PartialLogout to be a successful as SAML2 logout response, if configured explicitly. 

fbusselgln:
When sending a SAML2 logout response to a service, the IdP can respond with urn:oasis:names:tc:SAML:2.0:status:PartialLogout because it was unable to contact all participating services (SLO). The session has been terminated on the IdP however and since the logout request was initiated by the service, it may be ok to continue the flow even when SLO is not configured properly somewhere else.

This changeset introduces a new SAML2 client property, allowing to treat PartialLogout responses like a Success, if configured explicitly. The default behaviour is not changed for now, although it would probably make sense, because the logout process between Pac4j-Service and IdP is still successful in the described case.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
@@ -170,6 +170,7 @@ public class SAML2Client extends IndirectClient {
             this.configuration.getPostLogoutURL(), this.replayCache,
             this.configuration.getUriComparator());
         this.logoutValidator.setAcceptedSkew(this.configuration.getAcceptedSkew());
+        this.logoutValidator.setIsPartialLogoutTreatedAsSuccess(this.configuration.isPartialLogoutTreatedAsSuccess());
     }
 
     protected void initSAMLResponseValidator() {
--- a/pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/config/SAML2Configuration.java
@@ -109,6 +109,8 @@ public class SAML2Configuration extends BaseClientConfiguration {
 
     private String comparisonType = null;
 
+    private boolean isPartialLogoutTreatedAsSuccess = false;
+
     private String authnRequestBindingType = SAMLConstants.SAML2_POST_BINDING_URI;
 
     private String responseBindingType = SAMLConstants.SAML2_POST_BINDING_URI;
@@ -946,5 +948,13 @@ public class SAML2Configuration extends BaseClientConfiguration {
 
     public void setIdentityProviderMetadataResolver(final SAML2MetadataResolver identityProviderMetadataResolver) {
         this.identityProviderMetadataResolver = identityProviderMetadataResolver;
+    }
+
+    public boolean isPartialLogoutTreatedAsSuccess() {
+        return isPartialLogoutTreatedAsSuccess;
+    }
+
+    public void setPartialLogoutTreatedAsSuccess(boolean partialLogoutTreatedAsSuccess) {
+        isPartialLogoutTreatedAsSuccess = partialLogoutTreatedAsSuccess;
     }
 }
--- a/pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutValidator.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutValidator.java
@@ -6,6 +6,8 @@ import org.opensaml.saml.common.SAMLObject;
 import org.opensaml.saml.common.xml.SAMLConstants;
 import org.opensaml.saml.saml2.core.LogoutRequest;
 import org.opensaml.saml.saml2.core.LogoutResponse;
+import org.opensaml.saml.saml2.core.Status;
+import org.opensaml.saml.saml2.core.StatusCode;
 import org.opensaml.saml.saml2.encryption.Decrypter;
 import org.opensaml.saml.saml2.metadata.Endpoint;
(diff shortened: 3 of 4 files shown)


fbusselgln:
The current behaviour is: If status != success then throw SamlException with no configuration option whatsoever.

leleuj:
Will merge "as is" and make the treated-as-success property to `true`.

fbusselgln:
Thank you!

fbusselgln:
Do you think you will build a new version any time soon? Has been a while since 5.1.2... ;)

leleuj:
I just cut the v5.1.3 release. It will show up in the Maven central repository within a few minutes.

fbusselgln:
Thanks. However it looks like the method `SAML2Client.initSAMLLogoutResponseValidator()` which transports the new property's value only has one usage in `SAML2Client.internalInit()` which is not called when a SAML2Client is instantiated, at least in CAS. Is this a problem in Pac4j or rather a missing `SAML2Client.init()` call in CAS?

I already suspected this could be a problem, as I am getting a NullPointerException on the logoutValidator in the [DelegatedAuthenticationClientFinishLogoutAction:75](https://github.com/apereo/cas/blob/6e29bc0001e3c304375efc5f8cbb04918d8f8691/support/cas-server-support-pac4j-webflow/src/main/java/org/apereo/cas/web/flow/DelegatedAuthenticationClientFinishLogoutAction.java#L75) unless you set `lazy-init=false`, but that does not solve the missing transfer of the new property introduced with this pull request here.

fbusselgln:
tl;dr the original problem is not resolved. Pac4j issue or CAS issue?

leleuj:
I'm on holiday. I'll check that in September...

fbusselgln:
Ok, thanks for the info. Have a nice holiday! :)

leleuj:
The `SAML2Client.init()` method should be automatically called when using the `SAML2Client`. This should happen when performing the SAML login process, but maybe not in the `DelegatedAuthenticationClientFinishLogoutAction`. So the test may fail if you only send a SAML logout request without the login part.

fbusselgln:
Apparently that method is not called. The logout procedure went through the same CAS instance (and hence the same SAML2Client instance) which was used for logging in shortly before.

How do you suggest to proceed?

leleuj:
This `internalInit` method should be called during the login process. Can you put a breakpoint in it and perform a full login/logout process? Thx

fbusselgln:
Unfortunately not without hours of preparations. Don't the arguments convince you? :)

- Opposite to other pac4j subprojects, that method `SAML2Client.internalInit()` is never called in pac4j-saml (see screenshot)
- The only usage of `SAML2Client.initSAMLLogoutResponseValidator()` is in `SAML2Client.internalInit()` and it is obviously not called, because the property values are not being transported and the logoutValidator is not even instantiated.

internalInit-usages

fbusselgln:
Apologies, it seems like I need to reinvestigate what is happening here. The init procedures are called correctly apparently:

```
2021-09-07 07:42:16,527 DEBUG [org.apereo.cas.support.saml.OpenSamlConfigBean] - <Initialized OpenSaml successfully.>
2021-09-07 07:42:19,709 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: CasClient>
2021-09-07 07:42:19,728 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: CasConfiguration>
2021-09-07 07:42:19,757 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: DirectBasicAuthClient>
2021-09-07 07:42:19,760 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: DirectFormClient>
2021-09-07 07:42:19,762 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: DirectFormClient>
2021-09-07 07:42:19,762 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: DirectBasicAuthClient>
2021-09-07 07:42:19,764 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: DirectFormClient>
2021-09-07 07:42:19,765 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: DirectFormClient>
2021-09-07 07:42:19,781 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: HeaderClient>
2021-09-07 07:42:19,890 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: HeaderClient>
2021-09-07 07:42:19,894 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: DirectFormClient>
2021-09-07 07:42:19,899 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: DirectFormClient>
2021-09-07 07:42:23,444 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: SAML2Client>
2021-09-07 07:42:23,449 DEBUG [org.pac4j.core.util.InitializableObject] - <Initializing: SAML2Configuration>
```

fbusselgln:
Sorry, I should have started another debugging session right away instead of arguing based on thoughts that have been over a month old. Anyway, now I know what fails.

The new check in `SAML2LogoutValidator.validateSuccess` checks for equality of the status value and the constant `public static final String PARTIAL_LOGOUT = "urn:oasis:names:tc:SAML:2.0:status:PartialLogout";`. However the status response in my scenario is `urn:oasis:names:tc:SAML:2.0:status:Responder / urn:oasis:names:tc:SAML:2.0:status:PartialLogout`, which obviously is not a String equality.

Question: Should the Logout Validator be satisfied, if the response **contains** `urn:oasis:names:tc:SAML:2.0:status:PartialLogout`, regardless of what else is in the status?

leleuj:
OK. Let's do that.

PR #1998, opened 2021-09-07
Check if the PartialLogout response code is part of the response instead of checking for String equality

fbusselgln:
(Followup of https://github.com/pac4j/pac4j/pull/1941 )

Partial Logouts are accepted by the `SAML2LogoutValidator` since 5.1.3, however the actual validation currently relies on string equality between the constant `"urn:oasis:names:tc:SAML:2.0:status:PartialLogout"` and the response string. Some identity providers return more than the partial logout status, i.e. when an internal error occured. Example response: `urn:oasis:names:tc:SAML:2.0:status:Responder / urn:oasis:names:tc:SAML:2.0:status:PartialLogout`

This behaviour currently leads to an exception in the Pac4j flow. This PR introduces a solution.


fbusselgln:
Thanks! ;)

fbusselgln:
How about building version 5.1.4 ? ;)

mmoayyed:
Discussions about the release and roadmap should be on the pac4j mailing list. Please do not post phantom comments.

PR #2026, opened 2021-10-07
Allow security logic to forcefully execute clients

mmoayyed:
- Adds an optional boolean flag to the default security logic to not load profiles from the session and instead loop through the existing clients to renew the user profile. The default behavior is as it was before, without change.
- Add test cases to verify behavior.


mmoayyed:
Thanks for the notes. I think a name like `loadProfilesFromSession` should be a good enough solution for now. Not sure we need to bring back a full decision type of object yet. If the security logic gets more complicated in the future where there is a possibility for 1-2 other boolean flags, then it would make sense to bring back the decision object and clean things up. 

You're quite right about session issues. The real motivation for this change actually does stem from CAS where we have authorization and access-token endpoints protected by security interceptors and security logic components. Both interceptors look into the session and if they find a profile, they skip the logic execution. This causes a scenario where:

1. Your application authenticates with CAS with a username/password, puts that profile into the session, and then receives an oauth code
2. The application then submits a GET request (in the same session) to exchange that code for an access token. The interceptor for the access token finds the session in one, and ignores the credentials pass in the request to authenticate the request by client-id/client-secret. 
3. As a result, the request is rejected, because the profile in #1 is not a valid profile for #2 

This is a broader concern today where, 

- If you have two interceptors in the same session, a profile established in a session might not be immediately useable for the 2nd interceptor in the same session, and the 2nd interceptor must somehow clear the profile session before it can do its job.

I think the `loadProfilesFromSession=false` flag should allow the interceptor for the access token to always forcefully invoke the client regardless of the session credential.

mmoayyed:
Yes you're quite right. This, what I described, is only the symptom of the issue and not a problem in and of itself.

 Generally speaking, two interceptors on different endpoint patterns sharing the same session store run into a conflict if one needs to "force fetch" the credential again when a profile exists in the session already. That is the main use case here. The "invalid" oauth call is just a story that while incorrect, led to this discovery.

leleuj:
No problem. This happens.
But I think the code is wrong too:

```
var profiles = this.loadProfilesFromSession
                    ? List.<UserProfile>of()
                    : loadProfiles(manager, context, sessionStore, currentClients);
```

It should be:

```
var profiles = this.loadProfilesFromSession
                    ? loadProfiles(manager, context, sessionStore, currentClients)
                    : List.<UserProfile>of();
```

no?

leleuj:
The good point is that the double problem makes things work right ;-)
I will fix the code for the next release.

leleuj:
@mmoayyed I just released pac4j v5.1.5. Be careful when upgrading in CAS as the `loadProfilesFromSession` is now correct and is inverted compared to v5.1.4.

mmoayyed:
Sure. Thank you!

mmoayyed:
No issues with the upgrade. The original use case that prompted this change is also now working correctly, thanks to many tests verifying this!

Thanks for handling the release. onward and upward :)

PR #2029, opened 2021-10-13
Handle exceptions when SAML2 ogout action cannot be built

mmoayyed:
Adds additional error handling into the SAML2 logout action to avoid exceptions like this:

```
Caused by: org.pac4j.saml.exceptions.SAMLException: Identity provider has no single logout service available for the selected profile urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST
	at org.pac4j.saml.context.SAML2MessageContext.getIDPSingleLogoutService(SAML2MessageContext.java:135)
	at org.pac4j.saml.logout.impl.SAML2LogoutRequestBuilder.build(SAML2LogoutRequestBuilder.java:45)
	at org.pac4j.saml.logout.SAML2LogoutActionBuilder.getLogoutAction(SAML2LogoutActionBuilder.java:58)
	at org.pac4j.core.client.IndirectClient.getLogoutAction(IndirectClient.java:160)
```

Presently, this causes a crash which is against the method contract (that should return an `empty()` instead).

Changes are better reviewed using this link: https://github.com/pac4j/pac4j/pull/2029/files?w=1


leleuj:
Yes, I remember seeing this issue in the CAS server.

PR #2128, opened 2022-01-20
PartialLogout handling / SAML2 spec compliance

fbusselgln:
Hello everyone,

a while ago I submitted a pull request to support the handling of PartialLogout status codes in SAML2 Responses as successful logouts. However I made a mistake in the implementation, which doesn't comply with the SAML2 spec. Apologies for that. Our SAML2 IdP had that part incorrectly in their implementation too and only fixed it a while after we had been discussing the issue. Now their fix is live and I stumbled upon this topic again.

**Actual behaviour**: The current implementation searches for the PartialLogout code only on the top level.

**Expected behaviour**: The PartialLogout code is allowed as a sub-code of the Responder status-code. 

**Proof**: From the SAML 2.0 core spec ([link](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf)) lines 2641-2645:
> In the event that not all session participants successfully respond to these <LogoutRequest> messages
> (or if not all participants can be contacted), then the session authority MUST include in its
> <LogoutResponse> message a second-level status code of
> urn:oasis:names:tc:SAML:2.0:status:PartialLogout to indicate that not all other session
> participants successfully responded with confirmation of the logout.

Schema definition (same document, line 1700):
```
<element name="StatusCode" type="samlp:StatusCodeType"/>
<complexType name="StatusCodeType">
  <sequence>
    <element ref="samlp:StatusCode" minOccurs="0"/>
  </sequence>
  <attribute name="Value" type="anyURI" use="required"/>
</complexType>
```

However according to lines 1634 and following, PartialLogout is also allowed on top level. If it occurs on second level, then only the Responder status code seems to be eligible to carry the PartialLogout status as second level code.

**Impact of this pull request**: Now both top level and second level status codes of type PartialLogout are supported, as described by the OASIS spec.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutValidator.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/logout/impl/SAML2LogoutValidator.java
@@ -197,13 +197,19 @@ public class SAML2LogoutValidator extends AbstractSAML2ResponseValidator {
     @Override
     protected void validateSuccess(Status status) {
 
-        if (isPartialLogoutTreatedAsSuccess
-            && status != null && status.getStatusCode() != null && status.getStatusCode().getValue() != null
-            && status.getStatusCode().getValue().contains(StatusCode.PARTIAL_LOGOUT)) {
-            logger.debug(
-                "Response status code is {} and partial logouts are configured to be treated as success => validation successful!",
-                StatusCode.PARTIAL_LOGOUT);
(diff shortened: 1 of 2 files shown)


fbusselgln:
It seems like the Mergify-configuration is something i cannot fix?!

leleuj:
Yes, forget about it.

leleuj:
Thanks

PR #2174, opened 2022-03-09
add support for Windows ADFS Openid Connect using self-signed cert

yswang1024:
add support for Windows ADFS Openid Connect using self-signed certificate in development or testing environment by using custom SSLFactory.
use it like:
config.setSSLFactory("org.pac4j.oidc.test.XXXSSLSupportFactory");
it's working fine and passed the test.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
@@ -1,6 +1,7 @@
 package org.pac4j.oidc.config;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


yswang1024:
Hi,
A while ago, due to the company's system requirements, I began to integrate Windows ADFS Openid Connect into the project, but I encountered a problem: because it is a development / test environment, ADFS uses self signed certificates and needs to use custom SSLFactory, but pac4j-oidc does not provide relevant setters, so I modified the source code of OidcConfiguration to support custom SSLFactory, And it works fine in the development/test environment. passed the test, can I create a pull request for the file.

This request is to support enterprise‘s special case on ADFS.

yswang1024:
ALL CHECKED

PR #2179, opened 2022-03-17
Restoring lost methods.

aeismann:
During our migration to 5.3.2 we noticed that the remove*Attribute methods have been dropped from the UserProfile interface as it seems. They still exist on BasicUserProfile and there are no signs of deprecation, so it might be good to add them again.

Sorry, I couldn't follow the instructions in the contribution guide, because I don't and will not own a google-group-account.


leleuj:
I don't remember why I removed these methods from the interface. I can't recall the rationale behind that. We can add them back.

PR #2188, opened 2022-03-22
Implement RFC 9207 OAuth 2.0 Authorization Server Issuer Identification

papegaaij:
IdPs supporting this specification, will advertise this via the metadata. If the IdP indicates it supports RFC 9207, the RP must check the issuer parameter in authorization responses. Its value must match the issuer in the IdP metadata.

See https://www.rfc-editor.org/rfc/rfc9207.html for more information. It is a simple spec that protects RPs from mix-up attacks: a form of attack were a client is tricked into sending an authorization code for one IdP to another, compromised IdP.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/extractor/OidcExtractor.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/extractor/OidcExtractor.java
@@ -99,6 +99,12 @@ public class OidcExtractor implements CredentialsExtractor {
             logger.debug("Authentication response successful");
             var successResponse = (AuthenticationSuccessResponse) response;
 
+            var metadata = configuration.getProviderMetadata();
+            if (metadata.supportsAuthorizationResponseIssuerParam() &&
+                !metadata.getIssuer().equals(successResponse.getIssuer())) {
+                throw new TechnicalException("Issuer mismatch, possible mix-up attack.");
+            }
+
             if (configuration.isWithState()) {
                 // Validate state for CSRF mitigation
                 final var requestState = (State) configuration.getValueRetriever()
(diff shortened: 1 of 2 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/extractor/OidcExtractor.java
```diff
@@ -99,6 +99,11 @@ public OidcExtractor(final OidcConfiguration configuration, final OidcClient cli
             logger.debug("Authentication response successful");
             var successResponse = (AuthenticationSuccessResponse) response;
 
+            if (configuration.getProviderMetadata().supportsAuthorizationResponseIssuerParam() &&
```
leleuj, line 102: Should we check nullity on the `successResponse` or `getIssuer()`?

Can you create a variable for the `configuration.getProviderMetadata()` part outside the `if`? Thanks

leleuj (changes requested):
The idea is good. A few comments.

Can you sync with master too? Thanks

pac4j-oidc/src/main/java/org/pac4j/oidc/credentials/extractor/OidcExtractor.java
papegaaij: The response cannot be null. If the response cannot be parsed, the parse method will throw an exception. The issuer is a required parameter for the metadata, the code will refuse to instantiate a `AuthorizationServerMetadata` with a `null` issuer. When the server advertises support for this parameter, it is required to set it. Failing to so do must be treated as a mismatch, that's why the equals is written like this: metadata issuer is guaranteed to be not null, response issuer must be not null to match.

I'll update the code to extract a local var for the metadata and rebase on master.

papegaaij:
@leleuj I've rebased on master, extracted a var for the metadata and squashed the commits.

leleuj:
Thanks. Good. Merging

PR #2301, opened 2022-08-13
allow to remove or replace a specific default matcher or autorizer by name

SenCodeMaker:
My problem: I had a angular app authenticating through saml session. pac4j were sending csrftoken httpONly, so i had to  use withCredentials when making http request with angular httpClientModule. The problems began when a request comes after two subsequent requests. 

in that case the token sent by the browser is too old even if it's in the same second and then it is not seen by the crsfAutorizer as a current nor a previousToken. Moreoever before that i had to create an autorizer that would read into the cookies as the browser sent them because the default one looks only into header and request param.

Now this solve my issue but still i have to put my new autorizer into the config and now i see that i can only had new autorizers by beginning the matcher string with a "+" or i can replace them all. because the default one prevented my angular app to authentcate I had to replace it a thus I had to copy all the other default matchers so they're still existing in my new config.

This pull request tries to answer this problem by allow users to put at the end of the matcher (or autorizer) string a "-" specifying which of the default he wants to be removed from his config, the others would stay.

As an addendum I would think interesting to completely remove the new for this string for matcher or authorizer one could just give a map of the added matchers or a map of mathers that ovverides the defaults. For the case of removal the user would give a list of names in an enumeration that describe the default matchers

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/authorization/checker/DefaultAuthorizationChecker.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/authorization/checker/DefaultAuthorizationChecker.java
@@ -1,19 +1,34 @@
 package org.pac4j.core.authorization.checker;
 
-import org.pac4j.core.authorization.authorizer.*;
+import static org.pac4j.core.util.CommonHelper.areEqualsIgnoreCaseAndTrim;
+import static org.pac4j.core.util.CommonHelper.assertNotNull;
+import static org.pac4j.core.util.CommonHelper.assertTrue;
+import static org.pac4j.core.util.CommonHelper.isBlank;
+import static org.pac4j.core.util.CommonHelper.isNotEmpty;
+import static org.pac4j.core.util.CommonHelper.substringAfter;
+import static org.pac4j.core.util.CommonHelper.substringBetween;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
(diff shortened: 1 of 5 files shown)


leleuj:
I'm not convinced by this change, I'm even doubtful about the "+" option I added. Let's always try to be as easy as possible.

Can you give me an example? The configuration you needed and the new configuration thanks to this PR so I can assess the benefit.

SenCodeMaker:
> I'm not convinced by this change, I'm even doubtful about the "+" option I added. Let's always try to be as easy as possible.
> 
> Can you give me an example? The configuration you needed and the new configuration thanks to this PR so I can assess the benefit.

i have to reapt myself here, but the problem i stated is the default crsrAuthorizer was not working for me because it read the crsf token only from header and request param. My angular app through web browser with "withCredential" option only sent this crsf token **in the cookies** . So i needed to replace one default authorizer (the crsf authorizer) and hold the other defaults. this is not possible with the current code. I had to override all the defaults only to had my special crsfAuthorizer.

my pull request adresses this to allow someone to only override a particular autorizer or matcher and not all the default config

leleuj:
You are partially right. It is possible to override a default authorizer, but not a default matcher.

This is fixed: https://github.com/pac4j/pac4j/commit/9071747b66b4c62f652363305c78cfa0738eb92a

Can you test again with version 5.4.7-SNAPSHOT?
You can either override a specific matcher or the global "securityheaders" one. This should meet most (if not all) needs.

PR #2313, opened 2022-08-23
Remove xalan dependency

hmarzooq:
This is a pull request to remove Apache xalan dependency as the OWASP scan of `pac4j-saml:3.9.0` has marked Apache Xalan as having critical vulnerability of CVSSv3 base score `9.8` for `CVE-2022-34169`. Since Apache Xalan project is dormant therefore there are no plans to fix the CVE.

### DIFF
--- a/pac4j-saml/pom.xml
+++ b/pac4j-saml/pom.xml
@@ -22,7 +22,6 @@
         velocity 1.7. It can be removed as soon as velocity 1.7 is not used anymore-->
         <commons-collections.version>3.2.2</commons-collections.version>
         <cryptacular.version>1.2.4</cryptacular.version>
-        <xalan.version>2.7.2</xalan.version>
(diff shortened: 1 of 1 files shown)


mmoayyed:
How does this affect the SAML functionality?

mmoayyed:
Furthermore, the CVE says:

> The Apache Xalan Java XSLT library is vulnerable to an integer truncation issue when processing malicious XSLT stylesheets. This can be used to corrupt Java class files generated by the internal XSLTC compiler and execute arbitrary Java bytecode.

How exactly does this affect pac4j, and why should we care about malicious XSLT stylesheets when pac4j never processes any XSLT stylesheets and has nothing to do with that? 

Please demonstrate, with a test case, (not a description, an actual compilable passable reproducer) that this issue is real for pac4j and not just some random CVE.

leleuj:
Closed due to inactivity.

PR #2318, opened 2022-08-25
Type safe casts in CommonProfile and conversion of nested attributes for SAML

SiMiKohl:
Pull Request for problem described here: https://groups.google.com/g/pac4j-dev/c/Pn4d2wBIk8M


leleuj:
I have made several changes today. Can you sync with master? Thanks

This looks good to me even I fear side effects.

SiMiKohl:
I understand and even share the concerns very much. My thoughts on this are:

- Hard casting tends to be very error-prone. Even if the information is now lost whether the attribute has the appropriate data type, it is rather difficult to constantly have to consider the corresponding ClassCastExceptions. If the type would still be of interest, then the type can be determined and checked via the attribute map.
- The handling of lists in the converters is very probably intended only for SAML as far as I understand. Since here the first element of the list with appropriate type is already returned, it corresponds more to the normal treatment if otherwise also a conversion is attempted.

Of course other ideas for a solution are welcome. This way there is at least no more error when using the corresponding getter if a SAML response contains e.g. a gender attribute. Also, in my eyes, only this way a meaningful value can be returned.

leleuj:
Will merge in 5.5.0. Thanks for your contribution.

PR #2367, opened 2022-10-04
Allow SAML2 configuration to specify ACS and Request Initiator URLs

mmoayyed:
The previous/existing behavior is kept; if URLs are undefined, callback URL is used.


leleuj:
You can actually set the callback URL to whatever you want. Why do you need a different value for the ACS URL? And for the `requestInitiatorLocatiion`?

mmoayyed:
In CAS, the callback URL must always be the login endpoint with the client-name. We'd like to keep that as is, but instruct the IDP to use a separate URL for ACS in particular, in scenarios where SAML2 IDP needs or only supports IDP-initiated. This type of change will require follow-up dev, but the building blocks here allow it to be separate from the main login flow, in CAS.

leleuj:
OK. Good. Merging

PR #2369, opened 2022-10-06
Support OpenSAML v5

mmoayyed:
- XMLSec library is now removed
- The references to shibboleth-support-utilities module are now fixed
- Building against OpenSAML v5 SNAPSHOT until releases are made available.
- PMD plugin updated to handle JDK 17 better.
- Constructing the velocity engine is now handled directly.


leleuj:
Thanks. I'll fix the PMD failure after.

leleuj:
I fixed the PMD issues, but noticed there is a test failure in `pac4j-saml-opensamlv5` you will need to take a look at:

```java
[INFO] Running org.pac4j.saml.crypto.KeyStoreCredentialProviderTests
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.051 s - in org.pac4j.saml.crypto.KeyStoreCredentialProviderTests
[INFO] 
[INFO] Results:
[INFO] 
Error:  Failures: 
Error:    VelocityEngineFactoryTest.defaultProperties:19 expected:<org.apache.velocity.runtime.resource.loader.StringResourceLoader> but was:<null>
Error:  Errors: 
Error:    PostSAML2ClientTests.testCustomSpEntityIdForPostBinding:55 » SAML Error encodi...
Error:    PostSAML2ClientTests.testForceAuthIsSetForPostBinding:82 » SAML Error encoding...
Error:    PostSAML2ClientTests.testRelayState:100 » SAML Error encoding saml message
Error:    PostSAML2ClientTests.testSetComparisonTypeWithPostBinding:90 » SAML Error enco...
Error:    PostSAML2ClientTests.testStandardSpEntityIdForPostBinding:69 » SAML Error enco...
[INFO] 
Error:  Tests run: 91, Failures: 1, Errors: 5, Skipped: 0
```

mmoayyed:
Sure. Will do.

leleuj:
I have done a manual test on the `pac4j-saml-opensamlv5` module and I get the following error:

```java
2022-10-06T16:26:00.994+02:00 ERROR 19180 --- [nio-8080-exec-9] org.apache.velocity.loader               : ResourceManager: unable to find resource '/templates/saml2-post-binding.vm' in any resource loader.
2022-10-06T16:26:00.997+02:00 ERROR 19180 --- [nio-8080-exec-9] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed: org.pac4j.saml.exceptions.SAMLException: Error encoding saml message] with root cause

org.apache.velocity.exception.ResourceNotFoundException: Unable to find resource '/templates/saml2-post-binding.vm'
	at org.apache.velocity.runtime.resource.ResourceManagerImpl.loadResource(ResourceManagerImpl.java:465) ~[velocity-engine-core-2.3.jar:2.3]
	at org.apache.velocity.runtime.resource.ResourceManagerImpl.getResource(ResourceManagerImpl.java:346) ~[velocity-engine-core-2.3.jar:2.3]
	at org.apache.velocity.runtime.RuntimeInstance.getTemplate(RuntimeInstance.java:1677) ~[velocity-engine-core-2.3.jar:2.3]
	at org.apache.velocity.app.VelocityEngine.mergeTemplate(VelocityEngine.java:284) ~[velocity-engine-core-2.3.jar:2.3]
	at org.pac4j.saml.transport.Pac4jHTTPPostEncoder.postEncode(Pac4jHTTPPostEncoder.java:122) ~[pac4j-saml-opensamlv5-5.7.0-SNAPSHOT.jar:na]
	at org.pac4j.saml.transport.Pac4jHTTPPostEncoder.doEncode(Pac4jHTTPPostEncoder.java:93) ~[pac4j-saml-opensamlv5-5.7.0-SNAPSHOT.jar:na]
	at org.opensaml.messaging.encoder.AbstractMessageEncoder.encode(AbstractMessageEncoder.java:49) ~[opensaml-messaging-api-5.0.0-SNAPSHOT.jar:na]
	at org.pac4j.saml.profile.impl.AbstractSAML2MessageSender.sendMessage(AbstractSAML2MessageSender.java:95) ~[pac4j-saml-opensamlv5-5.7.0-SNAPSHOT.jar:na]
	at org.pac4j.saml.sso.impl.SAML2WebSSOProfileHandler.send(SAML2WebSSOProfileHandler.java:30) ~[pac4j-saml-opensamlv5-5.7.0-SNAPSHOT.jar:na]
	at org.pac4j.saml.sso.impl.SAML2WebSSOProfileHandler.send(SAML2WebSSOProfileHandler.java:16) ~[pac4j-saml-opensamlv5-5.7.0-SNAPSHOT.jar:na]
	at org.pac4j.saml.redirect.SAML2RedirectionActionBuilder.getRedirectionAction(SAML2RedirectionActionBuilder.java:40) ~[pac4j-saml-opensamlv5-5.7.0-SNAPSHOT.jar:na]
```

The CI tests failures are certainly relevant.

boris-petrov:
Thanks for doing the work with using the newest version of OpenSAML! However, depending on SNAPSHOT releases (in production artifacts) causes all kinds of problems - the SNAPSHOT releases change often so builds will be non-deterministic. Also, in my Gradle setup I have set `failOnNonReproducibleResolution` which causes my build to fail with a bunch of errors like this - `Could not resolve net.shibboleth:shib-velocity:9.0.0-SNAPSHOT: Resolution strategy disallows usage of changing versions`. These problems mean that I simply cannot use the new `pac4j-saml-opensamlv5` artifact.

In addition, the previous package `pac4j-opensaml` has been deprecated so I have to put `@SuppressWarnings("deprecation")` here and there. Not the biggest issue but I wanted to mention it.

In the end I think it's better if you release perhaps SNAPSHOT/RC releases when your transitive dependencies are such and not deprecate the old package before there is a stable release of the new one.

Otherwise thanks for the great work on `pac4j`!

leleuj:
I understand your point of view.

"deprecated" was meant to draw your attention on the fact that in the next major version of pac4j (v6), it can be removed. In fact, it will be. Only the deprecated `pac4j-javaee` module will remain in pac4j v6.

`pac4j-saml-opensamlv5` is for early adopters or deployers who align with the ETA of pac4j v6 / CAS v7 (February 2023).
You should still use `pac4j-saml` or is there a problem with the current implem?

Of course, I'll cut a RC as soon as possible.

leleuj:
BTW, in pac4j v6, `pac4j-saml-opensamlv5` will be renamed as `pac4j-saml` as it will be the only SAML support module.

boris-petrov:
Thanks for the answer! Well, the only reason I tried `pac4j-saml-opensamlv5` in the first place was because in the [documentation](https://www.pac4j.org/docs/clients/saml.html) it says that the other one is deprecated so I thought that it's the newer thing. Also, using `pac4j-saml` leads to deprecation warnings in my code which I also thought meant that I really should migrate to the other one.

It's strange - if `pac4j-saml-opensamlv5` will be renamed back to `pac4j-saml`... why is `pac4j-saml` deprecated? In pac4j v6 it will just contain a new version of OpenSAML in it and that's it (which is fine according to semver). It's not "deprecated" in that sense, on the contrary - it's what will exist in pac4j v6.

leleuj:
Yes, this is tricky. I'll communicate on that.
On the other side, if I had renamed the `pac4j-saml` module to `pac4j-saml-opensamlv4` and added a new `pac4j-saml` dependency based on OpenSAML v5 (and JDK 17), most deployers would have upgraded to the new `pac4j-saml` module without being even aware of it until the compilation issue.
*I can have the lesser of two evils.*

boris-petrov:
No, I think creating a new module for people to try out is totally fine. Also it's totally fine to replace the old module with the new one when you release the new major version of pac4j. The only "problematic" thing I believe is deprecating the old one - because people have no "correct" solution. They either use a deprecated package (and have deprecation warnings in their code) or use a development/non-final package.

PR #2371, opened 2022-10-06
Allow service-loader API to locate SAML2 metadata generators

mmoayyed:
This should allow pac4j to provide alternative implementations for metadata management based on JPA, Mongo, etc.

### DIFF
--- a/documentation/docs/clients/saml.md
+++ b/documentation/docs/clients/saml.md
@@ -189,6 +189,10 @@ You can generate the SP metadata in two ways:
 - either programmatically using the `SAML2Client`: `String spMetadata = client.getServiceProviderMetadataResolver().getMetadata();`
 - or by defining the appropriate configuration: `cfg.setServiceProviderMetadata(new FileSystemResource("/tmp/sp-metadata.xml"));`
 
+Furthermore, bootstrapping the SAML configuration harness can be done using Java's `ServiceLoader` API. In order to be discovered,
+you must create a provider configuration file: `META-INF/services/org.pac4j.saml.util.ConfigurationManager`. The content of the file is the
+fully qualified class name of the SPI implementation: `com.example.impl.MyConfigurationManager`.
+
 ## 3.1) Metadata and Keystore configuration:
 
 - For service provider metadata generation, *pac4j* offers the following components:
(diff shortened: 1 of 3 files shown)


leleuj:
Three questions:
- can you drop a line in the release notes?
- I guess you don't want to make that change in the "old" `pac4j-saml` module which is now deprecated and will be removed next year, right?
- we also have a `ConfigurationManager` in the SAML support: would it make sense to incorporate your change with it?

mmoayyed:
- I updated the docs and the release notes. 
- Could do it, sure but to me that module is dead and in pure maintenance mode given OpenSAML v4 is more or less EOL. News features should probably stick to the v5 module. If you'd like, I can certainly *backport*.
- If I understand correctly, the `ConfigurationManager` is entirely separate from here. One could use this SPI trick for `ConfigurationManager` or  `SAML2MetadataGenerator`, or both. I updated the docs to note this. I think what you're saying is, we could consider modifying the `SAML2MetadataGenerator`, so that it would allow one to also configure and define a `SAML2MetadataGenerator`? We could go that route, but I feel like it might be best to keep them separate. The change here mainly will be used in CAS to provide alternative stores for metadata management of the SP, and if I mix this with the configuration manager, I would also have to account for that implementation as well. Separation is a good thing here.

mmoayyed:
By the way, do you think pac4j should keep alternative impls of `SAML2MetadataGenerator`? based on JPA, Mongo, etc as a separate module? I have seen the request a few times on the list. My plan is to ultimately build support for this in CAS, and I would prefer if these alternative impls were offered by Pac4j and only configured in CAS, rather than CAS directly owning them. What do you think?

leleuj:
Thanks for the docs and release notes

No need to backport, let's focus our efforts on none-deprecated modules.

Yes, that was my question: can we merge the `ConfigurationManager` with what you propose in this PR? But this really was a naive question and I understand you prefer to keep them separated.

If new implementations (based on JPA, Mongo, etc) are created, it makes sense to host them in pac4j rather than in CAS as it could be used by other pac4j users.

mmoayyed:
Wonderful. I'll follow up on the dev list. Thanks again!

PR #2372, opened 2022-10-07
Support Mongo for SAML2 SP metadata

mmoayyed:
This PR adds `SAML2MongoMetadataGenerator` to the pac4j SAML2 portfolio. This feature is only available to pac4j-SAML that is based on OpenSAML v5. 

Surrounding components have been re-worked and refactored to support the change. Mongo dependencies are also made as optional, and need to be re-declared by users when used. 

Typical use case:

```java
var configuration = new SAML2Configuration();
...
configuration.setMetadataGenerator(new SAML2MongoMetadataGenerator(this.mongoClient));
....
configuration.init();
```

### DIFF
--- a/pac4j-saml-opensamlv5/pom.xml
+++ b/pac4j-saml-opensamlv5/pom.xml
@@ -17,9 +17,6 @@
         <joda-time.version>2.11.2</joda-time.version>
         <velocity.version>2.3</velocity.version>
         <xmlsec.version>3.0.1</xmlsec.version>
-        <!-- The dependency to commons-collections 3.2.2 is explicitly made to replace 3.2.1 (vulnerable) used by
-        velocity 1.7. It can be removed as soon as velocity 1.7 is not used anymore-->
-        <commons-collections.version>3.2.2</commons-collections.version>
         <cryptacular.version>1.2.5</cryptacular.version>
         <hazelcast.version>5.1.3</hazelcast.version>
         <java.version>17</java.version>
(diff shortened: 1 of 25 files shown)


mmoayyed:
Docs and release notes now updated.

pac4j-saml-opensamlv5/pom.xml
```diff
@@ -173,6 +181,20 @@
...
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.mongodb</groupId>
+            <artifactId>bson</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.mongodb</groupId>
+            <artifactId>mongodb-driver-core</artifactId>
```
leleuj, line 196: Why is this dependency not `optional`?

leleuj: I thought that this dependency should be explicitly added by the pac4j user.

leleuj:
LGTM. One question though about the `mongodb-driver-core` dependency...

pac4j-saml-opensamlv5/pom.xml
mmoayyed: You're right. Should be fixed and optional now.

mmoayyed:
I will also work on support for JPA in this area, and I think that should be enough for now. Then, will build support for both options in CAS v7.

leleuj:
OK. Good. Merging

PR #2373, opened 2022-10-07
Support JDBC for SAML2 SP metadata

mmoayyed:
Service provider metadata can alternatively be stored and managed via relational databases. The generator expects an instance of `JdbcTemplate` to work with your relational database. By default, the expected database table is named `sp-metadata` with two columns, `entityId` and `metadata` (which should allow for large text values). Metadata is base-64 encoded/decoded by this generator on save and/or fetch operations.

```java
var configuration = new SAML2Configuration();
...
var generator = new SAML2JdbcMetadataGenerator(jdbcTemplate, entityId);
generator.setTableName(...);
...
configuration.setMetadataGenerator(generator);
...
configuration.init();
```

There is also a related change to ensure metadata can be stored per entity id. This change also affects the Mongo implementation, and in general allows for pac4j to manage multiple SPs with multiple SAML2 clients.


leleuj:
LGTM. Merging

PR #2383, opened 2022-10-13
Delegating construction to CommonHelper.getConstructor()

m-froehlich:
Utilizing CommonHelper.getConstructor() to construct instance of ConfigFactory.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/config/ConfigBuilder.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/config/ConfigBuilder.java
@@ -19,14 +19,7 @@ public final class ConfigBuilder {
         try {
             logger.info("Build the configuration from factory: {}", factoryName);
 
-            var tccl = Thread.currentThread().getContextClassLoader();
-            final Class<ConfigFactory> clazz;
-            if (tccl == null) {
-                clazz = (Class<ConfigFactory>) Class.forName(factoryName);
-            } else {
-                clazz = (Class<ConfigFactory>) Class.forName(factoryName, true, tccl);
-            }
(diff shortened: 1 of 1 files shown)


pac4j-core/src/main/java/org/pac4j/core/config/ConfigBuilder.java
```diff
@@ -19,14 +19,7 @@ public synchronized static Config build(final String factoryName, final Object..
...
 
-            var tccl = Thread.currentThread().getContextClassLoader();
-            final Class<ConfigFactory> clazz;
-            if (tccl == null) {
-                clazz = (Class<ConfigFactory>) Class.forName(factoryName);
-            } else {
-                clazz = (Class<ConfigFactory>) Class.forName(factoryName, true, tccl);
-            }
-            final var factory = clazz.getDeclaredConstructor().newInstance();
+            final var factory = (ConfigFactory) org.pac4j.core.util.CommonHelper.getConstructor( factoryName ).newInstance();
```
mmoayyed, line 22: Is there a reason you're using the full package name here?

mmoayyed:
```suggestion
            final var factory = (ConfigFactory) CommonHelper.getConstructor(factoryName).newInstance();
```

m-froehlich: Not really. I wanted to keep my changes as locally as possible. Feel free to use an import.

leleuj: OK. Will fix

PR #2487, opened 2023-01-26
Capture relay-state for logout/decoding ops

mmoayyed:


mmoayyed:
After this change is in, I think pac4j is in a good enough shape for the next RC if you want to move fwd. 

I will continue to test, and it would also be OK if you wanted to shift the release into next week.

PR #2488, opened 2023-01-27
Resolve SAML2 idp metadata on initialization

mmoayyed:
This pull request also makes sure SAML2 message contexts always have access to an active `CallContext`, to avoid NPEs. It also removes a number of deprecated API calls from OpenSAML, and uses Lombok in certain cases instead of manual logger/getter/setter constructs.

### DIFF
--- a/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/client/SAML2Client.java
@@ -177,6 +177,7 @@ public class SAML2Client extends IndirectClient {
 
     protected void initIdentityProviderMetadataResolver() {
         this.identityProviderMetadataResolver = this.configuration.getIdentityProviderMetadataResolver();
+        this.identityProviderMetadataResolver.resolve();
     }
 
     protected void initDecrypter() {
--- a/pac4j-saml/src/main/java/org/pac4j/saml/context/SAML2ContextProvider.java
+++ b/pac4j-saml/src/main/java/org/pac4j/saml/context/SAML2ContextProvider.java
@@ -56,7 +56,7 @@ public class SAML2ContextProvider implements SAMLContextProvider {
 
     @Override
     public final SAML2MessageContext buildServiceProviderContext(final CallContext ctx, final SAML2Client client) {
-        val context = new SAML2MessageContext();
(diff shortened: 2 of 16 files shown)


leleuj:
@mmoayyed I made one more SAML login/logout test and it works. Can I cut the release?

mmoayyed:
Give me about an hour, if you don't mind and I will post back. I think we are in pretty good shape, but I want to run a few more tests.

mmoayyed:
[ It will likely be less than 1 hour; just trying to account for the worst case scenario :) ]

mmoayyed:
We are good to proceed. Thank you! It's possible we/I might find more stuff, but this is good enough for now.

leleuj:
Excellent! Starting the release...

mmoayyed:
By the way, I suppose you would also release https://github.com/pac4j/spring-webmvc-pac4j based on RC5?

leleuj:
Yes

leleuj:
I'm trying to. Sonatype is very slow (this morning for me).

mmoayyed:
I understand. It's been like that for the past few days I think. Thank you, as ever.

leleuj:
Learning patience ;-)

leleuj:
https://oss.sonatype.org/ is almost unreachable. Will try in a few hours or on Monday...

mmoayyed:
No problem.

leleuj:
image

😞😞😞😞😞😞😞😞😞😞😞😞😞😞😞😞😞😞

mmoayyed:
wow, that is super unfortunate. what can I do to help?

mmoayyed:
It looks like you were able to close the repository on staging, which is very good. publishing it might be tricky; keep trying :)

leleuj:
If you know how to contact the guys from the Sonatype Nexus UI, let them know that they need to restart the servers! ;-)

mmoayyed:
Sure. We should get in touch with Julien. You can find him here: https://issues.sonatype.org/browse/OSSRH-88250

Lots of other folks are having similar problems; you should comment on that thread and he's usually quick to follow up.

leleuj:
Thanks

leleuj:
I have eventually released pac4j v6.0.0-RC5 and spring-webmvc-pac4j v8.0.0-RC5.

PR #2495, opened 2023-01-31
Adding support for ByteArrayResource in SpringResourceLoader

ilgrosso:
This PR adds support to work with Spring's `ByteArrayResource`, to allow the `load()` method to read at least once the underlying byte array.

The current implementation, in fact, is catching the `FileNotFoundException` raised by `ByteArrayResource#lastModified`, thus resulting into the `hasChanged()` method always returing `false` and the `load()` method never reading the byte array.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/resource/SpringResourceLoader.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/resource/SpringResourceLoader.java
@@ -1,10 +1,12 @@
 package org.pac4j.core.resource;
 
+import java.util.concurrent.atomic.AtomicBoolean;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


ilgrosso:
Please note that I haven't provided any specific test case because I am unsure whether there is intention to support `ByteArrayResource` at all.
If such interest is confirmed, I'll add tests to this PR.
Thanks.

leleuj:
Why do you need a `ByteArrayResource`? And not a file or URL?

ilgrosso:
@leleuj because in my use case (for Apache Syncope) I am getting the actual metadata content from a database table, along with other conf params for the given IdP.

leleuj:
OK. This makes sense. You can add a test to confirm the behavior.

ilgrosso:
@leleuj test added

leleuj:
Excellent! Merging

ilgrosso:
Thank you @leleuj !

PR #2507, opened 2023-02-15
OIDC: allow to explicitly specify token auth methods supported by client

mloesch:
Background: solely relying on the first auth method supported by the server can be problematic in case the first method is e.g. `PRIVATE_KEY_JWT`, which requires a private key to be present on the client, which may not be the case.

With this change, we allow to optionally also specify a list of auth methods supported by the client. If present, it will be respected when selecting the first supported auth method from the list returned by the IDP.

### DIFF
--- a/pac4j-oidc/src/test/java/org/pac4j/oidc/credentials/authenticator/OidcAuthenticatorTest.java
+++ b/pac4j-oidc/src/test/java/org/pac4j/oidc/credentials/authenticator/OidcAuthenticatorTest.java
@@ -1,0 +1,80 @@
+package org.pac4j.oidc.credentials.authenticator;
+
+import static org.junit.Assert.assertEquals;
+
+import com.nimbusds.oauth2.sdk.auth.ClientAuthenticationMethod;
+import com.nimbusds.oauth2.sdk.id.Issuer;
+import com.nimbusds.openid.connect.sdk.SubjectType;
+import com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.List;
+import java.util.Set;
+import org.junit.Assert;
+import org.junit.Test;
+import org.pac4j.core.exception.TechnicalException;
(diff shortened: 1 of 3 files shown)


pac4j-oidc/src/test/java/org/pac4j/oidc/credentials/authenticator/OidcAuthenticatorTest.java
```diff
@@ -0,0 +1,80 @@
...
+import java.net.URISyntaxException;
+import java.util.List;
+import java.util.Set;
+import org.junit.Assert;
+import org.junit.Test;
+import org.pac4j.core.exception.TechnicalException;
+import org.pac4j.oidc.client.OidcClient;
+import org.pac4j.oidc.config.OidcConfiguration;
+
+public class OidcAuthenticatorTest {
```
leleuj, line 19: Can you add some javadoc to the class header (`@author` and `@since` at least)?

leleuj (changes requested):
Very good! See my feedback.

Another PR must also be submitted for the master branch.

pac4j-oidc/src/test/java/org/pac4j/oidc/credentials/authenticator/OidcAuthenticatorTest.java
mloesch: Yes sure, fixed in cfee7886c869ca4cb61c9cf820302ecbd6b3a9e2 :)

leleuj: Good. Please submit the same PR for the master branch.

mloesch: Opened #2510 for porting the changes to the master branch

leleuj: Excellent! Thanks

PR #2512, opened 2023-02-21
OIDC: introduce specific exceptions

mloesch:
We would like to be able to react differently to the various OIDC-related error cases


leleuj:
Good idea!

PR #2539, opened 2023-03-18
Create DefaultUrlResolver.java

wuwx:
Add ContextPath to url if application run with ContextPath value


leleuj:
You can't do that. You cannot cast to `JEEContext` in the `DefaultUrlResolver`.

Though, you can use the `getPath()` method from the `WebContext` interface.

PR #2565, opened 2023-04-20
Initialize SAMLOutboundProtocolMessageSigningHandler before invoke

ilgrosso:
Avoids

```
org.pac4j.saml.exceptions.SAMLException: net.shibboleth.shared.component.UninitializedComponentException: Unidentified Component has not yet been initialized and cannot be used.
        at org.pac4j.saml.profile.impl.AbstractSAML2MessageSender.invokeOutboundMessageHandlers(AbstractSAML2MessageSender.java:153) ~[pac4j-saml-6.0.0-RC6.jar!/:?]
```


ilgrosso:
Wondering if encoders created in the `getMessageEncoder()` methods might require the same `.initialize()` call before use.

PR #2601, opened 2023-06-02
Compatible with proxy scenarios

cabbage89:
When using a proxy to the backend, read the X-Forwarded-Proto request header

Before submitting any pull request, please read the contribution guide: https://www.pac4j.org/docs/contribute.html

### DIFF
--- a/pac4j-javaee/src/main/java/org/pac4j/jee/context/JEEContext.java
+++ b/pac4j-javaee/src/main/java/org/pac4j/jee/context/JEEContext.java
@@ -127,12 +127,17 @@ public class JEEContext implements WebContext {
 
     @Override
     public String getScheme() {
-        return this.request.getScheme();
+        return this.getProto(this.request);
     }
 
     @Override
     public boolean isSecure() {
-        return this.request.isSecure();
+        return this.request.isSecure() || "https".equalsIgnoreCase(getProto(request));
+    }
+
+    private String getProto(HttpServletRequest request) {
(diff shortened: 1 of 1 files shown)


cabbage89:
https://groups.google.com/g/pac4j-dev/c/s3eEdMLJ25g

mmoayyed:
I don't think we are looking at the 5.4.x release any more. You should consider repositioning your PR to target master.

papegaaij:
Also, this seems incorrect. It is the responsibility of the servlet container to interpret headers like `X-Forwarded-Proto`. Often, this needs to be enabled explicitly. These headers can be inserted by a malicious client and spoof the server into taken a wrong scheme.

@cabbage89 you probably need to check the documentation of whatever servlet container you are using. For example, on WildFly, you have to set `proxy-address-forwarding="true"` on the `http-listener`.

leleuj:
Yes, you should target both the master and 5.7.x branches.

leleuj:
And yes, this is a bit too hardcoded to be accepted "as is".

Though, thanks to the `WebContextFactory`,  you should be able to instantiate any `WebContext` you want so the idea is certainly to open the `JEEContext` to allow this customisation.

cabbage89:
As you said, I have solved my problem in another way, sorry to bother you, I will submit PR more carefully in the future

leleuj:
No problem. PRs are meant to be discussed.

PR #2629, opened 2023-07-10
Update SAML2AuthnRequestBuilder.java : I have fixed the Issue of Asse…

rohitpawar2811:
According to this,It is fixed in https://groups.google.com/g/pac4j-users/c/1Lxn5lq4SVk/m/RkYL_70dAwAJ in v5.1.4
but we are using 4.5.x. 
because we are using Java 8 can you build this change and make a minor cut?


mmoayyed:
No we cannot. Please upgrade or apply the patch manually.

PR #2635, opened 2023-07-19
Allow oidc metadata resolver to support hostname verification

mmoayyed:
...and SSL factories directly instead of using reflection to create factories.


pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
```diff
@@ -38,7 +36,7 @@
  */
 @Getter
 @Setter
-@ToString(exclude = {"secret", "providerMetadata"})
```
mmoayyed, line 41: Field no longer exists

PR #2668, opened 2023-09-07
Prevent NPE if oidcCredentials is null

jeremie-lesage:
Quick fix to Prevent NPE if oidcCredentials is null and isIncludeAccessTokenClaimsInProfile is true


leleuj:
Backported to the future v5.7.2: https://github.com/pac4j/pac4j/commit/1e3f970cfc4b5c3a57ff470d225aa68e4ae6007d

PR #2670, opened 2023-09-10
Allow serializing oidc credentials/profile into JSON

mmoayyed:
`OidcCredentials` class contains a number of fields from nimbus, such as `JWT`, `AccessToken`, etc that are not easily (or at all) serializable into JSON. Workarounds must be put in place to handle correct serialization of this type, including custom serializer/deserializers, mixins, etc. This pull request switches those field types to plain strings or `Map`s, and then allows for APIs that can reconstruct the actual needed type.

You'll find test cases here that demonstrate JSON serialization with Jackson's `ObjectMapper` with and without typing enabled.

PS It's likely that this problem also exists with other credential and user profile types. We can look at those separately.


leleuj:
LGTM.

Two comments though:

- Currently, the `jackson-databind` dependency is `optional` (Maven) in the `pac4j-core` module: I think it should become mandatory
- Can you drop a line in the release notes? Thx

mmoayyed:
Sure thing, done.

PR #2692, opened 2023-10-04
Ensure TokenCredentials are serializable to JSON

hsartoris-bard:
See also #2670.

Currently, the second test will fail unless additional typing information is introduced. Happy to tweak as needed.

### DIFF
--- a/pac4j-cas/src/main/java/org/pac4j/cas/profile/CasProxyProfile.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/profile/CasProxyProfile.java
@@ -17,7 +17,7 @@ import java.io.Serial;
  * @author Jerome Leleu
  * @since 1.4.0
  */
-@EqualsAndHashCode
+@EqualsAndHashCode(callSuper = true)
 public class CasProxyProfile extends CasProfile {
 
     @Serial
--- a/pac4j-cas/src/main/java/org/pac4j/cas/profile/CasProxyProfile.java
+++ b/pac4j-cas/src/main/java/org/pac4j/cas/profile/CasProxyProfile.java
@@ -1,5 +1,6 @@
 package org.pac4j.cas.profile;
 
+import lombok.EqualsAndHashCode;
 import lombok.val;
 import org.apereo.cas.client.authentication.AttributePrincipal;
 
@@ -16,6 +17,7 @@ import java.io.Serial;
  * @author Jerome Leleu
  * @since 1.4.0
  */
+@EqualsAndHashCode
 public class CasProxyProfile extends CasProfile {
 
     @Serial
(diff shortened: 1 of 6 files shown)


leleuj:
Good. What's your fix proposal?

hsartoris-bard:
https://github.com/pac4j/pac4j/pull/2692/commits/c2cff70ad78995ffae92b448875529f52a644361 provides a fix for the first test, which otherwise fails.

https://github.com/pac4j/pac4j/pull/2692/commits/16d55922b2a424b9703152100e835024ca857fbd addresses the latter, by explicitly annotating `UserProfile` with `@JsonTypeInfo`. I also added `@EqualsAndHashCode` on `BasicUserProfile`, because otherwise equality checking on `TokenCredentials` doesn't work, but this is optional - I can either check that the tokens match between the original and the deserialized credentials, or set `callSuper` to false on TokenCredentials. My sense is that, semantically, it doesn't make sense to say that two TokenCredentials objects are equal simply if their tokens are equal, even if they contain differing UserProfiles.

leleuj:
Can you check why the build fails: https://github.com/pac4j/pac4j/actions/runs/6470734206/job/17595797077?pr=2692?

hsartoris-bard:
Ah yes, needed to also override `equals` on CasProxyProfile

leleuj:
The build still fails: `Error:  Medium: org.pac4j.kerberos.profile.KerberosProfile doesn't override org.pac4j.core.profile.BasicUserProfile.equals(Object) [org.pac4j.kerberos.profile.KerberosProfile] At KerberosProfile.java:[line 1] EQ_DOESNT_OVERRIDE_EQUALS`

pac4j-cas/src/main/java/org/pac4j/cas/profile/CasProxyProfile.java
```diff
@@ -16,6 +17,7 @@
  * @author Jerome Leleu
  * @since 1.4.0
  */
+@EqualsAndHashCode
```
mmoayyed, line 20: Should this not be `@EqualsAndHashCode(callSuper = true)`?

pac4j-core/src/main/java/org/pac4j/core/profile/UserProfile.java
```diff
@@ -12,6 +14,7 @@
  * @author Jerome Leleu
  * @since 4.0.0
  */
+@JsonTypeInfo(use=JsonTypeInfo.Id.CLASS, include=JsonTypeInfo.As.PROPERTY, property="class")
```
mmoayyed, line 17:
```suggestion
@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "class")
```

mmoayyed (changes requested):

hsartoris-bard:
good points; changes made. I was able to run the tests locally by excluding the SAML test suite, which appears to expect some service to be running that I do not have.

leleuj:
Can you sync with master? I have excluded two failing SAML tests.

hsartoris-bard:
Done; tests all pass for me now.

PR #2706, opened 2023-10-17
Track credential source in extractor

mmoayyed:
UsernamePasswordCredentials is used in many places by different extractors. It's useful to know how the credential was extracted and track its source so that follow-up processes may be able to determine if they can accept and authenticate the credential given its source.

PS Can update release notes + tests if you think this is a good approach to continue.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/credentials/CredentialSource.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/credentials/CredentialSource.java
@@ -3,7 +3,7 @@ package org.pac4j.core.credentials;
 /**
  * The credential source indicates how the credential was extracted.
  *
- * @author Jerome LELEU
+ * @author Misagh Moayyed
  * @since 6.0.0
  */
 public enum CredentialSource {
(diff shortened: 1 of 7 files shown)


pac4j-core/src/main/java/org/pac4j/core/credentials/CredentialSource.java
```diff
@@ -0,0 +1,11 @@
+package org.pac4j.core.credentials;
+
+/**
+ * The credential source indicates how the credential was extracted.
+ *
+ * @author Jerome LELEU
```
leleuj, line 6: This is not me ;-)

leleuj:
I see your point. This makes sense. I'm just hesitating on the source type: is an `enum` enough to anticipate further needs? Wouldn't a `String` be better?

mmoayyed:
Good point. I was going for type-safety and my main scenario here deals with extracting stuff from the web (where the source is no more than 3-4 types) but I suppose an extractor would want to use a "free form" source for more general purposes. Maybe we switch the type to String, but allow the enum to exist to be used for relevant extractors, or is that too much?

leleuj:
I think it's a good compromise.

PR #2707, opened 2023-10-17
Allow form extraction to specify extraction mode

mmoayyed:
Presently, the form extractor extracts parameters from both query-string and request body. While it likely should only extract values from the request body (i.e. FORM POST), in this PR this behavior is not changed and instead, we provide options to specify the extraction mode. This should keep backward compatibility. 

Will add tests+docs if this looks good.

### DIFF
--- a/pac4j-core/pom.xml
+++ b/pac4j-core/pom.xml
@@ -41,6 +41,10 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.httpcomponents.client5</groupId>
+            <artifactId>httpclient5</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-core</artifactId>
             <optional>true</optional>
--- a/pac4j-core/pom.xml
+++ b/pac4j-core/pom.xml
@@ -41,10 +41,6 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.apache.httpcomponents.client5</groupId>
-            <artifactId>httpclient5</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-core</artifactId>
             <optional>true</optional>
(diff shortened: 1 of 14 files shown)


pac4j-core/src/test/java/org/pac4j/core/credentials/extractor/FormExtractorTests.java
```diff
@@ -0,0 +1,37 @@
...
+import org.pac4j.core.context.CallContext;
+import org.pac4j.core.context.MockWebContext;
+
+import static org.junit.Assert.assertTrue;
+
+/**
+ * This is {@link FormExtractorTests}.
+ *
+ * @author Misagh Moayyed
+ * @since 7.0.0
```
leleuj, line 14: We are only at version 7 in pac4j ;-)

pac4j-core/pom.xml
```diff
@@ -40,6 +40,10 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
+        <dependency>
```
leleuj, line 43: I'm a bit reluctant to add this new library just for the `UriBuilder`...

leleuj (changes requested):

pac4j-core/pom.xml
mmoayyed: I see, sure. Do you want to add a utility method somewhere that can be used here? I really did not want to duplicate the parameter parsing logic. It's very difficult to get it right!

pac4j-core/src/test/java/org/pac4j/core/credentials/extractor/FormExtractorTests.java
leleuj: Version 6 ! Sorry

pac4j-core/pom.xml
mmoayyed: Ah, absolutely sure but that only does the checking. Here we need the extraction. IOW I can check to see if a parameter is on the query string, but I'd have to parse the query string to extract it, and I find that prone to error. If you feel like it, I can still do it nonetheless.

leleuj: I think you can duplicate the logic I used in CAS: you can get the parameter with `webContext.getRequestParameter` and check if it is in the query string...

PR #2726, opened 2023-11-03
parameter-less ctr for DefaultRolesAuthorizationGenerator and additional setter

polli62:
To be able to use the DefaultRolesPermissionsAuthorizationGenerator within shiro.ini, a parameter-less ctr is necessary, that default roles can be defined like this:

```
roleGenerator = org.pac4j.core.authorization.generator.DefaultRolesPermissionsAuthorizationGenerator
roleGenerator.defaultRoles = role_a,role_b,role_c
myClient.authorizationGenerator = $roleGenerator
```
The PR adds the missing ctr + setter


leleuj:
The build fails. Please check it...

leleuj:
The build still fails.

PR #2730, opened 2023-11-06
5.7.x parameter-less-ctr-for-DefaultRolesAuthorizationGenerator-and-additional-setter

polli62:
Added a parameter-less ctr for DefaultRolesAuthorizationGenerator


leleuj:
The build fails.

PR #2845, opened 2024-04-02
Do not create OidcAuthenticator instance in renewUserProfile

ddqof:
In order to use custom `Authenticator` when getting new access token, I suggest to use `OidcAuthenticator` from `getAuthenticator`  instead of creating new instance of `OidcAuthenticator` each time


leleuj:
Thanks for your contribution

PR #2858, opened 2024-04-11
fix(deps): Fix lombok scope which is a compiletime only dependency

Vampire:
Lombok is a pure compiletime dependency and the final artifact should not depend on it.
Having a dependency brings it in needlessly and also causes IDEs to suggest using Lombok in own code, which is bad.
So as the Lombok documentation describes, the scope is changed to provided with this PR.


leleuj:
OK. It makes sense.

PR #2869, opened 2024-04-18
Return only SAML 2.0 support in the sp-metadata.xml

michaguenther:
According to the [latest pac4j SAML documentation](https://www.pac4j.org/docs/clients/saml.html), only SAML 2.0 is supported. In the `sp-metadata.xml` this is defined by protocolSupportEnumeration. Currently, the default configuration claims the support of SAML 1.0, SAML 1.1, and SAML 2.0. The default is changed to SAML 2.0 support only.


leleuj:
Good catch! Merging.

leleuj:
Backported to 5.x: https://github.com/pac4j/pac4j/commit/33bc34ba904b48212d1722e30f2ff35cc4da847d and 4.x: https://github.com/pac4j/pac4j/commit/36b7de84107d82db6314289acd83c1989b069705

PR #2880, opened 2024-04-29
Update Jooby urls in readme and documentation

0rzech:
Jooby maintainer no longer owns jooby.org domain
and the current one is jooby.io. See
https://github.com/jooby-project/jooby/issues/1513 for more details.

Also, link directly to Jooby's pac4j integration
instructions instead of just home page.


leleuj:
Thanks for the fix

PR #2885, opened 2024-05-07
Add `createTokenValidator` method in `OidcOpMetadataResolver` in order to reduce copy-pasted code

ddqof:
I suggest to add new protected method that allows to override `TokenValidator` creation logic in `OidcOpMetadataResolver` subclasses like `AzureAdOpMetadataResolver`


leleuj:
Sure. It's cleaner.

PR #2903, opened 2024-05-23
De-duplicate user profile attribute values

rrenomeron-gov:
I've discovered what I think is an issue with deserialization of the ``SAML2Profile`` class involving the ``authnContexts`` property.  When it is deserialized, Jackson will populate the ``authenticationAttributes`` map in the object normally.  But when it tries to populate the ``authnContexts`` property by calling its setter, it inserts a duplicate value into the ``authenticationAttributes`` map, as it tries to "merge" the multiple values.

As this happens on every read from JSON storage, the ``authenticationAttributes`` map gets really big  really quickly (doubles on every serialization round trip). This causes big issues in my CAS ticket registry storage for a fairly common usage
pattern (user logs in to 5 or 6 different services on a single SSO session).

To fix, I made a small change to ``BasicUserProfile.mergeCollectionAttributes()`` which simply deduplicates multiple values from the two ``Collections``.  I'm thinking there's not a use case out there where somebody needs to have duplicate attribute values.  I've done some testing in my nonproduction deployments, and it fixes the problem.

### DIFF
--- a/pac4j-core/src/main/java/org/pac4j/core/profile/BasicUserProfile.java
+++ b/pac4j-core/src/main/java/org/pac4j/core/profile/BasicUserProfile.java
@@ -159,7 +159,7 @@ public class BasicUserProfile implements UserProfile, Externalizable {
 
     private static <T> Collection<T> mergeCollectionAttributes(final Collection<T> existingCollection, final Collection<T> newCollection)
     {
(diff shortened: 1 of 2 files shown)


leleuj:
Excellent! Thanks

leleuj:
I just cut the v6.0.3 release. It will show up soon in the Maven central repository.

PR #2906, opened 2024-05-29
Fix: StaticOidcOpMetadataResolver should not enforce a discovery URI

mloesch:
Currently it is not possible to use StaticOidcOpMetadataResolver without having a discovery URI on the OIDC configuration, it will fail in [SpringResourceHelper](https://github.com/pac4j/pac4j/blob/master/pac4j-core/src/main/java/org/pac4j/core/resource/SpringResourceHelper.java#L88).

However, since StaticOidcOpMetadataResolver always returns the static OIDC metadata provided during initialization, it should not require a discovery URI.


leleuj:
Thanks for your contribution

PR #2926, opened 2024-06-21
Allow AWS S3 buckets to store SAML2 SP metadata

mmoayyed:
Allow Pac4j to provide an option to store SAML2 SP metadata in S3 buckets.  There is a default bucket that can optionally be created and each object inside the bucket belongs to an SP entity id.

Please note that tests are provided using mocks; writing real integration tests requires something like localstack as a docker container which pac4j is not fit to handle yet, and is beyond the scope of this PR. This functionality has been fully tested in CAS with proper integration tests using the localstack docker image. This feature will be available in CAS 7.1.x.

### DIFF
--- a/documentation/docs/clients/saml.md
+++ b/documentation/docs/clients/saml.md
@@ -284,6 +284,35 @@ You will need to include the following module, at a minimum, in your build:
 </dependency>
 ```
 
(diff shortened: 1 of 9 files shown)


leleuj:
Excellent! I plan to cut the v6.0.4 release tomorrow.

PR #2934, opened 2024-06-26
Force load SAML2 IDP metadata from URLs

mmoayyed:
When SAML2 IDP metadata is fetched from a URL, the following snippet presents an NPE:

```java
saml2Client.init();
var identityProviderMetadataResolver = saml2Client.getIdentityProviderMetadataResolver();
identityProviderMetadataResolver.resolve();
var entity = identityProviderMetadataResolver.getEntityDescriptorElement();
```

URL resources do not always present a reliable up-to-date check, which makes pac4j skip the metadata resolution step since it sees no change. In this changeset, if the metadata resource is a URL, the change detection logic is forced to reload the metadata from the URL and always be up to date.


leleuj:
Isn't it dangerous for performance? It will keep reloading the metadata over and over again, right?

mmoayyed:
That is a valid point, yes. We could consider building some kind of caching of this, but I am concerned if that makes the whole "change detection" logic more complicated?

mmoayyed:
We could also refactor the functionality a bit so the ability to force reload metadata is exposed in the API, and then callers can invoke that when they need to. That might be better.

mmoayyed:
Here is an alternative: https://github.com/pac4j/pac4j/pull/2935

leleuj:
Closed in favor of #2935

PR #2935, opened 2024-06-26
force init saml idp metadata from URLs #2

mmoayyed:


leleuj:
Yes, I think it's a better option to be able to force the reloading.

PR #2940, opened 2024-06-28
Configure hostname verifier on OIDC resource retriever

mmoayyed:
The resource retriever used in the OIDC configuration does not recognize the hostname verifier that is set in the configuration. The retriever is used to fetch remote OIDC keys for token validation, and without the hostname verifier, reaching out to resources that might present a mismatched name could fail (particularly useful for testing against localhost)

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
@@ -451,8 +451,8 @@ public class OidcConfiguration extends BaseClientConfiguration {
     private class OidcResourceRetriever extends DefaultResourceRetriever {
 
         public OidcResourceRetriever() {
-            setConnectTimeout(getConnectTimeout());
-            setReadTimeout(getReadTimeout());
+            setConnectTimeout(OidcConfiguration.this.getConnectTimeout());
+            setReadTimeout(OidcConfiguration.this.getReadTimeout());
             if (sslSocketFactory != null) {
                 setSslSocketFactory(sslSocketFactory);
             }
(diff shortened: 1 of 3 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
```diff
@@ -198,9 +253,7 @@ protected void internalInit(final boolean forceReinit) {
         // default value
         if (forceReinit || getResourceRetriever() == null) {
             try {
-                setResourceRetriever(sslSocketFactory == null
-                    ? new DefaultResourceRetriever(getConnectTimeout(),getReadTimeout())
-                    : new DefaultResourceRetriever(getConnectTimeout(),getReadTimeout(), 0, false, sslSocketFactory));
+                setResourceRetriever(new OidcResourceRetriever());
```
mmoayyed, line 256: This is the main change here

pac4j-oidc/src/main/java/org/pac4j/oidc/config/OidcConfiguration.java
```diff
@@ -404,4 +447,29 @@ public void setValueRetriever(ValueRetriever valueRetriever) {
...
+        @Override
+        protected HttpURLConnection openHTTPConnection(URL url) throws IOException {
+            var connection = super.openHTTPConnection(url);
+            if (connection instanceof HttpsURLConnection) {
+                if (sslSocketFactory != null) {
+                    ((HttpsURLConnection) connection).setSSLSocketFactory(sslSocketFactory);
+                }
+                if (hostnameVerifier != null) {
+                    ((HttpsURLConnection) connection).setHostnameVerifier(hostnameVerifier);
+                }
```
mmoayyed, line 470: We set the hostname verifier here if one is defined.

leleuj:
Don't forget to drop a line in the release notes.

mmoayyed:
I was able to test this with Keycloak as the delegated IDP in CAS, and this works correctly. This should be the last piece of the puzzle for now.

PR #3084, opened 2024-09-29
Allow serializing SAML2Profiles using the JsonSerializer with default typing

mkr:
Allow serializing SAML2Profiles using the JsonSerializer with default typing

* Adding the jackson-datatype-jsr310 dependency, using its JavaTimeModule to the JsonSerializer to enable (de-)serializing of java.time classes
* Changing the SAML2Profile#notBefore and SAML2Profile#notOnOrAfter properties to java.time.Instant as they don't need time-zone information

See https://github.com/pac4j/play-pac4j/issues/705 for context

Before submitting any pull request, please read the contribution guide: https://www.pac4j.org/docs/contribute.html

### DIFF
--- a/pac4j-core/pom.xml
+++ b/pac4j-core/pom.xml
@@ -24,7 +24,10 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
-
+        <dependency>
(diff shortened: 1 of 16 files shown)


leleuj:
Why do you move from `ZonedDateTime` to `Instant`?

leleuj:
Excellent! Merging in the master branch which is the new `6.1.0` version...

PR #3154, opened 2024-11-08
Remove Joda-Time dependency

mtughan:
Joda-Time is no longer used in the codebase and doesn't need to be listed as a dependency anymore. Remove it.


leleuj:
Good catch!

PR #3188, opened 2024-12-03
Resolving Maven Packages for springboot

rayacode:
Adding springboot.version to spring packages, adding m2 repository to resolve prometheus imports they are not present in spring milestone


leleuj:
Why adding `https://repo.maven.apache.org/maven2`? What problem are you trying to solve? As the `spring-boot-dependencies` dependency is used, no `${springboot.version}` definition should be needed.

leleuj:
This must be a problem on your environment as the build properly passes on the CI.

PR #3268, opened 2025-02-02
Exclude gpl-licensed findbugs-annotations

mmoayyed:
findbugs-annotations is under GPL which is not compatible with Apache v4: 
https://github.com/findbugsproject/findbugs/blob/release-3.0.1/findbugs/licenses/LICENSE.txt


leleuj:
Good catch! Thank you

PR #3272, opened 2025-02-06
Create TokenRequest with the configured OIDC scope

ilgrosso:
This PR bears a small change to pass the OIDC scope as configured to the `TokenRequest` constructor.

Side note: it seems that the constructor variants without the `scope` argument were deprecated.


leleuj:
Good! Can you drop a line in the release notes as well? Thx

ilgrosso:
@leleuj done
I hope the phrasing is clear and concise enough

PR #3289, opened 2025-02-18
feat: allow oidc metadata url to use hostname verifier/socket factory

mmoayyed:
When building the discovery URI for OIDC, pac4j ultimately builds a `UrlResource` from Spring:

```java
SpringResourceHelper.buildResourceFromPath(configuration.getDiscoveryURI());
```

This resource does not by default allow one to configure its hostname verifier and/or socket factory. When pac4j attempts to do an internal load operation, it reaches out to this URL to check for the last-modified date, and the check here could fail without the configured hostname verifier or socket factory. 

The PR creates a `OidcMetadataUrlResource` as an extension that allows us to set the hostname verifier and socket factory, which appears to be the only way to set these fields.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/metadata/OidcOpMetadataResolver.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/metadata/OidcOpMetadataResolver.java
@@ -6,6 +6,7 @@ import com.nimbusds.oauth2.sdk.ParseException;
 import com.nimbusds.oauth2.sdk.auth.*;
 import com.nimbusds.oauth2.sdk.id.ClientID;
 import com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
@@ -16,8 +17,12 @@ import org.pac4j.oidc.exceptions.OidcException;
 import org.pac4j.oidc.exceptions.OidcUnsupportedClientAuthMethodException;
 import org.pac4j.oidc.profile.creator.TokenValidator;
 import org.springframework.core.io.Resource;
+import org.springframework.core.io.UrlResource;
 
+import javax.net.ssl.HttpsURLConnection;
 import java.io.IOException;
+import java.net.URL;
+import java.net.URLConnection;
 import java.util.Arrays;
 import java.util.Collection;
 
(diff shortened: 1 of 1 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/metadata/OidcOpMetadataResolver.java
```diff
@@ -186,4 +194,27 @@ private static ClientAuthenticationMethod firstSupportedMethod(
     protected TokenValidator createTokenValidator() {
         return new TokenValidator(configuration, this.loaded);
     }
+
+    @EqualsAndHashCode(callSuper = true)
+    private static class OidcMetadataUrlResource extends UrlResource {
```
leleuj, line 199: Why is it not a public class to allow easy overriding?

mmoayyed: To be sure, are you asking for `OidcMetadataUrlResource` to be public?

Can do, but I am not sure why anyone would prefer to use an internal class rather than just extending `UrlResource` itself. Also, even if they extend this somehow, where and how could someone use this?

leleuj: Yes, that was my original question. Indeed, people would certainly prefer to extend `UrlResource`...

pac4j-oidc/src/main/java/org/pac4j/oidc/metadata/OidcOpMetadataResolver.java
```diff
@@ -60,7 +65,10 @@ public OidcOpMetadataResolver(final OidcConfiguration configuration) {
 
     private static Resource buildResource(final OidcConfiguration configuration) {
         if (configuration != null && configuration.getDiscoveryURI() != null) {
-            return SpringResourceHelper.buildResourceFromPath(configuration.getDiscoveryURI());
+            var resource = SpringResourceHelper.buildResourceFromPath(configuration.getDiscoveryURI());
+            if (resource instanceof final UrlResource urlResource) {
```
RichPavyVV, line 69: This introduces a regression: if `SpringResourceHelper.buildResourceFromPath` returns a ClassPathResource or a FileSystemResource, the resource ends up being null.

leleuj: Good catch! This should be fixed: https://github.com/pac4j/pac4j/commit/3d33ac4a98bdb1921b70456a3c8b2b6f85fdeec0

RichPavyVV: Merci!

PR #3315, opened 2025-03-04
Fixes initial init with forceReinit == true

wograbow:
In https://github.com/apereo/cas/blob/master/support/cas-server-support-pac4j-core/src/main/java/org/apereo/cas/web/flow/DefaultDelegatedClientIdentityProviderConfigurationProducer.java in initializeClientIdentityProvider method init is called with forceReinit == true
As client is uninitialized its unnecessary to force reinit although it shouldn't hurt.  
AzureAd2OidcConfiguration properly sets opMetadataResolver with new AzureAdOpMetadataResolver(this) but after this super.internalInit(forceReinit: true);  is called and in OidcConfiguration class proper value of opMetadataResolver is overwritten by new OidcOpMetadataResolver(this). 
Proposed fix makes AzureAd2OidcConfiguration immune for init called with force == true; It still gets proper opMetadataResolver

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -8,6 +8,7 @@ title: Release notes&#58;
 **v6.1.2**:
 - Use the configured scope in OpenID Connect authenticator
 - Fix the `getFullRequestURL` method
+- Fixes setting proper implementation of OidcOpMetadataResolver in OidcConfiguration and its descendants when `internalInit` is called with `forceReinit` set to true
 
 **v6.1.1**:
 - Protect the `getRequestAttribute` method for Jetty 12.0.8+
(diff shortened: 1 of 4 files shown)


leleuj:
LGTM.

Can you do the same for `AppleOidcConfiguration`  and `KeycloakOidcConfiguration` as well as drop a line in the release notes: https://github.com/pac4j/pac4j/blob/master/documentation/docs/release-notes.md ?

And fix the build ("unused import")...

wograbow:
Done for `AppleOidcConfiguration` but `KeycloakOidcConfiguration` uses default `OidcOpMetadataResolver` so overriden method would only call its super or have only code copied from its super

leleuj:
There won't any 6.0.x release, please close the PR. The next release will be the v6.1.2. Please update the release notes.

I don't see any change for Keycloak...

wograbow:
> I don't see any change for Keycloak...

KeycloakOidcConfiguration uses default OidcOpMetadataResolver so overriden method would only call its super or have only code copied from its super

leleuj:
OK. good

PR #3342, opened 2025-03-25
feat: Resolve saml2 metadata url concurrently via ranges

mmoayyed:
This pull request attempts to download and resolve metadata concurrently using the number of cores available to the system, if the metadata resource is a URL and supports the `Accept-Ranges` header as `bytes`. When this condition is true, the metadata download is split into multiple threads and downloaded partially, and then assembled into a temporary file before it's parsed and resolved. 

If the resource is not a URL or if the URL does not support this concept, we fall back to the current method.  

This is primarily useful with large metadata artifacts, such as those that are aggregates. 

The test demonstrates the performance improvement using `https://md.incommon.org/InCommon/InCommon-metadata-idp-only.xml` which is approximately 50+ MBs. 

On my workstation with approximately 10 cores:

- Without this change: `Metadata resolution took: 38881 ms`
- With this change: `Metadata resolution took: 8629 ms`

As you observe, the performance gain is roughly 4x.

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -8,7 +8,8 @@ title: Release notes&#58;
 **v6.1.2**:
 - Use the configured scope in OpenID Connect authenticator
 - Fix the `getFullRequestURL` method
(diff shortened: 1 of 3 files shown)


leleuj:
Interesting performance improvement!

PR #3350, opened 2025-03-28
Allow SP entities to be merged together

mmoayyed:
## What

This pull request allows `SAML2Client` instances to merge their SP metadata together, if they share the same SP metadata reference. Presently, if there are multiple clients sharing the same metadata file, the first one that is initialized creates the metadata, and the second one is ignored. So we end up with:

```xml
<md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://example.org/auth?client_name=SAML2Client1" index="0"/>
```

But we want:

```xml
<md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://example.org/auth?client_name=SAML2Client1" index="1"/>
<md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://example.org/auth?client_name=SAML2Client2" index="2"/>
<md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://example.org/auth?client_name=SAML2Client3" index="3"/>
```

## How

If the metadata resource already exists, the process merges relevant entries from the new entity into the existing entity, and the result is stored if there are any changes. The changes specifically consider:

1. Request Initiators
2. SLO Services
3. ACS Services

These changes are only processed and included if the metadata entity already does not contain them, to avoid creating duplicate references. This change works with all metadata generators, and is 100% binary backward compatible

## What about the other stuff? 

The following changes/optimizations are in place:

1. The SAML2 parser pool is adjusted to have a larger pool size to improve performance.
2. There is `SAML2DelegatingMetadataResolver` that delegates resolution tasks to an existing metadata resolver, for performance reasons.
3.  The `SAML2InMemoryMetadataGenerator`only keeps data in memory and acts as a virtual generator, for performance reasons.
4. Specific properties in the SAML2Client are only initialized if they are not already set, to avoid useless object creation that might be exactly the same.

## Test Scenario

I have the following SAML2Client configurations:

1. `SAML2Client1` talks to a single IDP 1
2. `SAML2Client2` talks to an IDP metadata aggregate, that contains about 7000 IDPs. 

These two clients **ARE EXACTLY THE SAME** in every which way, except that their names are different, and therefore their URLs would be different. Thus, the SP metadata they share must be merged and adjusted to account for both. It is possible of course to separate their SP metadata, via say, dedicated files, but that would be a useless and confusing thing to maintain.

## What performance reasons? 

`SAML2Client2` talks to an IDP metadata aggregate, that contains about 7000 IDPs. When pac4j is initialized, `SAML2Client2` is unpacked, and translates to many different `SAML2Client` instances, for each individual IDP. So we end up with 7000 `SAML2Client` instances in theory, that are all *only virtually and lazily loaded* (this is where the extra metadata resolver components come in, noted above) to avoid heap explosion. This is where performance comes in: the original `SAML2Client` that is initialized once can be **REUSED** for the other 7000, and there is no need to keep creating new objects that are exactly the same as the one before.

Once `SAML2Client2` is unpacked to 7000 instances, it itself goes away and is discarded. 

The changes here allow one and enable this sort of optimization. With the recent changes in metadata resolution and the ability to avoid object creation unnecessarily, I was able to bring down the load time from 33 seconds to less than 3 seconds, on a 10-core M1 machine. There are many other optimizations  in CAS to enable this, but the main bottleneck is initializing `SAML2Client` instances per IDP and removing the need for object creation.

This has been tested with some confidence, but I suspect there are rough edges that need to be polished. If/When merged, I'd like to run the full CAS test suite against the pac4j SNAPSHOTs to verify that all is well as it should be.


mmoayyed:
If it helps, I am testing this in CAS with a puppeteer scenario that runs:

```json
    "--cas.authn.pac4j.saml[0].keystore-password=pac4j-demo-passwd",
    "--cas.authn.pac4j.saml[0].private-key-password=pac4j-demo-passwd",
    "--cas.authn.pac4j.saml[0].service-provider-entity-id=cas:apereo:pac4j:saml",
    "--cas.authn.pac4j.saml[0].metadata.service-provider.file-system.location=${PWD}/ci/tests/puppeteer/scenarios/${SCENARIO}/saml-md/sp-metadata.xml",
    "--cas.authn.pac4j.saml[0].keystore-path=${PWD}/ci/tests/puppeteer/scenarios/${SCENARIO}/saml-md/samlKeystore.jks",
    "--cas.authn.pac4j.saml[0].metadata.identity-provider-metadata-path=http://localhost:9443/simplesaml/saml2/idp/metadata.php",
    "--cas.authn.pac4j.saml[0].destination-binding=urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect",
    "--cas.authn.pac4j.saml[0].client-name=SAML2Client",

    "--cas.authn.pac4j.saml[1].keystore-password=pac4j-demo-passwd",
    "--cas.authn.pac4j.saml[1].private-key-password=pac4j-demo-passwd",
    "--cas.authn.pac4j.saml[1].service-provider-entity-id=cas:apereo:pac4j:saml",
    "--cas.authn.pac4j.saml[1].metadata.service-provider.file-system.location=${PWD}/ci/tests/puppeteer/scenarios/${SCENARIO}/saml-md/sp-metadata.xml",
    "--cas.authn.pac4j.saml[1].keystore-path=${PWD}/ci/tests/puppeteer/scenarios/${SCENARIO}/saml-md/samlKeystore.jks",
    "--cas.authn.pac4j.saml[1].metadata.identity-provider-metadata-path=file:/etc/cas/config/InCommon-metadata-idp-only.xml",
    "--cas.authn.pac4j.saml[1].metadata.identity-provider-metadata-aggregate=true",
    "--cas.authn.pac4j.saml[1].destination-binding=urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect",
    "--cas.authn.pac4j.saml[1].client-name=SAML2InCommon",
```

Note:

```
   "--cas.authn.pac4j.saml[1].metadata.identity-provider-metadata-path=file:/etc/cas/config/InCommon-metadata-idp-only.xml",
    "--cas.authn.pac4j.saml[1].metadata.identity-provider-metadata-aggregate=true",
```

And the fact that clients are almost exactly the same.

mmoayyed:
Once the changes here are good, and the review is done, I can also update the release notes.

leleuj:
It looks very good to me. Just add the release notes and let's merge it...

PR #3369, opened 2025-04-08
Resolve file-system sp metadata once w/o reloads

mmoayyed:
This pull request removes the `FilesystemMetadataResolver` and replaces it with a DOM parser only. This change allows the metadata to only be parsed once, and without background timer threads that tend to refresh the metadata periodically (and cannot be disabled).

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -5,6 +5,9 @@ title: Release notes&#58;
 
 ### JDK17:
 
+**v6.1.3**:
+- SAML2 operations that use `FilesystemMetadataResolver` are replaced with a DOM parser instead.
+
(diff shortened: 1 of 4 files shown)


leleuj:
Can you elaborate? Why did you need the `destroy` and `close` methods? Submit a PR if need be...

boris-petrov:
Well, `destroy` is/was a public method in `SAML2Client`. So simply removing it is a breaking change that should wait for a major version of `pac4j` (version 7). But no big deal, we can just remove the call to `destroy`. As to why I was using it - well, when shutting down my app, I tend to clean up all used resources. `destroy` seemed the right method to call on `SAML2Client` to do that.

PR #3430, opened 2025-05-30
Allow SAML2 idp url resource to consider SAML2 connection config

mmoayyed:
## Problem

This patch fixes a minor problem (or adds an enhancement) when the SAML2 identity provider metadata is configured using a URL. When the resolution process kicks in, the loading process attempts to check for changes by invoking `hasChanged()` and eventually `lastModified` on the URL resource. If the URL causes a SSL handshake exception, metadata is never downloaded which eventually causes the application to error out.

This generally is expected; however if the SAML2 configuration is configured to use a custom SSL context and hostname verifier, these are ignored during this step. They are only taken into account when the metadata is itself downloaded, but we are not quite there while checking for changes to determine the need for the download.

## Fix

The fix here wraps the current URL resource into one that is aware of the SAML2 configuration, and customizes the connection as necessary before interacting with the real source.

Tests are provided to demonstrate the difference with or without configuration options.

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -5,6 +5,9 @@ title: Release notes&#58;
 
 ### JDK17:
 
+**v6.1.4**:
+- Downloading SAML2 metadata over a URL is able to support SSL context and hostname verification options when checking for metadata updates.
+
 **v6.1.3**:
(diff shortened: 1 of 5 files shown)


leleuj:
LGTM. Yet there are test failures to fix.

PR #3535, opened 2025-08-22
SunJaasKerberosTicketValidator fails with spaces in keytab path

valenpo:
Description: SunJaasKerberosTicketValidator fails to load keytab files when path contains spaces due to URL encoding issue.

Steps to Reproduce:
Place keytab in path with spaces: /Library/Application Support/app/keytab.keytab

1
validator.setKeyTabLocation(new FileSystemResource("/Library/Application Support/app/keytab.keytab"));
Attempt Kerberos authentication
Observe error: Cannot find key of appropriate type to decrypt AP-REQ
 Root Cause:  In internalInit(), the code removes file: prefix but doesn't decode URL-encoded characters like %20:


leleuj:
OK. It makes sense, but isn't URL decoding too much for a regular path containing a percent for example or will the percent be URL encoded as well via `new FileSystemResource`?

leleuj:
Yes, that makes sense and it seems safer to me. Fix the build as well

leleuj:
Closing as I don't see any follow-up here...

valenpo:
I'll provide today.

leleuj:
No problem. I thought I gave up. Reopening...

PR #3865, opened 2026-05-13
Remove outdated lifecycle warning in documentations

zackman0010:
v6.1.3 removed the `FilesystemMetadataResolver` that made this note in the documentation necessary, it's now outdated and no longer applies.

An AI agent I was overseeing was attempting to implement this, found the documentation, saw the note about needing to call `destroy`, then went down a 30 minute rabbit hole trying to find a `destroy` method that no longer existed. To be fair to the AI, I probably would have done the same thing.


leleuj:
Good catch! Thanks

PR #3871, opened 2026-05-18
OpenID Federationg explicit registration fix

xpavlic:
According to the OpenID Federation 1.0 explicit registration request Content type MUST be `application/entity-statement+jwt` https://openid.net/specs/openid-federation-1_0.html#section-12.2.1-6 and explicit registration response Content type must be `application/explicit-registration-response+jwt` https://openid.net/specs/openid-federation-1_0.html#section-12.2.3-9. 

Another issue was the absence of a REQUIRED audience claim in the explicit registration request. 

This PR fixes both issues and also includes updated tests.

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -5,6 +5,11 @@ title: Release notes&#58;
 
 ### JDK17:
 
+**v6.5.2**:
+- Fix OpenID federation explicit registration request Content-Type value to `application/entity-statement+jwt`
+- Fix OpenID federation explicit registration response Content-Type validation to expect `application/explicit-registration-response+jwt`
+- Added missing audience claim to the OpenID federation explicit registration request
+
 **v6.5.1**:
 - Security fix on the `pac4j-core` module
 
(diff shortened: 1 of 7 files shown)


leleuj:
LGTM.

Two comments:
- can you add a Javadoc at the class level for `ExplicitRegistrationResponse`? I'm a bit surprised that Nimbus is late on the spec, this might be in a future version...
- can you drop a line on the release notes? -> https://github.com/pac4j/pac4j/blob/master/documentation/docs/release-notes.md (v6.5.2)

One question:
- how did you find these problems? With which OIDC federation provider did you test?

xpavlic:
I am part of the EOSC Beyond project, within which we are extending the capabilities of multiple AAI solutions, including support for OpenID Federation. I am the lead developer for the Perun AAI, and we are currently replacing the old proxy software stack with Apereo CAS. As you probably know, CAS uses Pac4j for delegated authentication, so I used your implementation in our Perun CAS overlay and tested it in the ESOC Beyond OpenID Federation sandbox.

leleuj:
Excellent! Merged. I will cut the release pretty soon.

PR #3877, opened 2026-05-21
feat: openid federation explicit clinet registration respect client poperties from the response

xpavlic:
This PR adjusts the handling of the explicit registration response, ensuring that RP respects the `scope`, `token_endpoint_auth_method`, `request_object_signing_alg`, `id_token_signed_response_alg` and `response_types` claims included in the metadata in the OpenID federation explicit registration response. 

The metadata can possibly include all of the claims described here: [IANA Dynamic registration](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#:~:text=%5BRFC9679%5D-,OAuth%20Dynamic%20Client%20Registration%20Metadata,-Registration%20Procedure(s)), so I added only those that are already represented in the client configuration.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/metadata/registration/FederationClientRegister.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/metadata/registration/FederationClientRegister.java
@@ -2,11 +2,15 @@ package org.pac4j.oidc.metadata.registration;
 
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.nimbusds.jose.JOSEException;
 import com.nimbusds.jose.JOSEObjectType;
+import com.nimbusds.jose.JWSAlgorithm;
 import com.nimbusds.jose.jwk.JWKSet;
 import com.nimbusds.jose.proc.BadJOSEException;
+import com.nimbusds.oauth2.sdk.ResponseType;
+import com.nimbusds.oauth2.sdk.auth.ClientAuthenticationMethod;
 import com.nimbusds.oauth2.sdk.http.HTTPRequest;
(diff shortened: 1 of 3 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/metadata/registration/FederationClientRegister.java
```diff
@@ -33,6 +37,15 @@
 @Slf4j
 public class FederationClientRegister {
 
+    private static final List<String> PREFERRED_RESPONSE_TYPES = List.of(
```
leleuj: Could we use `AUTHORIZATION_CODE_FLOWS`, `IMPLICIT_FLOWS` and `HYBRID_CODE_FLOWS` from the `OidcConfiguration` class?

pac4j-oidc/src/main/java/org/pac4j/oidc/metadata/registration/FederationClientRegister.java
```diff
@@ -122,6 +135,32 @@ protected void performExplicitRegistration(final OidcConfiguration configuration
                         FileHelper.savePrivateFile(path, clientSecret);
                     }
                     logSeparator();
+                    val scope = orp.path("scope").asText();
+                    if (StringUtils.isNotBlank(scope)) {
+                        configuration.setScope(scope);
+                    }
+                    val tokenEndpointAuthMethod = orp.path("token_endpoint_auth_method").asText();
+                    if (StringUtils.isNotBlank(tokenEndpointAuthMethod)) {
+                        configuration.setClientAuthenticationMethodAsString(tokenEndpointAuthMethod);
```
leleuj: I think we should check the method against the `SUPPORTED_METHODS` from `DefaultClientAuthenticationBuilder`(after making it public).

leleuj:
I was about to cut the release but this can wait for this one. LGTM. 2 comments

leleuj (approved):

leleuj:
I just cut the v6.5.2 release.

PR #3998, opened 2026-08-18
JEESessionStore - Check if session is already invalidated

weber-semedy:
Use case OIDC backchannel logout: This may happen when we configure e.g. back channel logout and central logout, where is necessary to use JEESessionStore to track sessions. A user may click logout button, invalidating the session. The IdP may send another request for logout (backchannel logout). If this happens and do not check whether session is invalidated, we may run into an `IllegalStateException` as the code tries to remove the session again, calling the set method in JEESessionStore.

See https://openid.net/specs/openid-connect-backchannel-1_0.html#BCActions:

>  If the identified End-User is already logged out at the RP when the logout request is received, the logout is considered to have succeeded.

Maybe there is a quicker way to avoid calling the destroy method altogether (OidcLogoutProcessor) to comply with the spec.

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -10,6 +10,7 @@ title: Release notes&#58;
 - Rework the `SpringResourceLoader` behavior to improve the OIDC/SAML metadata loading resilience:
   - retry every 2 seconds before anything has been loaded
   - retry every 60 seconds after the first loading and return the old data in case of error
+- JEESessionStore: Consider invalidated session in set method (handle IllegalStateException)
 
 **v6.5.5**:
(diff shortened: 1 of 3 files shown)


leleuj:
Good point!

weber-semedy:
@leleuj for when do you plan the release of 6.6.0?

leleuj:
v6.6.0 is a new stream with a big development planned on it so it won't be available before a few months.

Yet, a security release (v6.5.6) is planned today so I backported your changes on the 6.5.x branch to be embedded with the security release.

weber-semedy:
Thank you @leleuj .

PR #4007, opened 2026-08-24
OIDC: Fix nonce requirmemt for stateless api access via bearer token

weber-semedy:
`OidcProfileCreator` attempts to look up a session-bound nonce even when processing a stateless API request using a `BearerAccessToken`. In stateless environments, this triggers unnecessary session lookups and forces developers to manually disable` setUseNonce(false)` globally.

Solution:Reuse the existing `regularOidcFlow` flag to gate the nonce lookup. The session check now runs only for standard browser authentication (`OidcCredentials`). For direct bearer tokens, we now skip the logic that applies to regular oidc flow only.

### DIFF
--- a/documentation/docs/release-notes.md
+++ b/documentation/docs/release-notes.md
@@ -12,6 +12,9 @@ title: Release notes&#58;
   - retry every 60 seconds after the first loading and return the old data in case of error
 - Fix OpenID federation forced signing of the auth request object for explicitly registered clients when it is not mandatory.
 
+**v6.5.7**:
+- OIDC: ProfileCreator - skip nonce when Bearer Access Token is used
+
 **v6.5.6**:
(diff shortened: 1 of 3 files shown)


leleuj:
I'm in line with this change. Though, I'm not sure you target the right branch, the release notes says v6.5.7 but the PR targets the master (master is v6.6.0-SNAPSHOT / the 6.5.x branch is v6.5.7-SNAPSHOT).

Which version do you target? Maybe both (in that case, you need to submit 2 PRs)

weber-semedy:
@leleuj Thanks for the guidance, I applied the changes to the 6.5.x branch in this PR https://github.com/pac4j/pac4j/pull/4011 and added the changes in the rease notes there.

leleuj:
Thanks

PR #4008, opened 2026-08-24
fix: allow client_secret_basic and client_secret_post for explicitly registered clients

xpavlic:
The code was forcing the signing of the auth request object for all OpenID federation clients. When the client is registered using explicit client registration, the auth request object signing does not have to be mandatory, as the client with client credentials can use `client_secret_basic` and `client_secret_post` methods if supported.

This fixes the issue and makes signing of the auth request object optional for explicitly registered clients.


leleuj:
Can you add some test(s) to confirm the fix?  Thanks

leleuj:
Merging. Will add the tests by myself... Thanks for the contribution

PR #4019, opened 2026-08-30
feat: allow login-hint per request attribute

mmoayyed:
`login_hint` often needs to be determined dynamically, based on user provided input or session data. This change allows it to be found dynamically via request attributes, optionally, rather than just statically configured.


leleuj:
It makes sense. Can you add tests to confirm the change and a line in the release notes? Thx

PR #4028, opened 2026-09-03
Fix: Clear attemptedAuthentication flag on OIDC validation failure in callback to prevent 401 on retry

weber-semedy:
Problem:
We recently stumbled over this problem when retrying a login:
When OIDC token validation fails, `validateCredentials` returns null. This triggers a cascade into the finally block, where `IndirectClient#checkCredentials` blindly writes `attemptedAuthentication = "true"` into the session. If the application routes the user to an error page, any manual retry link immediately hits a 401 Unauthorized in the `SecurityFilter`. The framework mistakes the user's manual action for an redirect to 401 because of the leftover session flag.

Solution:
Overrode `checkCredentials` in OidcClient to intercept this specific failure state. If a validation fails on the callback path, the method actively clears the `attemptedAuthentication` flag from the session store and aborts execution. This short-circuits the base class logic and prevents the blocking flag from being written.

### DIFF
--- a/pac4j-oidc/src/main/java/org/pac4j/oidc/client/OidcClient.java
+++ b/pac4j-oidc/src/main/java/org/pac4j/oidc/client/OidcClient.java
@@ -6,6 +6,7 @@ import lombok.val;
 import org.pac4j.core.client.IndirectClient;
 import org.pac4j.core.context.CallContext;
 import org.pac4j.core.credentials.Credentials;
+import org.pac4j.core.http.url.DefaultUrlResolver;
 import org.pac4j.core.profile.UserProfile;
 import org.pac4j.oidc.config.OidcConfiguration;
 import org.pac4j.oidc.credentials.OidcCredentials;
@@ -93,10 +94,11 @@ public class OidcClient extends IndirectClient {
 
         if (credentials == null && getProfileFactoryWhenNotAuthenticated() == null) {
             val currentPath = webContext.getPath();
+            val defaultUrlResolver = new DefaultUrlResolver(true);
             val resolver = this.getCallbackUrlResolver();
             if (resolver != null && currentPath != null) {
                 val computedCallbackUrl = resolver.compute(
-                    this.getUrlResolver(),
+                    defaultUrlResolver,
                     this.getCallbackUrl(),
                     this.getName(),
                     webContext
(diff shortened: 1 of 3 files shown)


pac4j-oidc/src/main/java/org/pac4j/oidc/client/OidcClient.java
```diff
@@ -85,6 +86,39 @@ protected void internalInit(final boolean forceReinit) {
...
+                );
+                if (computedCallbackUrl != null) {
+                    try {
+                        val expectedCallbackPath = new java.net.URI(computedCallbackUrl).getPath();
+                        if (currentPath.equals(expectedCallbackPath)) {
+                            logger.debug("OIDC callback execution failed validation on explicit path '{}'. " +
+                                "Cleaning attempted authentication to prevent 401 error.", currentPath);
+                            cleanAttemptedAuthentication(webContext, sessionStore);
+                            return;
+                        }
```
weber-semedy, line 114: I would highly appreciate your review on this logic.
The check for the path must be exact and consider the type of callback resolver.
Also, we need to take into account the settinngs of UrlResolver. Thus the code is not working yet for each case. Maybe we should create a separate url resolver for this case instead on relying on what is configured fr this client.

leleuj:
I see your point. Though, this solution really seems too fragile. Tests are necessary and the PR must address the master first.

This behavior was made to deal with cancelled authentications.

The `CredentialsExtractor` and the `Authenticator` have similar contracts: empty credentials or a `CredentialsException`.

My feeling is that the pac4j code is buggy from the start: the `checkCredentials` should not happen in the `finally`, but in the `catch` and in the regular flow. Any other unexpected/unwanted/uncontrolled exception should allow the retry.

Here is the change I made: https://github.com/pac4j/pac4j/commit/d36633a026c9e3595bc24bfd23e80315f4c87745

This will stay in master as it is somehow a breaking change. You can locally override your `BaseClient` class with the same change in v6.5.x.

weber-semedy:
> I see your point. Though, this solution really seems too fragile. Tests are necessary and the PR must address the master first.

I wanted to check in first before putting too much work into it. I would of course add tests if there would have been a green light by you.
I agree the callback check might have been too fragile and might have not been sufficient to deal with the framework complexities and there are better spots to address this issue.

Thanks for providing a fix.

